PR #1, opened 2025-03-13
Initial commit of the server implementation

jhernand:


romfreiman:
@coderabbitai review

Containerfile
```diff
@@ -0,0 +1,38 @@
...
+RUN ./dev.py setup
+
+# Copy the rest of the source code and build the binary:
+COPY . /source
+RUN ./dev.py build
+
+FROM registry.access.redhat.com/ubi9/ubi:9.5-1739751568 AS runtime
+
+# Install the binary:
+COPY --from=builder /source/bin/main /usr/local/bin
```
mhrivnak, line 38: We should give it a better name than `main`.

README.md
```diff
@@ -1,8 +1,199 @@
...
-> material containing Red Hat confidential content, including information about
-> future product development plans.
+## Recommended development setup
+
+This is an opinionated setup, the one used by the person that initially wrote this document. You can find other ways
+to prepare your environment. As long as the resulting code works fine you are good.
+
+### Make sure that you have a version 3.12 or newer of Python available
+
+This is needed because the build tools (the `dev.py` script) of the project are written in Python and uses some recent
```
mhrivnak, line 12: Can you summarize what this does and why it is valuable? On first look, the python appears to be a lot of boilerplate around easy commands like `go build` and `podman build`. I'm just looking for justification for requiring a bunch of python environment setup just to build and manage a go project. Needing to manage a python environment has its own cost of complexity.

manifests/publisher.yaml
```diff
@@ -0,0 +1,43 @@
...
+    spec:
+      containers:
+      - name: publisher
+        image: main
+        imagePullPolicy: IfNotPresent
+        command:
+        - /usr/local/bin/main
+        - start
+        - publiser
+        - --log-level=debug
```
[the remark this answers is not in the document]
jhernand: Done.

internal/database/database_tool.go
```diff
@@ -0,0 +1,276 @@
...
+	"github.com/golang-migrate/migrate/v4/source/iofs"
+	"github.com/jackc/pgx/v5"
+	"github.com/jackc/pgx/v5/pgxpool"
+	"github.com/spf13/pflag"
+)
+
+//go:embed migrations
+var migrationsFS embed.FS
+
+type Tool interface {
```
mhrivnak, line 37: Could you add a doc block explaining what this is? I'm not finding that the name "tool" is self-evident.

internal/logging/logger.go
```diff
@@ -0,0 +1,348 @@
...
+	"sort"
+	"strings"
+	"time"
+
+	"github.com/spf13/pflag"
+)
+
+// LoggerBuilder contains the data and logic needed to create a logger. Don't create instances of this directly, use the
+// NewLogger function instead.
+type LoggerBuilder struct {
```
mhrivnak, line 30: What's the reason for creating our own vs. importing one of the commonly-used existing loggers?

manifests/database.yaml
```diff
@@ -0,0 +1,78 @@
...
+        app: fulfillment-database
+    spec:
+      volumes:
+      - name: data
+        persistentVolumeClaim:
+          claimName: fulfillment-database
+      containers:
+      - name: server
+        image: postgres
+        imagePullPolicy: IfNotPresent
```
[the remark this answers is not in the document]
jhernand: This is replaced by the image+digest during deployment, by kustomize.

manifests/database.yaml
```diff
@@ -0,0 +1,78 @@
...
+      storage: 10Gi
+
+---
+
+apiVersion: v1
+kind: Service
+metadata:
+  name: fulfillment-database
+  labels:
+    app: fullfillment-database
```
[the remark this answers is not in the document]
jhernand: Done.

dev/kustomize.py
```diff
@@ -0,0 +1,73 @@
...
+        with tempfile.NamedTemporaryFile(delete=False) as manifests_file:
+            command.run(
+                cwd=tmp_dir,
+                args=[
+                   tools.KUSTOMIZE.name,
+                   "build", ".",
+                ],
+                stdout=manifests_file,
+            )
+            return pathlib.Path(manifests_file.name)
```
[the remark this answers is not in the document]
jhernand: Done.

dev/kustomize.py
```diff
@@ -0,0 +1,73 @@
...
+                args=[
+                    "podman",
+                    "inspect",
+                    f"--format={{{{ index .RepoDigests 0 }}}}",
+                    image,
+                ],
+            )
+            if code != 0:
+                raise Exception(f"Failed to get image digest for image {image}")
+            command.run(
```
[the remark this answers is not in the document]
jhernand: Done.

mhrivnak (changes requested):

internal/auth/auth_context.go
```diff
@@ -0,0 +1,44 @@
...
+// context.
+func SubjectFromContext(ctx context.Context) *Subject {
+	subject := ctx.Value(subjectContextKey)
+	switch subject := subject.(type) {
+	case *Subject:
+		return subject
+	default:
+		panic("failed to get subject from context")
+	}
+}
```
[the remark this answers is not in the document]
jhernand: Panic is desired here: this will never happen.

manifests/kustomization.yaml
```diff
@@ -0,0 +1,39 @@
...
+# specific language governing permissions and limitations under the License.
+#
+
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+resources:
+- namespace.yaml
+- database.yaml
+- service.yaml
```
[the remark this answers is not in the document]
jhernand: Done.

dev/build.py
```diff
@@ -0,0 +1,116 @@
...
+    """
+    Builds the bundle image.
+    """
+    command.run(
+        args=[
+            "podman", "build",
+            "--file", "Containerfile",
+            "--tag", f"{repository}:{tag}"
+            ".",
+        ],
```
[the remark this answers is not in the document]
jhernand: This code is never used, removed it.

internal/database/models/cluster_order_model.go
```diff
@@ -0,0 +1,74 @@
...
+	case "ACCEPTED":
+		*s = ClusterOrderStateAccepted
+	case "REJECTED":
+		*s = ClusterOrderStateRejected
+	case "FAILED":
+		*s = ClusterOrderStateFailed
+	case "CANCELED":
+		*s = ClusterOrderStateCanceled
+	case "FULLFILLED":
+		*s = ClusterOrderStateFulfilled
```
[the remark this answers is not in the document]
jhernand: Done.

internal/database/models/cluster_order_model.go
```diff
@@ -0,0 +1,74 @@
...
+	case ClusterOrderStateFulfilled:
+		return "FULFILLED"
+	default:
+		return fmt.Sprintf("UNKNOWN(%d)", s)
+	}
+}
+
+func (s *ClusterOrderState) Scan(value any) error {
+	text := value.(string)
+	switch text {
```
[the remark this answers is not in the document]
jhernand: Done.

internal/database/migrations/3_change_detector.up.sql
```diff
@@ -0,0 +1,51 @@
...
+  ) values (
+    tg_table_name,
+    lower(tg_op),
+    row_to_json(old.*),
+    row_to_json(new.*)
+  );
+  notify changes;
+  return null;
+end;
+$$ language plpgsql;
```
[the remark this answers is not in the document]
jhernand: This works fine for both INSERT and DELETE, no need to check for NULL, the database handles it correctly and generates an empty JSON document.

internal/database/migrations/3_change_detector.up.sql
```diff
@@ -0,0 +1,51 @@
...
+--
+
+create table changes (
+  serial serial not null primary key,
+  timestamp timestamp with time zone not null default now(),
+  source text,
+  operation text,
+  old jsonb,
+  new jsonb
+);
```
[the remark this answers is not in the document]
jhernand: No need for that, this table is only queried by `serial`.

internal/network/grpc_client.go
```diff
@@ -0,0 +1,194 @@
...
+		flag string
+		err  error
+	)
+	failure := func() {
+		b.logger.Error(
+			"Failed to get flag value",
+			slog.String("flag", "flag"),
+			slog.String("error", err.Error()),
+		)
+	}
```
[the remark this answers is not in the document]
jhernand: Done.

internal/servers/clusters_server.go
```diff
@@ -0,0 +1,161 @@
...
+}
+
+func (s *ClustersServer) mapOutbound(from *models.Cluster, to *api.Cluster) error {
+	to.Id = from.ID
+	if to.Spec != nil {
+		to.Spec = &api.ClusterSpec{}
+	}
+	if to.Status != nil {
+		to.Status = &api.ClusterStatus{}
+	}
```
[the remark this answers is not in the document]
jhernand: Done.

internal/database/dao/cluster_orders_dao.go
```diff
@@ -0,0 +1,314 @@
...
+	defer func() {
+		err := tx.Commit(ctx)
+		if err != nil {
+			d.logger.ErrorContext(
+				ctx,
+				"Failed to commit transaction",
+				slog.String("error", err.Error()),
+			)
+		}
+	}()
```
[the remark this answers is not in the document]
jhernand: This is true, but we will handle it in a larger refactor to introduce a transaction interceptor.

internal/servers/cluster_orders_server.go
```diff
@@ -0,0 +1,318 @@
...
+		)
+		return
+	}
+	if template == nil {
+		err = grpcstatus.Errorf(
+			grpccodes.Internal,
+			"cluster template with identifier '%s' doesn't exist",
+			templateId,
+		)
+		return
```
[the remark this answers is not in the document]
jhernand: This isn't correct. this is a request to create an order. Not finding the template shouldn't result in a NotFound. Maybe a InvalidArgument.

Containerfile
jhernand: Sure, what do you suggest? `fulfillment-service` maybe?

manifests/kustomization.yaml
```diff
@@ -0,0 +1,39 @@
...
+
+namespace: innabox
+
+generatorOptions:
+  disableNameSuffixHash: true
+
+images:
+
+- name: main
```
larsks, line 29: We should call this something like `fulfillment-service` instead of `main` (to avoid conflicts with other manifests if these get incorporated into a larger project).

manifests/kustomization.yaml
```diff
@@ -0,0 +1,39 @@
...
+  newName: quay.io/innabox/fulfillment-service
+  newTag: latest
+
+- name: envoy
+  newName: docker.io/envoyproxy/envoy
+  newTag: v1.33.0
+
+- name: postgres
+  newName: quay.io/sclorg/postgresql-15-c9s
```
larsks, line 38: What is `sclorg` and what does that get us that is different from the stock `postgres` image?

manifests/publisher.yaml
```diff
@@ -0,0 +1,43 @@
...
+        app: fulfillment-publisher
+    spec:
+      containers:
+      - name: publisher
+        image: main
+        imagePullPolicy: IfNotPresent
+        command:
+        - /usr/local/bin/main
+        - start
+        - publiser
```
larsks, line 35: This looks like a typo.

README.md
jhernand: This Python code tries to replace the typical Makefiles and hacky shell scripts that every project ends up creating to do things that looked initially easy commands. It is just the same apparently easy things but written in a decent language.

internal/database/database_tool.go
jhernand: Yes, done.

Containerfile
larsks: Maybe `fulfillment-service` :)

internal/logging/logger.go
jhernand: All the code, except the main program, uses the `log\slog` structured logging package that was introduced in Go 1.21. This code just builds such structured logger from the command line flags, and customizes the output to our liking. It also adds a couple of field rewritings, for example to redact the fields that start with the `!` character. Let me stress again that the rest of the code uses directly the `slog.Logger` interface that is part of the Go standard library since 1.21.

internal/logging/logging_suite_test.go
```diff
@@ -0,0 +1,72 @@
...
+			continue
+		}
+		ExpectWithOffset(1, scanner.Err()).ToNot(HaveOccurred())
+		var message map[string]any
+		err := json.Unmarshal(line, &message)
+		ExpectWithOffset(1, err).ToNot(HaveOccurred())
+		result = append(result, message)
+	}
+	return result
+}
```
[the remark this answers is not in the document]
jhernand: Done.

README.md
```diff
@@ -1,8 +1,199 @@
...
+This project contains the code for the fulfillment service.
 
-> [!WARNING]
-> Be mindful of the content you commit to this repository. Do not commit any
-> material containing Red Hat confidential content, including information about
-> future product development plans.
+## Recommended development setup
+
+This is an opinionated setup, the one used by the person that initially wrote this document. You can find other ways
+to prepare your environment. As long as the resulting code works fine you are good.
```
larsks, line 8: I am concerned at the length and granular nature of the instructions in this document.  wonder if we could adopt here the same principal we have adopted for API design, which is "start with the minimal necessary implementation".

internal/servers/events_server.go
```diff
@@ -0,0 +1,103 @@
...
+			s.logger.ErrorContext(
+				ctx,
+				"Failed to send event",
+				slog.String("error", err.Error()),
+			)
+		}
+		time.Sleep(1 * time.Second)
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
jhernand: Done.

README.md
```diff
@@ -1,8 +1,199 @@
...
+
+    $ cd ~/fulfillment-service/repository
+    $ ./dev.py setup
+
+## Development
+
+To build the binary locally use `dev.py build`. Then you can run the `bin/main` binary directly, for example, to start
+the gRPC server:
+
+    $ bin/main start server --db-url=postgres://service:service123@localhost:5432/servcie"
```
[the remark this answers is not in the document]
jhernand: Done.

README.md
```diff
@@ -2,7 +1,199 @@
...
+directory that points to that binary:
+
+    $ python --version
+    Python 3.9.21  <-- This will not work
+
+    $ which python3.12
+    /usr/bin/python3.12
+
+    $ ln -s ~/bin/python /usr/bin/python3.12
+
```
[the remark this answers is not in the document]
jhernand: Done.

manifests/kustomization.yaml
jhernand: `sclorg` is Red Hat Software Collections: https://developers.redhat.com/products/red-hat-software-collections/overview . The benefit, if any, is that it is based on a RHEL image. I don't have any objection to use the stock image.

manifests/publisher.yaml
jhernand: It is, will be fixed in the next version of the patch.

manifests/kustomization.yaml
jhernand: No objection, done.

README.md
jhernand: The very minimum is to check-out the code and then run `go build`, or `podman build`, or `ginkgo run`. I try to set this up so there are no surprises in the usage of the most common Go tools.

However there are situations where those tools aren't enough. For example, to re-generate the gRPC code from the API specification it is necessary to fetch the API code (from the `fulfillment-api` repo), checkout a particular tag, generate the right `buf.gen.yaml` file, and then call `buf`. Typically that will be in a (usually awful) Makefile or shell script, but in this case it is in a better structured Python script. The instructions I wrote are a bit complicated, yes, but once done the result is a comfortable environment, where you can just run `dev generate` to do all that. I would very much like to keep that in the repository.

manifests/kustomization.yaml
larsks: This is fine; I just wasn't familiar with software collections.

README.md
```diff
@@ -1,8 +1,199 @@
...
+Later you will configure your environment (via the `.envrc` file ) to use that version.
+
+### Make sure that you have a PostgreSQL 13 or newer database available
+
+Once you have installed the server you will need to create a user and a database for the service. In RHEL 9, for
+example, with the root user:
+
+    # dnf install -y postgresql-server
+    # postgresql-setup --initdb
+    # systemctl enable postgresql --now
```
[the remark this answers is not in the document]
jhernand: I believe that for development running the system service is easier. Would it be enough it I mention in the document that you can also use any other procedure to install PostrgreSQL, including running it in a container?

Containerfile
jhernand: `fulfillment-service` it is.

README.md
jhernand: Looks nice. Can I do that change later, once I have time to try and learn `uv`? Care to open a ticket about that?

mhrivnak: I appreciate the goal of automating common developer tasks. But 1800 lines of custom python is not necessarily an obvious improvement over a Makefile. Again to the idea that we're doing a PoC here and keeping the implementation minimal, would you mind if we just leave out all of this python for now? If/when we need something to serve this role, we can have a discussion about `make`, `task` (which is designed in/for Go: [website](https://taskfile.dev/) ), and other options.

I noted that @larsks recommended `uv` and/or `pipenv` **for a python project**, which this is not. I'd love to not commit us to requiring python and a whole associated environment unless we really agree that it's the best approach.

In the meantime if we prove out what really needs automation in the form of shell scripts, or even a Makefile, we can all agree to not make them hacky. :wink:

README.md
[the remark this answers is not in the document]
mhrivnak: +1 for using a container. We can replace pretty much all of this with a one-liner:

```
podman run -d --name postgresql_database -e POSTGRESQL_USER=user -e POSTGRESQL_PASSWORD=pass -e POSTGRESQL_DATABASE=db -p 5432:5432 
```

https://quay.io/repository/fedora/postgresql-16

That's also more portable for people who, er, might not use linux for development. :scream:

README.md
mhrivnak: I generally agree that this is a lot to read through and learn.

I would suggest that we not necessarily recommend/prescribe using `envdir`, the instructions for which take up a lot of this document and which is probably new to a lot of people, and instead just specify what environment variables need to be set. I had to read the whole envdir section to figure out "oh it's just going to automatically set some environment variables, and I could do that a more familiar way if I want".

And per the other thread, if we likewise stop requiring all of the python setup, that'll help a lot with keeping this doc focused on how to build and run the service.

README.md
jhernand: The new version of the patch doesn't include the Python code.

README.md
jhernand: The new version of the patch doesn't contain those instructions.

larsks (approved):
Looks good to me.

mhrivnak (approved):

PR #11, opened 2025-03-19
Update to version 0.0.3 of the API

jhernand:
The more relevant changes in the new version of the API are the following:

- Replace copies of packages with buf dependencies.
- Add endpoint to retrieve admin Kubeconfig.

The patch adds an implementation of the admin Kubeconfig that returns always an empty Kubeconfig.

## Summary by CodeRabbit

- **New Features**
  - Introduced new ways to access cluster configuration details, streamlining the process of retrieving important setup information.
  - Added methods to retrieve Kubernetes configuration data based on cluster identification.

- **Chores**
  - Improved dependency management by updating version settings and adjusting module configurations for enhanced system stability.
  - Disabled specific modules to optimize the management of dependencies.


internal/servers/clusters_server.go
```diff
@@ -147,6 +148,71 @@ func (s *ClustersServer) Get(ctx context.Context,
...
+		)
+		err = grpcstatus.Errorf(grpccodes.Internal, "failed to get cluster with id '%s'", request.ClusterId)
+		return
+	}
+	if !ok {
+		err = grpcstatus.Errorf(grpccodes.NotFound, "cluster with id '%s' not found", request.ClusterId)
+		return
+	}
+
+	// TODO: Do we want to store the Kubeconfig in the database, or fetch it from somewhere else?
```
larsks, line 176: I think we're going to want to fetch it from the management cluster rather than storing it in the database (I'd like to avoid increasing the number of places in which we're storing persistent credentials).

internal/servers/clusters_server.go
```diff
@@ -147,6 +148,71 @@ func (s *ClustersServer) Get(ctx context.Context,
...
+			slog.String("cluster_id", request.ClusterId),
+			slog.String("error", err.Error()),
+		)
+		err = grpcstatus.Errorf(grpccodes.Internal, "failed to get cluster with id '%s'", request.ClusterId)
+		return
+	}
+	if !ok {
+		err = grpcstatus.Errorf(grpccodes.NotFound, "cluster with id '%s' not found", request.ClusterId)
+		return
+	}
```
larsks, line 206: This is all a duplicate of the code in `GetKubeConfig`. Can we factor it out into a common helper function?

larsks (approved):

internal/servers/clusters_server.go
jhernand: Yes, I will.

jhernand: Done.

internal/servers/clusters_server.go
jhernand: Yes. @coderabbitai open an issue in this project explaining that we need to fetch the admin Kubeconfig for clusters in such a way that we don't store the credentials persistently in the database.

PR #21, opened 2025-03-31
Add database listener and notifier

jhernand:
This patch adds utilities to simplify the usage of PostgreSQL's LISTEN/NOTIFIY mechanism. The `Listener` object takes care of keeping the connection to the database open, even if the database is restarted.

Both the `Listener` and `Notifier` object use protocol buffers messages, and they take care of encoding them so that they can be sent as payload of the LISTEN/NOTIFY mechanism.

## Summary by CodeRabbit

- **New Features**
	- Introduced a robust PostgreSQL notification system allowing the application to both send and listen for database notifications with configurable callbacks and error handling.
- **Tests**
	- Added comprehensive tests to ensure reliable creation, operation, and error handling of the new notification features.

### DIFF
--- a/internal/database/database_listener.go
+++ b/internal/database/database_listener.go
@@ -1,0 +1,406 @@
+/*
+Copyright (c) 2025 Red Hat Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with the
+License. You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific
+language governing permissions and limitations under the License.
+*/
+
+package database
+
+import (
+	"context"
+	"encoding/base64"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"log/slog"
+	"slices"
+	"time"
+
+	"github.com/jackc/pgx/v5"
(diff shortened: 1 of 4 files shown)


internal/database/database_listener.go
```diff
@@ -0,0 +1,406 @@
...
+	logger           *slog.Logger
+	url              string
+	channel          string
+	readyCallbacks   []ListenerReadyCallback
+	payloadCallbacks []ListenerPayloadCallback
+	waitTimeout      time.Duration
+	retryInterval    time.Duration
+}
+
+// Listener knows how to liste for notifications using PostgreSQL's `listen` mechanism.
```
adriengentil, line 51:
```suggestion
// Listener knows how to listen for notifications using PostgreSQL's `listen` mechanism.
```

internal/database/database_listener.go
```diff
@@ -0,0 +1,406 @@
...
+		if err != nil {
+			l.logger.InfoContext(
+				ctx,
+				"Failed to close connection",
+				slog.Any("error", err),
+			)
+		}
+		l.conn = nil
+	}
+	conn, err := pgx.Connect(ctx, l.url)
```
adriengentil, line 256: is this function takes a connection from a pool? If no, I think it might be worth to have one if we need to often open/close connections.

adriengentil (commented):
LGTM! just one question

internal/database/database_listener.go
jhernand: The way that this works is that it opens a connection and keeps it open till it fails or till the wait timeout expires. That wait timeout is 5 minutes by default. It will never use more than one connection simultaneously. So I think a pool is not necessary or convenient for this.

adriengentil (approved):

PR #29, opened 2025-04-15
Get cluster Kubeconfig from hub

jhernand:
This patch implements the method that retrieves the admin Kubeconfig for a cluster. To do so it makes the following changes:

- Add a private table, DAO and gRCP server for clusters, where we will store additional information about clusters, in particular the identifier of the hub where the cluster lives.

- Add a tool to simplify use of `jq`. This is used to extract information from the Kubernetes `ClusterOrder` and `HostedCluster` objects without having to import the corresponding Go type definitions.

- Small refactor to the files where we have the Kubernetes GVKs (group/version/kind) to make them public and usabe from multiple packages, in particular from the controller and also from the gRPC clusters server.

- Implement the logic to fetch the Kubeconfig.

Getting the Kubeconfig requires the following steps:

1. Fetch the internal cluster details from the database to obtain the identifier of the hub where the cluster and the identifier of the original cluster order.

2. Fetch the description of the hub to obtain the Kubeconfig to connect to the hub, as well as the namespace where `ClusterOrder` Kubernetes objects are created.

3. Fetch from the hub the `ClusterOrder` Kubernetes object that corresponds to the identifier of the original cluster order.

4. Extract from the `ClusterOrder` the location (namesspace and name) of the `HostedCluster` object.

5. Fetch the `HostedCluster` object from the hub.

6. Extract from the `HostedCluster` object the location of the Kubeconfig secret.

7. Fetch the Kubeconfig secret from the hub.

This means that the identitity used by the server (the Kubeconfig of the hub) needs to be granted permission to read `ClusterOrders` in the namespace of the hub, permission to read `HostedClusters` in all namespaces and permssion to read secrets also in all namespaces.

Note that this change is on top of #25, when reviewing focus only on the last commit.


internal/jq/jq_query.go
```diff
@@ -0,0 +1,217 @@
...
+// methods of the Tool type.
+type Query struct {
+	logger    *slog.Logger
+	source    string
+	variables []string
+	code      *gojq.Code
+}
+
+// Evaluate evaluates the query with on the given input. The output should be a pointer to a variable where the result
+// will be stored. Optional named variables can be passed.
```
larsks, line 38:
```suggestion
 // Evaluate evaluates the query on the given input. The output should be a pointer to a variable where the result
// will be stored. Optional named variables can be passed.
```

internal/jq/jq_query.go
```diff
@@ -0,0 +1,217 @@
...
+func (q *Query) evaluate(input any, output any, variables []any) error {
+	// Check that the output is a pointer:
+	outputType := reflect.TypeOf(output)
+	if outputType.Kind() != reflect.Pointer {
+		return fmt.Errorf("output should be a pointer, but it is of type '%T'", output)
+	}
+
+	// The library that we use expects the input to bo composed only of primitive types, slices and maps, but we
+	// want to support other input types, like structs. To achieve that we serialize the input and deserialize it
+	// again.
```
larsks, line 67: I am a little confused about how the `convert` function works; if something is a struct, it's going to fall through to the default case, `return q.clone(input, output)`, so if there was a struct on input we're still going to have a struct in the output. How does this achieve the stated goal?

It also looks like when calling `convert` with a `*any` in the second parameter -- as we do here -- that it's effectively a no-op, since in all cases we have:

```
		case *any:
			*output = input
```

larsks (approved):

internal/jq/jq_query.go
jhernand: The result is achieved because of the temporary variable `tmp`, whose type is `any`. As a result the input is converted into whatever the JSON unmarshaling generates. For an integer it will be `int`, for a boolean will be `bool`, and for a struct it will be a `map[string]any`.

On top of that, which is implemented by the  `clone` function, the `convert` function is intended as a shortcut for some specific types, like integers, strings, etc, to avoid the theoretical overhead of JSON marshal/unmarshal. I don't have number to back that theoy. I acknowledge that it is a bit complicated, and not strictly needed. I will remove it.

jhernand: In the new version of the patch the `convert` function is gone.

PR #31, opened 2025-04-25
Automatically create private clusters orders and clusters

jhernand:
Currently the objects containing private information about cluster orders and clusters need to be created explicitly by the components that use them. To simplify that this patch changes the server so that it creates them automatically when the corresponding public object is created. To do so it uses database triggers. The resulting SQL files are much larger, so the patch also splits them into separate files for the public and private schemas.


larsks (approved):
This seems reasonable. 

For my education: Why are we maintaining both public and private versions of these resources?

jhernand:
> For my education: Why are we maintaining both public and private versions of these resources?

The reason to do this is that there are parts of the information that are of no interest to the user of the API. For example, the user of the API doesn't want (and shouldn't) know what is the hub where a cluster lives. On the other hand that information is critical for the system. Having it in a place where only the system can read/write it is convenient.

PR #32, opened 2025-04-25
Add workflow to publish image

jhernand:
This patch adds a GitHub workflow that publishes the container image to the GitHub container image registry for each push to the main branch and for each version tag.

In addition to the `latest` tag, for pushes to the main branch add a tag containing the short commit hash, for example:

    ghcr.io/innabox/fulfillment-service:4719ed7

And for version tags like `v0.0.1` it will add the that version tag, for example:

    ghcr.io/innabox/fulfillment-service:v0.0.1


tzumainn (approved):
Looks good!

PR #33, opened 2025-04-25
Add host classes server

jhernand:
This patch adds the implementation of the `HostClasses` service.


adriengentil:
We don't need this change anymore, right?

jhernand:
@adriengentil I believe we still want this, although it needs to be rebased. I am on it.

PR #34, opened 2025-04-30
Implement filter using CEL

jhernand:
The `List` method used to retrieve lists of objects has a `filter` parameter intended for filtering the result sets. Currently that paraemeter is ignored by the server. That means that users don't have any way to search for specific objects other than retrieving the complete list and then implementing their own filtering. That isn't currently an issue because the number of objects is very small, but will become an inconvenience and a problem as soon as the number of objects grows.

This patch adds support for that `filter` parameter.

The server will assume that the `filter` parameter contains a [CEL language](https://cel.dev) boolean expression that will be evaluated for each potentially returned object. Only if the expression evalutes to `true` will the object be returned by the server.

The expression can use the built-in `this` variable that represents the object on which the expression is being evaluated. For example, if the user wants to find the templates that have `OpenShift` inside the description:

```java
this.description.contains('OpenShift')
```

From `this` all the fields of the object are available, using the field names as defined in the protocol buffers specification of the type. For example, for a cluster template all these are valid:

```java
this.id
this.title
this.description
this.metadata.creation_timestamp
this.metadata.deletion_timestamp
```

In addition to `this` there is also a `now` built-in variable that contains the current date. That is intended to be able to compare dates with the current date.

The expression can also use a subset of the operators available in the CEL language. For example, to find all the tamplates that have an identifier starting with `ocp` and don't contain `4.18` in the title:

```java
this.id.startsWith('ocp') && !this.title.contains('4.18')
```

To check for presence of optional fields the `has` macro can also be used. For example, to find all the templates that have a description:

```java
has(this.description)
```

The server uses the CEL parser to translate the `filter` parameter into a CEL abstract syntax tree, and then translates that tree into a SQL `where` statement matching our database schema. Thus the actual evaluation of the expression is performed by the database. This means that the server only uses the CEL parser; it doesn't build or execute CEL programs.

A consequence of not building or executing CEL programs is that the server needs to impose restrictions on the CEL features that can be used. Some of these restrictions are because of the nature of SQL `where` statements, others are because we just didn't implement the translation. For example, CEL provides a `startsWith` method for strings that checks if a string starts with a given prefix:

```java
this.id.startsWith('ocp')
```

The server translates that to a SQL `like` construct like this:

```sql
id like 'ocp%'
```

Because of that it needs to impose a limit limits on the type of the argument of the `startWith` method: in CEL it can be any expression that evaulates to a sring, but the server restricts it to string literals.

The same happens for the `endsWith` and `contains` functions.

Another example is the `in` operator:

```java
this.id in ['my', 'your']
```

In CEL the values can be any expression, but the server restrict them to literals so that it can translate the filter into the corresponding SQL `in` syntax:

```sql
id in ('my', 'your')
```

Currently the following CEL features are supported:

- `null` literal.
- Boolean literals.
- Integer literals.
- String literals.
- Field selection using the dot notation.
- Parenthesis.
- `has` macro.
- `+` arithmetic operator.
- `-` arithmetic operator.
- `*` arithmetic operator.
- `/` arithmetic operator.
- `%` arithmetic operator.
- `==` comparison operator.
- `!=` comparison operator.
- `>` comparison operator.
- `>=` comparison operator.
- `<` comparison operator.
- `<=` comparison operator.
- `!` logical operator.
- `&&` logical operator.
- `||` logical operator.
- `in` operator.
- `contains` function.
- `startsWith` function.
- `endsWith` function.

Anything outside this list of features isn't supported and will be rejected by the server.


mhrivnak:
It's challenging to review a PR this big without at least a little bit of documentation.
- What does the feature do exactly?
- How would someone use it? (including examples please)
- Why do we need it? (in this case, why do we need something so generic vs. targeted filtering?)

Specific questions about this PR:
- what are the boundaries of what someone can do with it? How complex can it get? Is there any way for a user to traverse rows or tables?
- Is CEL->SQL a common approach for grpc APIs?

jhernand:
> It's challenging to review a PR this big without at least a little bit of documentation.
> 
> * What does the feature do exactly?
> * How would someone use it? (including examples please)
> * Why do we need it? (in this case, why do we need something so generic vs. targeted filtering?)

I tried to answer these questions in the updated commit message.

> Specific questions about this PR:

> * what are the boundaries of what someone can do with it? How complex can it get? Is there any way for a user to traverse rows or tables?

The boundary is very simple: the expression is evaluated with a single `this` variable that contains the object being filtered. There is no way to get out of that. The filter can't access other objects, or other tables.

> * Is CEL->SQL a common approach for grpc APIs?

I am not aware of any project that uses this approach.

larsks (approved):
This seems like a lot to get "filter deleted objects", but I guess it may prove useful in the future if the number of objects grows large.

PR #35, opened 2025-04-30
Add cluster reconciler

jhernand:
This adds a new reconciler that watches the _Cluster_ objects and updates the Kubernetes _ClusterOrder_ accordinlgy. In particular it copies the definitions of the node sets into the `nodeRequests` spec of the _ClusterOrder_.

In order to do that change the patch also moves the cache of hub connections to a separate component, so that it can be used by both the cluster order and cluster reconcilers.


larsks (approved):

jhernand:
@larsks this is now rebased without the host classes server.

tzumainn (approved):
Seems reasonable! Quick question - are the `host_class_` files required as part of the reconciler change?

jhernand:
> Seems reasonable! Quick question - are the `host_class_` files required as part of the reconciler change?

The aren't used. But the update to version 0.0.7 of the API produces them because the `HostClass`  type has been added to the API in that version. We can remove or rename them later if we decide to remove or rename `HostClass`.

PR #36, opened 2025-05-04
Add `Authorization` header to Envoy CORS filter

jhernand:
Currently the CORS configuration of the Envoy proxy doesn't explicitly list `Authorization` as an allowed header. This seems to work fine with Firefox, but not with Chrome. This patch adds it.


tzumainn (approved):
Confirm that it fixes the CORS issues on Chrome. Thanks!

PR #39, opened 2025-05-07
Revert "Create non-expiring token for controller service account (#37)"

jhernand:
This reverts commit 69a3ef5eb4f81b5d2b04b725ff1c123b81690e9d.

This is necessary because the tokens created using this mechanism don't have an `iat` claim. For example:

```json
{
  "iss": "kubernetes/serviceaccount",
  "kubernetes.io/serviceaccount/namespace": "innabox",
  "kubernetes.io/serviceaccount/secret.name": "controller-token",
  "kubernetes.io/serviceaccount/service-account.name": "controller",
  "kubernetes.io/serviceaccount/service-account.uid": "abec75cc-021a-4e2f-8a3f-d6e37a28c650"
  "sub": "system:serviceaccount:innabox:controller"
}
```

And our JWT authenticator rejects that with the following error message:

```json
{
  "time": "2025-05-07T17:31:25Z",
  "level": "ERROR",
  "msg": "Watch failed",
  "error": {
    "message": "rpc error: code = Unauthenticated desc = Bearer token doesn't contain required claim 'iat'",
    "goroutine": 15,
    "stack": [
      {
        "function": "github.com/innabox/fulfillment-service/internal/controllers.(*Reconciler[...]).watchLoop",
        "source": "/source/internal/controllers/reconciler.go:320"
      }
    ]
  }
}
```

Instead of that non-expiring token we will change the gRPC client so that it always reads the latest token from the file.

### DIFF
--- a/manifests/controller/deployment.yaml
+++ b/manifests/controller/deployment.yaml
@@ -26,21 +26,11 @@ spec:
         app: fulfillment-controller
     spec:
       serviceAccountName: controller
-
-      volumes:
-
-      - name: controller-token
-        secret:
-          secretName: controller-token
-
       containers:
 
       - name: controller
         image: fulfillment-service
         imagePullPolicy: IfNotPresent
-        volumeMounts:
-        - name: controller-token
-          mountPath: /run/secrets/token
         env:
         - name: NAMESPACE
           valueFrom:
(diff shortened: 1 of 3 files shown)


jhernand:
The alternative is #38 .

PR #45, opened 2025-05-21
Don't use int32 in unstructured objects

jhernand:
We use the `Unstructured` type of the Kubernetes client package to interact with the `ClusterOrder` object. That type doesn't support the `int32` type, but we use it because it is what the gRPC returns. To avoid the issue we need to explicitly convert it to `int64`.


jhernand:
Replaced by #46, which includes this and other fixes.

PR #47, opened 2025-05-22
Implement the `GetPassword` method

jhernand:
This patch implements the `GetPassword` and `GetPasswordViaHttp` method of the clusters service. These methods return the admnistrator password of the cluster.


jhernand:
Note that this will fail some of the tests because the API changes (in https://github.com/innabox/fulfillment-api/pull/29) need to be merged first.

internal/servers/clusters_server.go
```diff
@@ -387,6 +387,237 @@ func (s *ClustersServer) getKubeconfig(ctx context.Context, clusterId string) (r
...
+		return
+	}
+	response = &httpbody.HttpBody{
+		ContentType: "text/plain",
+		Data:        []byte(password),
+	}
+	return
+}
+
+func (s *ClustersServer) getPassword(ctx context.Context, clusterId string) (result string, err error) {
```
tzumainn, line 415: A fair amount of logic in this function seems pretty similar to that in `getKubeConfig` - would it make sense to consolidate that code?

jhernand: Yes, I will try.

tzumainn (approved):

PR #48, opened 2025-05-23
Add Kustomize overlays for OpenShift and Kind

jhernand:
Currently we can only deploy to _OpenShift_ because we depend on routes and the service signer. This makes it difficult to deploy on other Kubernetes clusters, in particular in _Kind_, which we would like to use for testing pull requests. To make that possible this patch introduces Kustomize overlays for OpenShift and Kind. The changes required to do the that are the following:

1. Move most of the manifests to a new `bases` directory.

2. Create the `overlays/openshift` and `overlays/kind` directories, with their corresponding _kustomization_ files.

3. Move the creation to the OpenShift route from `bases` to `overlays/openshift`.

4. Add to the _Kind_ overlay a patch that changes the API service to use a node port, so that we can use it without installing an ingress or gateway API controller.

5. Change the service so that it uses a certificate generated by _cert-manager_ instead of depending on the OpenShift signer.

6. Change the gRPC client so that it automatically trusts the Kubernetes CA and the OpenShift service signer CA.

7. Change the controller so that it connects to the API using a DNS name without the namespace (as that is the name used by the certificate generated by _cert-manager_).

8. Change the OpenShift route so that it uses `passthrough` instead of `reencrypt` (this solves the gRPC issue with ALPN).

After these changes the user can deploy the application to OpenShift with this command:

```shell
$ oc apply -k manfiests/overlays/openshift
```

And to Kind with this command:

```shell
$ kubectl apply -k manfiests/overlays/kind
```

In OpenShift the application will be available in the same route. In Kind it will be available in `localhost:8000`.


larsks (approved):
Looks good.

Note that rather than creating a Route in the openshift overlay, it is possible to  create an Ingress in `base`: this will work great on vanilla kubernetes if there is an ingress service available, and OpenShift will automatically create a Route for you from the Ingress resource. See in particular "[Creating a route through an Ingress object](https://docs.redhat.com/en/documentation/openshift_container_platform/4.17/html/networking/configuring-routes#nw-ingress-creating-a-route-via-an-ingress_route-configuration)", which lists the annotations you can add to an Ingress resource to set Route-specific configuration.

The only tricky part is that an Ingress requires you to set an explicit hostname, which means the resource would need patching in any case.

PR #52, opened 2025-06-04
Merge public and private objects

jhernand:
Currently we store public and private data in different objects, endpoints and tables. For example, for cluster orders we have a `fulfillment.v1.ClusterOrder` object that has the fields that are relevant for the user of the API, like the `template_id` and `template_parameters`, and a `private.v1.ClusterOder` object that has the fields that are only relevant for the system, like the `hub_id`. This separation means that internal components often need to fetch and then save both the public and private objects, which is cumbersome and makes it possible to have errors where one is saved and the other isn't. To avoid that this patch changes the private objects so that they contain all the fields of the public objects plus the private ones. Note that it isn't possible to extend or embed a protobuf object into another, so the private object just redeclares all the fields of the public object. The storage is also unified into a single table, and all the tables, triggers and functions used to synchronize the public and private objects is removed. Internal components are changed so that they only use the private objects. Clients of the public API aren't affected: the server automatically translates the private objects to the corresponding public objects, and the other way around.


jhernand:
@adriengentil  this still needs to be merged because it is about the infrastructure to have the public and private parts of objects in the same place. The API pull request that you mention is mostly independent of this.

PR #59, opened 2025-07-04
Set default node sets from template

jhernand:
This patch changes the clusters server so that it sets the value of the `spec.node_sets` of clusters from the template. This used to be part of the operator, because the node sets where not part of the cluster order, but now that `ClusterOrder` and `Cluster` have been merged the fulfillment service can do it.

The patch also introduces validations to ensure that users can't create node sets that aren't defined in the template, or use host classes that don't match the template.

### DIFF
--- a/internal/servers/clusters_server.go
+++ b/internal/servers/clusters_server.go
@@ -18,8 +18,10 @@ import (
 	"errors"
 	"fmt"
 	"log/slog"
+	"sort"
 	"sync"
 
+	"golang.org/x/exp/maps"
 	"google.golang.org/genproto/googleapis/api/httpbody"
 	grpccodes "google.golang.org/grpc/codes"
 	grpcstatus "google.golang.org/grpc/status"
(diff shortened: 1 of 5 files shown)


internal/servers/clusters_server.go
```diff
@@ -152,6 +164,107 @@ func (s *ClustersServer) Get(ctx context.Context,
...
+		err = grpcstatus.Errorf(grpccodes.InvalidArgument, "template '%s' doesn't exist", templateId)
+		return
+	}
+	if template.GetMetadata().HasDeletionTimestamp() {
+		err = grpcstatus.Errorf(grpccodes.InvalidArgument, "template '%s' has been deleted", templateId)
+		return
+	}
+
+	// Check that all the node sets given in the cluster correspond to node sets that exist in the template:
+	templateNodeSets := template.GetNodeSets()
```
adriengentil, line 199: are we aligned on the structure of the template node sets? What ansible pushes today is a bit different: https://github.com/innabox/cloudkit-aap/blob/main/collections/ansible_collections/cloudkit/service/plugins/filter/find_template_roles.py#L167-L189

adriengentil (approved):

PR #68, opened 2025-07-10
Add IT_KEEP_KIND to preserve kind cluster used for integration tests

jhernand:
This patch adds a new `IT_KEEP_KIND` environment variable that can be used to preserve the kind cluster that is created for integration tests. This is convenient in the development environment, as it allows the developer to inspect the cluster, in particular the logs of the services and the contents of the database.


jhernand:
Duplicate of #66 , sorry.

PR #73, opened 2025-07-10
Add public clusters API integration tests

jhernand:
This patch adds a set of integration tests for the public clusters API, including the methods to get the kubeconfig of a cluster.


jhernand:
Note that this depends on #68, #69, #70, #71.

PR #89, opened 2025-07-31
Replace custom auth with Authorino

jhernand:
This commit replaces the custom authentication and authorization implementation with Authorino, a Kubernetes-native authorization service. This change enables more flexible and configurable authentication and authorization capabilities for future requirements.

Key changes:
- Remove custom JWKS authentication function (~900 lines)
- Remove custom rules-based authorization function (~300 lines)
- Remove CEL expression evaluation logic
- Remove gRPC authorization interceptor and related infrastructure
- Add new external authentication function that reads from X-Subject header
- Add Authorino operator deployment and AuthConfig manifests
- Update Envoy configuration to integrate with Authorino
- Add OPA dependency for policy evaluation
- Simplify server startup by removing authorization interceptor

Benefits:
- Enables flexible authentication methods (OIDC, API keys, mTLS, etc.)
- Supports complex authorization policies via OPA/Rego
- Reduces maintenance burden of custom auth code
- Leverages proven, community-supported authorization service
- Allows runtime policy changes without service restarts

The new architecture uses Authorino as an external authorization service that validates requests before they reach the fulfillment service, with the service receiving authenticated user information via the X-Subject header.


manifests/base/service/authconfig.yaml
```diff
@@ -0,0 +1,118 @@
...
+      kubernetesTokenReview:
+        audiences:
+        # TODO: This is the default audience for the Kubernetes API server. We should probably create a custom audience,
+        # for example `fulfillment-api`, but then the users will need to create tokens for that audience. For example:
+        #
+        # $ kubectl create token -n innabox --audience=fulfillment-api client
+        #
+        # It is not clear to me if we can also use that audience for the tokens generated automatically for the service
+        # accounts of other pods, for example for the `controller` pods. It may be possible using a projected service
+        # account token, something like this:
```
adriengentil, line 32: FYI, AAP enforce projected service account when creating pods that runs the playbooks. So I think it's ok.

https://github.com/innabox/cloudkit-aap/blob/main/collections/ansible_collections/cloudkit/config_as_code/roles/aap/vars/controller.yml#L177-L199

adriengentil (approved):

jhernand:
@jkary this may have a rather significant impact in the installation, as we will need to install the Authorino operator. So I won't merge it unless you tell me it is OK for you.

tzumainn:
> @jkary this may have a rather significant impact in the installation, as we will need to install the Authorino operator. So I won't merge it unless you tell me it is OK for you.

Given this, would it make sense to wait until after the August 12th deadline?

jhernand:
> Given this, would it make sense to wait until after the August 12th deadline?

No objection, I will move to draft to make it clear that it shouldn't be merged yet.

tzumainn (approved):

PR #102, opened 2025-09-03
add vm template services

adriengentil:
- **Update fulfillment-api version to pull VM definitions**
- **Add VirtualMachine and VirtualMachineTemplate private types and services**
- **Fix VirtualMachineTemplate event handling and add comprehensive tests**


internal/utils/template_parameters.go
```diff
@@ -0,0 +1,199 @@
...
+		actualParameters[templateParameterName] = actualParameter
+	}
+
+	return actualParameters
+}
+
+// ClusterTemplateAdapter adapts ClusterTemplate to the Template interface
+type ClusterTemplateAdapter struct {
+	*privatev1.ClusterTemplate
+}
```
adriengentil, line 160: if we used this template pattern more, it would be good to have only one protobuf object

internal/servers/private_virtual_machine_templates_server.go
```diff
@@ -0,0 +1,106 @@
+/*
+Copyright (c) 2025 Red Hat Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with the
+License. You may obtain a copy of the License at
+
```
adriengentil, line 6: not sure if it's worth to add test to the private API, they are similar to the public one. WDYT @jhernand ?

internal/utils/template_parameters.go
jhernand: I think it could be good to have a single "TemplateParameter" type, as it is unlikely that we will want to change that in a way that they are different for clusters, VMs, etc. But I wouldn't try to have a single "Template" type, as it is very likely that we we will want to different things to the "ClusterTemplate" and the "VirtualMachineTemplate", and having a unique one would complicate that. For example, we would probably want to add "disks" and "nics" field (fields, not parameters) to the virtual machine template type, and those don't make sense for clusters.

internal/servers/private_virtual_machine_templates_server.go
jhernand: It is worth to add to the private API tests for the things that are only (or will be only) in the private API. For example, the capability to update a cluster template will be only in the private API because users will not have permission to do it. Currently it is still part of the public API, and disabled via permissions, but it should be removed at some point. When we do that it would be nice to have a test that verifies that the private API can do the updates.

PR #110, opened 2025-09-16
Add HostPool and Host types and services

tzumainn:
* create database schemas and migrations
* create public/private API and servers
* update events to support HostPool and Host
* update API authorization rules


tzumainn:
@jhernand I tested by starting the grpc server and verifying that grpcurl List commands would return the appropriate things; are there other tests you might suggest I do?

PR #117, opened 2025-09-18
Add keep alive to watch streams to prevent proxy disconnections

jhernand:
This change introduces keep alive functionality for event watch streams in an attempt to avoid connections being forcibly closed by proxies, particularly Envoy in our deployment environment. We have been experiencing intermittent connection issues that we haven't been able to fully identify, but they appear to be related to idle connection timeouts at the proxy layer.

The implementation adds a new `keep_alive` field to the `EventsWatchRequest` protobuf message, allowing clients to specify an interval for periodic keep alive events. When configured, the server will send empty events of type `EVENT_TYPE_KEEP_ALIVE` at the specified interval to maintain connection activity. This should significantly reduce the chances of proxy-initiated connection drops during periods of low event activity.

The reconciler builder now includes a `SetKeepAliveInterval` method that allows fine-grained control over this behavior, with a sensible default of one minute and the ability to disable keep alive entirely by setting the interval to zero. The reconciler properly handles and filters out keep alive events to avoid unnecessary processing overhead.


internal/controllers/reconciler.go
```diff
@@ -338,37 +355,63 @@ func (c *Reconciler[O]) watchLoop(ctx context.Context) {
...
+	}
+	stream, err := c.eventsClient.Watch(ctx, request)
 	if err != nil {
 		return err
 	}
+	c.logger.DebugContext(
+		ctx,
+		"Started watching events",
+		slog.String("filter", request.GetFilter()),
+		slog.Duration("keep_alive", c.keepAliveInterval),
```
adriengentil, line 372: `keepAliveInterval` has no use in the reconciler, right? It's just here for logs?

adriengentil: ah, forget, it's passed in the request :man_facepalming:

adriengentil (approved):

PR #119, opened 2025-09-19
Add host and host pool reconcilers

tzumainn:
Adds host and host pool reconcilers, following the pattern used for clusters and virtual machines.


internal/controllers/host/host_reconciler_function.go
```diff
@@ -0,0 +1,307 @@
...
+}
+
+func (t *task) update(ctx context.Context) error {
+	// Set the default values:
+	t.setDefaults()
+
+	// TODO: add host state field so we can see if the update is needed
+	//if t.host.GetStatus().GetState() != privatev1.HostState_HOST_STATE_PROGRESSING {
+	//	return nil
+	//}
```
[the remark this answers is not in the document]
jhernand: I think it is fine.

internal/controllers/host/host_reconciler_function.go
```diff
@@ -0,0 +1,307 @@
...
+
+	// Select the hub:
+	err := t.selectHub(ctx)
+	if err != nil {
+		return err
+	}
+
+	// Save the selected hub in the private data of the host:
+	// Note: Host doesn't have a hub field in status, we'll track this differently
```
jhernand, line 144: How? If we don't remember the hub then the next time the user does an update or delete we may select a different hub, and we will end up with two `Host` CRD instances created in different hubs. Is that OK? Furthermore, those two instances may diverge over time, and then we will have no way to know which one is the good one.

tzumainn: Whoops - this is something I meant to ask about; I was comparing this code to clusters, and I'm not entirely sure where the hub data is specified in fulfillment-api. Is there something I'm missing there?

jhernand (approved):

PR #128, opened 2025-10-02
Make tenancy logic mandatory

jhernand:
Currently the storage layer uses the empty tenancy logic by default. This is to avoid setting it explicily in unit tests that aren't about tenancy. A potential side effect of that is that if some code path forgets to explicitly set it can result in a security vulnerability. To avoid that this patch changes the storage layer so that the the tenancy logic is required, and is explicitly set in all the places where it is needed.

Note that this doesn't change the behaviour of the system, as the tenancy logic was already explicitly set when the server was started.


computate:
Thanks for making these changes to tenancy by default.

PR #147, opened 2025-10-28
Use `main` instead of `latest` as image tag

jhernand:
Some time ago we changed the job that publishes images. As a result we now use `main` (the name of the branch) instead of `latest`. We need to change the manifests accordingly.


trewest (approved):
/lgtm

PR #173, opened 2025-11-04
Select for update before updating DB

adriengentil:
Make sure that what we update is what was selected before.

I did not face any issues related to that, but I think it can lead
to weird behaviors.

### DIFF
--- a/internal/database/dao/generic_dao.go
+++ b/internal/database/dao/generic_dao.go
@@ -476,11 +476,11 @@ func (d *GenericDAO[O]) Get(ctx context.Context, id string) (result O, err error
 		return
 	}
 	defer tx.ReportError(&err)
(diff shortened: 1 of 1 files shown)


internal/database/dao/generic_dao.go
```diff
@@ -518,6 +518,9 @@ func (d *GenericDAO[O]) get(ctx context.Context, tx database.Tx, id string) (res
 		d.table,
 		filterBuffer.String(),
 	)
+	if forUpdate {
+		sqlBuffer.WriteString(" FOR UPDATE")
```
jhernand: Can you change the SQL code to lowercase to match the rest of SQL code in this file?

jhernand (commented):
Looks good, just a small ask.

PR #175, opened 2025-11-04
Prevent update loop between server and controller

adriengentil:
Align with https://github.com/innabox/fulfillment-service/pull/174 to
prevent a potential update loop between the server and the controller.


adriengentil:
@tzumainn it would be great if you can confirm this change doesn't cause a regression on your side :pray:

larsks (approved):

PR #182, opened 2025-11-13
Label VirtualMachine with tenant ID

adriengentil:
Since virtual machines belonging to a tenant will be created in the same
UDN network, we need to pass tenant's information to VirtualMachine CR.

This change labels the VirtualMachine CR with the tenant ID. The ID is
generated from the aggregation of the tenants available in fulfillment
metadata, along with a hash to comply with Kubernetes naming and avoid
collisions.

I could have just hashed the tenant name, but I think it would help with
maintenance and debugging to have tenant's name in the ID.

### DIFF
--- a/internal/controllers/vm/vm_reconciler_function.go
+++ b/internal/controllers/vm/vm_reconciler_function.go
@@ -175,6 +175,7 @@ func (t *task) update(ctx context.Context) error {
 		object.SetGenerateName(objectPrefix)
 		object.SetLabels(map[string]string{
 			labels.VirtualMachineUuid: t.vm.GetId(),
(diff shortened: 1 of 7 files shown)


jhernand (commented):
This seams reasonable to me. But depending on how you are going to use that label it may be better to add an annotation containing the complete list of tenants.

I'd also suggest, if you need the label, to use the same algorithm that Kubernetes uses to generate the names of the replica sets of a deployment for example. It is a combination of the [FNV-1a hash](https://pkg.go.dev/hash/fnv@go1.25.4#New32a) function and base 36 encoding. Not that SHA256 has anything wrong, but looks a bit wasteful to do that and then get only the first few bytes.

PR #215, opened 2025-12-01
Reject object creation for JWT users with empty groups claim

tzvatot:
## Summary
- Validates that objects are never created with empty tenants
- Treats empty tenant validation as a database integrity check, not a permissions check
- Validation is performed at the DAO layer to ensure integrity regardless of authentication method

## Architectural Rationale
Moving validation to the DAO layer ensures database integrity is maintained regardless of:
- Authentication method (JWT, API key, system, etc.)
- Future changes to tenancy logic
- Direct database access patterns

This approach treats empty tenants as a data integrity violation rather than a permission issue, which is more appropriate since the `tenants` column must never be empty for proper multi-tenancy support.

## Test Plan
- [x] Added test in `servers/tenancy_logic_test.go` to verify server-level rejection
- [x] Added test in `database/dao/tenancy_logic_test.go` to verify DAO-level rejection
- [x] Updated all server tests to use `SystemTenancyLogic` (assigns "shared" tenant)
- [x] All unit tests passing (683/683 specs across 13 suites)
- [x] All integration tests passing (31/31 specs)
- [x] Code formatted with `go fmt ./...`
- [x] Pre-commit hooks passed

Fixes https://github.com/innabox/fulfillment-service/issues/177

### DIFF
--- a/internal/database/dao/generic_dao_create.go
+++ b/internal/database/dao/generic_dao_create.go
@@ -82,6 +82,12 @@ func (r *CreateRequest[O]) do(ctx context.Context) (response *CreateResponse[O],
 		return
 	}
 
+	// Validate that tenants is not empty:
+	if len(tenants) == 0 {
+		err = errors.New("cannot create object with empty tenants")
+		return
+	}
+
 	// Save the object:
 	data, err := r.marshalData(r.object)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


tzvatot:
> Same for the "Update" operation: it should fail if it results in empty tenants.

@jhernand it seems like the UPDATE statement only modifies name, finalizers, and data - it never changes tenants.
So should the tenants considered for validation during update are the ones in the DB or the context?
If it's the DB tenants - then they are immutable. So validating on create is sufficient. If it's the context, and tenants are not immutable - then tenants should be updated as well?

jhernand (approved):

PR #229, opened 2025-12-11
Only update finalizers and status from reconcilers

jhernand:
Currently reconcilers send full updates of objects, which sometimes results in overwriting changes made by users to the spec of the object. To reduce the chances of that happening, this patch changes all reconcilers so that they will use an update mask to ensure that they only update the `metadata.finalizers` and the `status` of the objects.

This issue was discovered by Elad Tabak, and he is working on a more general solution in https://github.com/innabox/fulfillment-service/pull/214. Meanwhile, I am copying part of his work to make the integration tests less flaky.


jhernand:
FYI @tzvatot

PR #267, opened 2026-01-09
Add service account auth error and isolation tests

trewest:
## Summary

This PR adds comprehensive integration tests for multitenancy authentication error handling and basic tenant isolation. The tests verify that the fulfillment service correctly handles authentication errors and enforces tenant boundaries for service account-based authentication.

#### Authentication Error Handling Tests
- **Invalid token handling**: Verifies that requests with invalid JWT tokens return `401 Unauthorized` with a "permission denied" error message across fulfillment API endpoints (clusters, cluster_templates, host_pools, host_classes)
- **Unauthenticated requests**: Ensures that requests without authentication tokens are properly rejected with appropriate error responses

#### Tenant Isolation Tests
- **Cluster resource isolation**: 
  - Verifies that clusters are shared within the same tenant (users in the same tenant can see each other's clusters)
  - Verifies that clusters are isolated between different tenants (users cannot see clusters from other tenants)
- **Host pool resource isolation**:
  - Verifies that host pools are shared within the same tenant
  - Verifies that host pools are isolated between different tenants

### Infrastructure Updates: `it_tool.go`

- **Added `createUserServiceAccounts` function**: Automatically creates test service accounts and their associated namespaces during test setup
- **Updated `makeKubernetesTokenSource`**: Now accepts a `namespace` parameter to support multi-tenant token creation

### Test Variables: `testvars/testvars.go`

- **`ServiceAccountTenants`**: Defines test service accounts and their tenant assignments:
  - Tenant "a": alice, bob
  - Tenant "b": carol, dave
- **`MultiTenantUsers`**: Defines additional multi-tenant user configurations for future use

### DIFF
--- a/it/it_multitenancy_test.go
+++ b/it/it_multitenancy_test.go
@@ -1,0 +1,368 @@
+package it
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"net/http"
+	"time"
+
+	"github.com/innabox/fulfillment-common/testing"
+	ffv1 "github.com/innabox/fulfillment-service/internal/api/fulfillment/v1"
+	privatev1 "github.com/innabox/fulfillment-service/internal/api/private/v1"
+	"google.golang.org/grpc"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("Multitenancy authentication error handling", Label("multitenancy", "autherrors"), func() {
+	DescribeTable(
+		"returns error when authenticating with invalid token",
+		func(testUrl string) {
+			invalidToken := testing.MakeTokenString("test", 0)
+
+			req, err := http.NewRequest("GET", testUrl, nil)
+			Expect(err).ToNot(HaveOccurred())
+
+			// Set invalid token in request header
+			req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", invalidToken))
+
+			// Make request with invalid token
+			resp, err := http.Get(testUrl)
+			Expect(err).ToNot(HaveOccurred())
+
+			// Check that unauthorized status code is returned
+			Eventually(func(g Gomega) {
+				resp, err = http.Get(testUrl)
+				g.Expect(err).ToNot(HaveOccurred())
+				g.Expect(resp).ToNot(BeNil())
+				Expect(resp.StatusCode).To(BeNumerically(">=", http.StatusUnauthorized))
+				g.Expect(resp.StatusCode).To(Equal(http.StatusUnauthorized))
+			}).WithTimeout(100 * time.Millisecond).WithPolling(10 * time.Millisecond).Should(Succeed())
+
+			defer resp.Body.Close()
+
+			// Check that permission denied error is returned in response body
+			body, err := io.ReadAll(resp.Body)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(body).To(ContainSubstring("permission denied"))
+		},
+		Entry("clusters", fmt.Sprintf("https://%s/api/fulfillment/v1/clusters", serviceAddr)),
+		Entry("cluster templates", fmt.Sprintf("https://%s/api/fulfillment/v1/cluster_templates", serviceAddr)),
+		Entry("host pools", fmt.Sprintf("https://%s/api/fulfillment/v1/host_pools", serviceAddr)),
+		Entry("host classes", fmt.Sprintf("https://%s/api/fulfillment/v1/host_classes", serviceAddr)),
+	)
+
+	DescribeTable(
+		"returns error when user is not authenticated",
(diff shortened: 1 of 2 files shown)


it/it_multitenancy_test.go
```diff
@@ -0,0 +1,372 @@
...
+				"hY2UiOiJhIiwic2VydmljZWFjY291bnQiOnsibmFtZSI6ImFsaWNlIiwidWlkIjoiYWNmNTg3OGUtMjcwMi00NWRmLTgwYj" +
+				"MtOGU4MDVmNDU2YzdhIn19LCJuYmYiOjE3NjE2ODcxMjUsInN1YiI6InN5c3RlbTpzZXJ2aWNlYWNjb3VudDphOmFsaWNlI" +
+				"n0.Vee6HU5tV19XAkquoyJxcCoF3dg0yBMbfeOx2hagec0zZg4F4WS_zLc4BVU8TldQ_Fhauaq6rPCIOv8yWMfvrczIC-e3" +
+				"oMD8EaZ0MGzXIO4Lj3WWgjx0MccqGH8FZT4LjikPNXhHC-1QMjg74fcgsWXH5N4I2qVyCbbbOjNcFOkAraXD9CGji50hSw2" +
+				"6IyaiABjis8LeUv1QPURn5hr9n521uZ6a16O8fPodVEythRkwh7wjFPPakwUqOJe0kk0pfwo_W-Vndi669bm39KncUX-3ma" +
+				"EvH2wT1KXtYsCJPh03CGe3aZgFB7Csc6wpW3o4Qv6Y7fwHiqRBKoscAeWAAA"
+
+		for _, url := range testUrls {
+			req, err := http.NewRequest("GET", url, nil)
+			Expect(err).ToNot(HaveOccurred(), "error creating request to %s", url)
```
jhernand: It isn't usually necessary to add custom error messages. The default messages generated by the testing framework are usually good enough.

it/it_multitenancy_test.go
```diff
@@ -0,0 +1,372 @@
...
+			req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", invalidToken))
+
+			// Make request with invalid token
+			resp, err := http.Get(url)
+			Expect(err).ToNot(HaveOccurred(), "error making request to %s", url)
+
+			// Check that unauthorized status code is returned
+			Eventually(func() int {
+				resp, err = http.Get(url)
+				Expect(err).ToNot(HaveOccurred(), "error making request to %s", url)
```
jhernand: If you use `Expect` inside a `Eventually` function, and it fails, it will immediately fail the complete test, because it doesn't know that it is inside an `Eventually` function. The correct way to do this is to pass a `Gomega` object to the function, and then use the `Expect` method of that object:

```go
Eventually(func(g Gomega) int {
        resp, err = http.Get(url)
        g.Expect(err).ToNot(HaveOccurred())
})...
```

If you do that you can also use a function that returns no value, and check the condition inside the function:

```go
Eventually(func(g Gomega) {
        resp, err = http.Get(url)
        g.Expect(err).ToNot(HaveOccurred())
        g.Expect(resp.StatusCode).To(Equal(http.StatusUnauthorized))
})...
```

I also wonder why to use `Eventually` for this case; shouldn't this always return the unauthorized response code?

Finally, polling every two seconds and waiting ten seconds would make the test unnecessary long. If it is really needed consider making those settings shorter, in the order of milliseconds if possible.

it/testvars/testvars.go
```diff
@@ -0,0 +1,8 @@
+package testvars
+
+var ServiceAccountTenants = map[string]string{
```
jhernand: Can we have this variable in the `it_multitenancy_test.go` file directly? It is only used there, so there is no need for a new package, I think.

trewest: Moved it to `it_tool.go` let me know if you would like it somewhere else

it/it_multitenancy_test.go
```diff
@@ -0,0 +1,364 @@
...
+		resp, err := http.Get(testUrl)
+		Expect(err).ToNot(HaveOccurred())
+
+		// Check that unauthorized status code is returned
+		Eventually(func(g Gomega) {
+			resp, err = http.Get(testUrl)
+			g.Expect(err).ToNot(HaveOccurred())
+			Expect(resp.StatusCode).To(BeNumerically(">=", http.StatusUnauthorized))
+			g.Expect(resp.StatusCode).To(Equal(http.StatusUnauthorized))
+			g.Expect(resp).ToNot(BeNil())
```
jhernand: If you are going to check that `resp` is not nil it should be before using it, otherwise, if it was nil, then line 37 would fail with a nil pointer error.

jhernand (approved):
Looks good. Some minor comments that you may want to address before merging.

PR #373, opened 2026-04-01
MGMT-22988: Introduce IdP APIs and Keycloak Admin APIs

CrystalChun:
Identity Provider APIs allow an admin to CRUD organizations and users.

Currently, Keycloak is used in OSAC, but there are no hooks from the fulfillment service to actually manage users, organizations, etc. This adds APIs to allow a cluster or tenant admin to be able to manage users and organizations directly from the fulfillment service to create a seamless experience from OSAC.

This change is one part of the [enhancement for OSAC authentication](https://github.com/osac-project/enhancement-proposals/blob/main/enhancements/organizations/README.md).

Assisted-by: Claude

## Summary by CodeRabbit

* **New Features**
  * Identity provider public API: organization, user, credential and role models plus a Client interface for IdP operations
  * Keycloak backend integration and an organization manager with automated initial admin provisioning and role assignment
* **Tests**
  * HTTP integration tests for the Keycloak client and unit tests covering the organization manager and rollback behavior

### DIFF
--- a/internal/idp/client.go
+++ b/internal/idp/client.go
@@ -1,0 +1,65 @@
+/*
+Copyright (c) 2026 Red Hat Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with the
+License. You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific
+language governing permissions and limitations under the License.
+*/
+
+package idp
+
+import (
+	"context"
+)
+
+// Client is the generic interface for identity provider admin operations.
+// Different IdP providers (Keycloak, Auth0, Okta, etc.) implement this interface.
+type Client interface {
+	// Organization operations
+	CreateOrganization(ctx context.Context, org *Organization) (*Organization, error)
+	GetOrganization(ctx context.Context, name string) (*Organization, error)
+	DeleteOrganization(ctx context.Context, name string) error
+
+	// User operations
+	CreateUser(ctx context.Context, organizationName string, user *User) (*User, error)
+	GetUser(ctx context.Context, organizationName, userID string) (*User, error)
+	ListUsers(ctx context.Context, organizationName string) ([]*User, error)
+	DeleteUser(ctx context.Context, organizationName, userID string) error
+
+	// Role operations
+	// Roles can be at the organization level or client level
+	ListOrganizationRoles(ctx context.Context, organizationName string) ([]*Role, error)
+	ListClientRoles(ctx context.Context, organizationName, clientID string) ([]*Role, error)
+
+	// User role assignments
+	AssignOrganizationRolesToUser(ctx context.Context, organizationName, userID string, roles []*Role) error
+	AssignClientRolesToUser(ctx context.Context, organizationName, userID, clientID string, roles []*Role) error
+	RemoveOrganizationRolesFromUser(ctx context.Context, organizationName, userID string, roles []*Role) error
+	RemoveClientRolesFromUser(ctx context.Context, organizationName, userID, clientID string, roles []*Role) error
+	GetUserOrganizationRoles(ctx context.Context, organizationName, userID string) ([]*Role, error)
+	GetUserClientRoles(ctx context.Context, organizationName, userID, clientID string) ([]*Role, error)
+
+	// Admin permissions
+	// AssignOrganizationAdminPermissions grants full administrative access to an organization for the specified user.
+	// The implementation is provider-specific:
+	// - Keycloak: Assigns realm-management client roles (manage-realm, manage-users, etc.)
+	// - Auth0: Assigns organization Admin role
+	// - Okta: Assigns Organizational Administrator role
+	// - Azure AD: Assigns Global Administrator or Organizational Administrator role
+	AssignOrganizationAdminPermissions(ctx context.Context, organizationName, userID string) error
+
+	// AssignIdpManagerPermissions grants limited IdP management permissions to the specified user.
+	// This is used for the break-glass account which can manage users and identity providers
+	// but cannot modify critical organization settings.
+	// The implementation is provider-specific:
+	// - Keycloak: Assigns limited realm-management client roles (manage-users, view-users, etc.)
+	// - Auth0: Assigns organization Member Manager role
+	// - Okta: Assigns User Administrator role
+	// - Azure AD: Assigns User Administrator role
+	AssignIdpManagerPermissions(ctx context.Context, organizationName, userID string) error
+}
--- a/internal/idp/keycloak/client.go
+++ b/internal/idp/keycloak/client.go
@@ -1,0 +1,641 @@
+/*
+Copyright (c) 2026 Red Hat Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with the
+License. You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific
+language governing permissions and limitations under the License.
+*/
+
+package keycloak
+
+import (
+	"context"
+	"crypto/x509"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"log/slog"
+	"net/http"
+	"net/url"
+	"slices"
+	"strings"
+	"sync"
+
+	"github.com/google/uuid"
+
+	"github.com/osac-project/fulfillment-service/internal/apiclient"
+	"github.com/osac-project/fulfillment-service/internal/auth"
+	"github.com/osac-project/fulfillment-service/internal/idp"
+)
+
+// Client is a Keycloak-specific implementation of the idp.Client interface.
+type Client struct {
+	logger     *slog.Logger
+	httpClient *apiclient.Client
+
+	// realm-management client UUIDs by organization name
+	realmManagementUUIDs map[string]string
+	mu                   sync.RWMutex
+}
+
+// ClientBuilder builds a Keycloak client.
+type ClientBuilder struct {
+	logger      *slog.Logger
+	baseURL     string
+	tokenSource auth.TokenSource
+	caPool      *x509.CertPool
+	httpClient  *http.Client
+}
+
+// Ensure Client implements idp.Client at compile time
+var _ idp.Client = (*Client)(nil)
+
+// NewClient creates a builder for a Keycloak admin client.
+func NewClient() *ClientBuilder {
+	return &ClientBuilder{}
+}
+
+// SetLogger sets the logger.
+func (b *ClientBuilder) SetLogger(value *slog.Logger) *ClientBuilder {
+	b.logger = value
+	return b
+}
+
+// SetBaseURL sets the base URL of the Keycloak server.
+func (b *ClientBuilder) SetBaseURL(value string) *ClientBuilder {
+	b.baseURL = value
+	return b
+}
+
+// SetTokenSource sets the token source for authentication.
+func (b *ClientBuilder) SetTokenSource(value auth.TokenSource) *ClientBuilder {
+	b.tokenSource = value
+	return b
+}
+
+// SetCaPool sets the CA certificate pool.
+func (b *ClientBuilder) SetCaPool(value *x509.CertPool) *ClientBuilder {
+	b.caPool = value
+	return b
+}
+
+// SetHTTPClient sets a custom HTTP client.
+func (b *ClientBuilder) SetHTTPClient(value *http.Client) *ClientBuilder {
+	b.httpClient = value
+	return b
+}
+
+// Build creates the Keycloak client.
+func (b *ClientBuilder) Build() (result *Client, err error) {
+	if b.logger == nil {
+		err = errors.New("logger is mandatory")
+		return
+	}
+	if b.baseURL == "" {
+		err = errors.New("base URL is mandatory")
+		return
+	}
+	if b.tokenSource == nil {
+		err = errors.New("token source is mandatory")
+		return
+	}
+
+	// Build the underlying HTTP client
+	httpClientBuilder := apiclient.NewClient().
+		SetLogger(b.logger).
+		SetBaseURL(strings.TrimSuffix(b.baseURL, "/")).
+		SetTokenSource(b.tokenSource)
+
+	if b.caPool != nil {
+		httpClientBuilder = httpClientBuilder.SetCaPool(b.caPool)
+	}
+	if b.httpClient != nil {
+		httpClientBuilder = httpClientBuilder.SetHTTPClient(b.httpClient)
+	}
+
+	httpClient, err := httpClientBuilder.Build()
+	if err != nil {
+		return nil, fmt.Errorf("failed to build HTTP client: %w", err)
+	}
+
+	result = &Client{
+		logger:               b.logger,
+		httpClient:           httpClient,
+		realmManagementUUIDs: make(map[string]string),
+	}
+	return
+}
+
+// CreateOrganization creates a new organization (Keycloak realm).
+// Returns the created organization with server-assigned ID and any server defaults.
+func (c *Client) CreateOrganization(ctx context.Context, org *idp.Organization) (*idp.Organization, error) {
+	kcRealm := toKeycloakRealm(org)
+	response, err := c.httpClient.DoRequest(ctx, http.MethodPost, "/admin/realms", kcRealm)
+	if err != nil {
+		var apiErr *apiclient.APIError
+		if errors.As(err, &apiErr) && apiErr.StatusCode == http.StatusConflict {
+			return nil, fmt.Errorf("organization %q already exists: %w", org.Name, err)
+		}
+		return nil, fmt.Errorf("failed to create organization: %w", err)
+	}
+	response.Body.Close()
+
+	// Keycloak's POST /admin/realms returns 201 with no body, so we fetch the created realm
+	// to get the server-assigned ID and verify the realm was actually created
+	return c.GetOrganization(ctx, org.Name)
+}
+
+// GetOrganization retrieves an organization (Keycloak realm) by name.
+func (c *Client) GetOrganization(ctx context.Context, name string) (*idp.Organization, error) {
+	response, err := c.httpClient.DoRequest(ctx, http.MethodGet, fmt.Sprintf("/admin/realms/%s", url.PathEscape(name)), nil)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get organization: %w", err)
+	}
+	defer response.Body.Close()
+
+	var kcRealm keycloakRealm
+	if err = json.NewDecoder(response.Body).Decode(&kcRealm); err != nil {
+		return nil, fmt.Errorf("failed to decode organization response: %w", err)
+	}
+	return fromKeycloakRealm(&kcRealm), nil
+}
+
+// DeleteOrganization deletes an organization (Keycloak realm) by name.
+func (c *Client) DeleteOrganization(ctx context.Context, organizationName string) error {
+	response, err := c.httpClient.DoRequest(ctx, http.MethodDelete, fmt.Sprintf("/admin/realms/%s", url.PathEscape(organizationName)), nil)
+	if err != nil {
+		return fmt.Errorf("failed to delete organization: %w", err)
+	}
+	defer response.Body.Close()
+
+	// Clear cached realm-management UUID for this organization to prevent stale data
+	// if the organization is recreated later
+	c.mu.Lock()
+	delete(c.realmManagementUUIDs, organizationName)
+	c.mu.Unlock()
+
+	return nil
+}
+
+// CreateUser creates a new user in an organization.
+// Returns the created user with ID populated.
+func (c *Client) CreateUser(ctx context.Context, organizationName string, user *idp.User) (*idp.User, error) {
+	kcUser := toKeycloakUser(user)
+	response, err := c.httpClient.DoRequest(ctx, http.MethodPost, fmt.Sprintf("/admin/realms/%s/users", url.PathEscape(organizationName)), kcUser)
+	if err != nil {
+		var apiErr *apiclient.APIError
+		if errors.As(err, &apiErr) && apiErr.StatusCode == http.StatusConflict {
+			return nil, fmt.Errorf("user %q already exists: %w", user.Username, err)
+		}
+		return nil, fmt.Errorf("failed to create user: %w", err)
+	}
+	defer response.Body.Close()
+
+	// Extract user ID from Location header
+	// Location format: https://keycloak.example.com/admin/realms/{realm}/users/{user-id}
+	location := response.Header.Get("Location")
+	if location == "" {
+		return nil, fmt.Errorf("Location header not present in create user response")
+	}
+
+	// Extract the last segment of the URL path as the user ID
+	parts := strings.Split(strings.TrimSuffix(location, "/"), "/")
+	if len(parts) == 0 {
+		return nil, fmt.Errorf("failed to extract user ID from Location header: %s", location)
+	}
+
+	userID := parts[len(parts)-1]
+	if userID == "" {
+		return nil, fmt.Errorf("extracted user ID is empty from Location header: %s", location)
+	}
+
+	// Return a copy of the user with ID populated
+	createdUser := &idp.User{
+		ID:              userID,
+		Username:        user.Username,
+		Email:           user.Email,
+		EmailVerified:   user.EmailVerified,
+		Enabled:         user.Enabled,
+		FirstName:       user.FirstName,
+		LastName:        user.LastName,
+		Attributes:      user.Attributes,
+		Groups:          user.Groups,
+		Credentials:     user.Credentials,
+		RequiredActions: user.RequiredActions,
+	}
+
+	return createdUser, nil
+}
+
+// GetUser retrieves a user by ID.
+func (c *Client) GetUser(ctx context.Context, organizationName, userID string) (*idp.User, error) {
+	response, err := c.httpClient.DoRequest(ctx, http.MethodGet, fmt.Sprintf("/admin/realms/%s/users/%s", url.PathEscape(organizationName), url.PathEscape(userID)), nil)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get user: %w", err)
+	}
+	defer response.Body.Close()
+
+	var kcUser keycloakUser
+	if err = json.NewDecoder(response.Body).Decode(&kcUser); err != nil {
+		return nil, fmt.Errorf("failed to decode user response: %w", err)
+	}
+	return fromKeycloakUser(&kcUser), nil
+}
+
+// ListUsers lists all users in an organization.
+func (c *Client) ListUsers(ctx context.Context, organizationName string) ([]*idp.User, error) {
+	var allUsers []*idp.User
+	const maxPerPage = 100
+	first := 0
+
+	// Fetches all pages to ensure no users are missed due to Keycloak's pagination.
+	for {
+		// Check if context is cancelled before making the next API call
+		if err := ctx.Err(); err != nil {
+			return nil, err
+		}
+
+		// Fetch one page of users
+		path := fmt.Sprintf("/admin/realms/%s/users?first=%d&max=%d",
+			url.PathEscape(organizationName), first, maxPerPage)
+
+		response, err := c.httpClient.DoRequest(ctx, http.MethodGet, path, nil)
+		if err != nil {
+			return nil, fmt.Errorf("failed to list users: %w", err)
+		}
+
+		var kcUsers []keycloakUser
+		err = json.NewDecoder(response.Body).Decode(&kcUsers)
+		response.Body.Close()
+
+		if err != nil {
+			return nil, fmt.Errorf("failed to decode users response: %w", err)
+		}
+
+		// Convert and append this page
+		for _, kcUser := range kcUsers {
+			allUsers = append(allUsers, fromKeycloakUser(&kcUser))
+		}
+
+		// If we got fewer than max, we've reached the last page
+		if len(kcUsers) < maxPerPage {
+			break
+		}
+
+		// Move to next page
+		first += maxPerPage
+	}
+
+	return allUsers, nil
+}
+
+// DeleteUser deletes a user by ID.
+func (c *Client) DeleteUser(ctx context.Context, organizationName, userID string) error {
+	response, err := c.httpClient.DoRequest(ctx, http.MethodDelete, fmt.Sprintf("/admin/realms/%s/users/%s", url.PathEscape(organizationName), url.PathEscape(userID)), nil)
+	if err != nil {
+		var apiErr *apiclient.APIError
+		if errors.As(err, &apiErr) && apiErr.StatusCode == http.StatusNotFound {
+			return fmt.Errorf("user %q not found in organization %q: %w", userID, organizationName, err)
+		}
+		return fmt.Errorf("failed to delete user: %w", err)
+	}
+	defer response.Body.Close()
+	return nil
+}
+
+// ListOrganizationRoles lists all organization-level (realm) roles.
+func (c *Client) ListOrganizationRoles(ctx context.Context, organizationName string) ([]*idp.Role, error) {
+	response, err := c.httpClient.DoRequest(ctx, http.MethodGet, fmt.Sprintf("/admin/realms/%s/roles", url.PathEscape(organizationName)), nil)
+	if err != nil {
+		return nil, fmt.Errorf("failed to list realm roles: %w", err)
+	}
+	defer response.Body.Close()
+
+	var kcRoles []keycloakRole
+	if err = json.NewDecoder(response.Body).Decode(&kcRoles); err != nil {
+		return nil, fmt.Errorf("failed to decode realm roles response: %w", err)
+	}
+
+	roles := make([]*idp.Role, len(kcRoles))
+	for i, kcRole := range kcRoles {
+		roles[i] = fromKeycloakRole(&kcRole)
+	}
+	return roles, nil
+}
+
+// ListClientRoles lists all roles for a specific client.
+//
+// The clientID parameter accepts either format for convenience:
+//   - Human-readable clientId: "realm-management", "account", "my-app"
+//   - Internal UUID: "a1b2c3d4-e5f6-7890-abcd-ef1234567890"
+func (c *Client) ListClientRoles(ctx context.Context, organizationName, clientID string) ([]*idp.Role, error) {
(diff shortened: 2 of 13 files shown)


internal/idp/keycloak/client.go
```diff
@@ -0,0 +1,528 @@
...
+// isUUID checks if a string looks like a UUID.
+func isUUID(s string) bool {
+	// Simple check: UUIDs are 36 chars with hyphens in specific positions
+	if len(s) != 36 {
+		return false
+	}
+	return s[8] == '-' && s[13] == '-' && s[18] == '-' && s[23] == '-'
+}
+
+// AssignRealmRolesToUser assigns realm-level roles to a user.
```
mhrivnak: Does this overwrite existing roles? Or add to existing roles?

internal/idp/keycloak/client.go
```diff
@@ -0,0 +1,528 @@
...
+		err = fmt.Errorf("failed to send HTTP request: %w", err)
+		return
+	}
+
+	if response.StatusCode >= 400 {
+		defer response.Body.Close()
+		bodyBytes, _ := io.ReadAll(response.Body)
+		err = fmt.Errorf("Keycloak API error: status=%d body=%s", response.StatusCode, string(bodyBytes))
+		return
+	}
```
akshaynadkarni: When we return from this block, the `response` is still non-nil. In case a caller does something like: 
```
  if response != nil {
     data, err := io.ReadAll (response.Body)
      defer response.Body.Close()    // double-close! body was already closed
  }
```
we may get an error depending on the underlying implementation. 

For safety, it might be better to set the `response` to `nil` after explicitly closing the body: 

```suggestion
    if response.StatusCode >= 400 {
        bodyBytes, _ := io.ReadAll(response.Body)
        response.Body.Close()
        err = fmt.Errorf("Keycloak API error: status=%d body=%s", response.StatusCode, string(bodyBytes))
        response = nil
        return
    }
```

This will ensure callers can't access a closed response.

internal/idp/manager_test.go
```diff
@@ -0,0 +1,200 @@
...
+	return nil, nil
+}
+
+func (m *mockClient) ListClientRoles(ctx context.Context, organization, clientID string) ([]*Role, error) {
+	// Only return roles for realm-management client
+	if clientID == "realm-management" {
+		return []*Role{
+			{ID: "1", Name: "manage-realm", ApplicationRole: true},
+			{ID: "2", Name: "manage-users", ApplicationRole: true},
+			{ID: "3", Name: "manage-clients", ApplicationRole: true},
```
akshaynadkarni: In order to test that the filtering works, can we include roles that do not belong to the client? The code should only pass the supported roles.

internal/idp/manager_roles_test.go
akshaynadkarni: Please see comment here: https://github.com/osac-project/fulfillment-service/pull/373/changes#r3048304164

internal/idp/client.go
```diff
@@ -0,0 +1,93 @@
...
+}
+func (c *client) RemoveClientRolesFromUser(ctx context.Context, organization, userID, clientID string, roles []*Role) error {
+	return nil
+}
+func (c *client) GetUserRealmRoles(ctx context.Context, organization, userID string) ([]*Role, error) {
+	return nil, nil
+}
+func (c *client) GetUserClientRoles(ctx context.Context, organization, userID, clientID string) ([]*Role, error) {
+	return nil, nil
+}
```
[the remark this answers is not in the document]
CrystalChun: Very true, thank you Akshay! I've removed the dummy implementation.

internal/idp/keycloak/client.go
```diff
@@ -0,0 +1,528 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to delete organization: %w", err)
+	}
+	defer response.Body.Close()
+	return nil
+}
+
+// CreateUser creates a new user in an organization.
+// On success, populates user.ID with the ID of the created user.
+func (c *Client) CreateUser(ctx context.Context, organization string, user *idp.User) error {
```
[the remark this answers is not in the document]
CrystalChun: Great idea! Currently, Keycloak's APIs take the realm's name, but I wonder if we should make the realm name based off of the organization's ID. I'll update this with more documentation and better descriptive variable names. Thank you!

internal/idp/keycloak/client.go
```diff
@@ -0,0 +1,528 @@
...
+		if len(parts) > 0 {
+			user.ID = parts[len(parts)-1]
+		}
+	}
+
+	return nil
+}
+
+// GetUser retrieves a user by ID.
+func (c *Client) GetUser(ctx context.Context, organization, userID string) (*idp.User, error) {
```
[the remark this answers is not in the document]
CrystalChun: Agreed, thanks! Keycloak defines it as the realm name, so I've modified it to use OrganizationName. What do you think?

internal/idp/keycloak/client_test.go
```diff
@@ -0,0 +1,221 @@
...
+	"log/slog"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/osac-project/fulfillment-service/internal/auth"
+	"github.com/osac-project/fulfillment-service/internal/idp"
+)
+
+func TestKeycloakClient_CreateOrganization(t *testing.T) {
```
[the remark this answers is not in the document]
CrystalChun: I think so 😅 but do let me know if it's lacking!

internal/idp/keycloak/client.go
```diff
@@ -0,0 +1,642 @@
...
+	"github.com/osac-project/fulfillment-service/internal/idp"
+)
+
+// Client is a Keycloak-specific implementation of the idp.Client interface.
+type Client struct {
+	logger     *slog.Logger
+	httpClient *authclient.Client
+
+	// realm-management client UUIDs by organization name
+	realmManagementUUIDs map[string]string
```
akshaynadkarni, line 42: Just curious -- 
This is a specific map per org for `realm-management`. In the future, if there were other clients, would we update this into a more generic map for e.g. `map[org][client] = uuid`? OR would you create separate maps for each client? 

In Keycloak, does an org have a limited number of supported clients?

internal/idp/keycloak/client.go
```diff
@@ -0,0 +1,642 @@
...
+	return
+}
+
+// CreateOrganization creates a new organization (Keycloak realm).
+// Returns the created organization.
+func (c *Client) CreateOrganization(ctx context.Context, org *idp.Organization) (*idp.Organization, error) {
+	kcRealm := toKeycloakRealm(org)
+	response, err := c.httpClient.DoRequest(ctx, http.MethodPost, "/admin/realms", kcRealm)
+	if err != nil {
+		if strings.Contains(err.Error(), "already exists") {
```
akshaynadkarni: nit: Here we depend on the string content not changing to return the error. In case Keycloak changes the response string from `"already exists"` to `"exists already"`, this check would fail. 

One way to fix this (as per Claude): 

```
**Return a typed error with the status code** 

  type APIError struct {
      StatusCode int
      Body       string
  }

  func (e *APIError) Error() string {
      return fmt.Sprintf("API error: status=%d body=%s", e.StatusCode, e.Body)
  }

  Then the Keycloak client could do:
  response, err := c.httpClient.DoRequest(ctx, http.MethodPost, "/admin/realms", kcRealm)
  if err != nil {
      var apiErr *authclient.APIError
      if errors.As(err, &apiErr) && apiErr.StatusCode == http.StatusConflict {
          return nil, fmt.Errorf("organization %q already exists", org.Name)
      }
      return nil, fmt.Errorf("failed to create organization: %w", err)
  }
```

You don't have to address this as part of this PR. It's ok to tackle this in future PRs.

internal/idp/keycloak/client.go
```diff
@@ -0,0 +1,642 @@
...
+	}
+
+	return allUsers, nil
+}
+
+// DeleteUser deletes a user by ID.
+func (c *Client) DeleteUser(ctx context.Context, organizationName, userID string) error {
+	response, err := c.httpClient.DoRequest(ctx, http.MethodDelete, fmt.Sprintf("/admin/realms/%s/users/%s", url.PathEscape(organizationName), url.PathEscape(userID)), nil)
+	if err != nil {
+		if strings.Contains(err.Error(), "not found") {
```
akshaynadkarni: nit: Same comment as https://github.com/osac-project/fulfillment-service/pull/373/changes#r3065970368

internal/idp/manager.go
```diff
@@ -0,0 +1,290 @@
...
+	BreakGlassEmail string
+
+	// BreakGlassPassword is the temporary password for the break-glass account
+	// This is mandatory and must be changed on first login
+	BreakGlassPassword string
+}
+
+// CreateOrganization creates a complete IdP organization setup with a break-glass account.
+// Returns the break-glass account username, email, password, and error.
+func (m *OrganizationManager) CreateOrganization(ctx context.Context, config *OrganizationConfig) (username, email, password string, err error) {
```
akshaynadkarni: (can be addressed in a future PR)

- Why are we discarding the organizationID? 
- Would it be better to create a `User struct` or `Credentials struct`? 
  In this case it would be the breakglass User. This could include `username, email, password`. This would perhaps simplify the error cleanup and also enable extension of the struct to add any additional information vs. changing the function signature. 
  - Another option is to have a `CreateOrganizationResult` struct which includes organization and credential information.

internal/idp/keycloak/client.go
```diff
@@ -0,0 +1,642 @@
...
+}
+
+// ClientApp represents a Keycloak client application with both its identifiers.
+//
+// Keycloak clients have two types of identifiers:
+//   - ID: Internal UUID (e.g., "a1b2c3d4-e5f6-7890-...") used in API endpoints
+//   - ClientID: Human-readable identifier (e.g., "realm-management", "account") used in configuration
+//
+// Most Keycloak Admin REST API endpoints require the internal UUID, not the human-readable clientId.
+type ClientApp struct {
```
akshaynadkarni: Are there plans to use this struct in the future?

akshaynadkarni (approved):
Thanks @CrystalChun for addressing the comments. The code is so much better :) . 

I've left a few minor comments. Please take a look before merging, in case you want to handle these in the future. There's no need to handle them as part of this PR. 

Also, please use `/unhold` once you've read the comments to enable auto-merge.

internal/idp/keycloak/client.go
CrystalChun: > This is a specific map per org for realm-management. In the future, if there were other clients, would we update this into a more generic map for e.g. map[org][client] = uuid? OR would you create separate maps for each client?

Good question! For our current use-cases, I'm unsure if we'll use Keycloak's client construct for anything other than the admin permissions. 

If we were to use additional clients, I wonder how frequently they'd be called/used. If it's quite frequent, then a more generic map might be better. If it's infrequent, I wonder if we can get away with calling the API to resolve the UUID.

> In Keycloak, does an org have a limited number of supported clients?

From what I've read, it seems like there's no limit on clients per realm.

internal/idp/keycloak/client.go
CrystalChun: Great suggestion! Thank you! I've added this to the changes. How does it look?

internal/idp/keycloak/client.go
CrystalChun: Added! Thank you!

internal/idp/keycloak/client.go
CrystalChun: Whoops, good catch! This was leftover from a previous change. Removed it. Thank you!

internal/idp/manager.go
CrystalChun: > Why are we discarding the organizationID?

Good question. I pigeon-holed this implementation for keycloak, which doesn't track or care about the ID of a realm (organization), but it is a good idea to keep the organization ID in case other identity providers do use it.

I've updated the implementation to use the returned Organization object.

> Would it be better to create a User struct or Credentials struct?
In this case it would be the breakglass User. This could include username, email, password. This would perhaps simplify the error cleanup and also enable extension of the struct to add any additional information vs. changing the function signature.
> Another option is to have a CreateOrganizationResult struct which includes organization and credential information.

Yes, great suggestion! It would be unideal to change the function signature. I've updated this to use a `BreakGlassCredentials` struct to keep it minimal. What do you think?

internal/idp/keycloak/client_test.go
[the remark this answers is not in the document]
akshaynadkarni: I think you were in the midst of adding unit-tests when I commented :smile: . Coverage looks good :+1: .

internal/idp/keycloak/client.go
```diff
@@ -0,0 +1,669 @@
...
+//
+// This is needed because Keycloak's role-mapping API endpoints require the internal UUID,
+// but users typically work with human-readable clientIds.
+//
+// Example:
+//
+//	uuid, err := client.GetClientByClientID(ctx, "my-org", "realm-management")
+//	// Returns: "a1b2c3d4-e5f6-7890-..." (internal UUID)
+func (c *Client) GetClientByClientID(ctx context.Context, organizationName, clientID string) (string, error) {
+	// Try to look up as human-readable clientId first
```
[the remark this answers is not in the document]
akshaynadkarni: I think this works for now. Thanks for updating the code. In the future, if there are more clients, we can extend the cache: `org -> clientId -> uuid`

internal/idp/keycloak/client.go
akshaynadkarni: Looks good :+1:

internal/idp/manager.go
akshaynadkarni: Looks good :+1: . Thanks for making the change. 
In the future, if we identify we need generic credentials for other user types, we can generalize this into a `UserCredentials` struct.

akshaynadkarni (approved):
Nice changes. LGTM.

tzvatot (commented):
# Review: fulfillment-service#373

**MGMT-22988: Introduce IdP APIs and Keycloak Admin APIs** — +3692/-0 across 13 new files.

Well-architected PR introducing an IdP abstraction layer with a Keycloak implementation and organization lifecycle manager. Clean separation: generic `idp.Client` interface → Keycloak-specific implementation → `OrganizationManager` orchestration with rollback. Good test coverage with HTTP-level integration tests for Keycloak and unit tests with mock client for the manager.

| Category | Count |
|----------|-------|
| 🔴 Critical | 1 |
| 🟡 Important | 5 |
| 💡 Suggestion | 6 |

---

### 🔴 Critical: Break-glass password generation uses UUID (low entropy, hex-only)

`internal/idp/manager.go:224-225`
```go
password = strings.ReplaceAll(uuid.NewString(), "-", "")[:24]
```

UUID v4 provides ~122 bits of randomness, but after stripping hyphens and truncating to 24 chars, the password contains only hex characters (`[0-9a-f]`). That's 4 bits per character × 24 = 96 bits — but more importantly, the character diversity is terrible (16 possible chars vs 72+ for mixed-case alphanumeric + symbols). This makes it weaker against brute-force attacks that know the generation pattern.

**Recommendation:** Use `crypto/rand` with a full alphanumeric character set. Example:
```go
const charset = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789!@#$%"
b := make([]byte, 24)
for i := range b {
    n, _ := rand.Int(rand.Reader, big.NewInt(int64(len(charset))))
    b[i] = charset[n.Int64()]
}
password = string(b)
```

Also: the `BreakGlassCredentials` struct returns the plaintext password. Document that the caller is responsible for secure handling (encrypted storage, secure transmission).

---

### 🟡 Important: Rollback order is suboptimal — user deletion before org deletion

`internal/idp/manager.go:169-204`

The rollback deletes the user first, then the organization. In Keycloak, deleting a realm cascade-deletes all its users, clients, roles, etc. If the realm is in a partially broken state (which is why we're rolling back), the user deletion API call may fail because it targets a broken realm. Deleting the realm directly is simpler and more reliable.

**Recommendation:** Delete the organization first (cascade-deletes everything), skip explicit user deletion in rollback. The user deletion step is redundant.

---

### 🟡 Important: `DoRequest` returns nil response on error — unusual contract

`internal/authclient/client.go:240-249`

When status >= 400, the response body is read, the body is closed, `response` is set to nil, and an `APIError` is returned. This means callers get `(nil, error)` — which is fine if they check `err` first, but unusual for HTTP clients. A future caller who does `if resp.StatusCode == ...` before checking `err` will panic.

**Recommendation:** The godoc on `DoRequest` already says "response: HTTP response (caller must close response.Body)" but doesn't state that response is nil on error. Add: "Returns nil response when an error occurs (including HTTP status >= 400)."

---

### 🟡 Important: `realmManagementUUIDs` cache never invalidated

`internal/idp/keycloak/client.go` — The `realmManagementUUIDs` map caches the internal UUID of the `realm-management` client per organization, but:
1. It grows unbounded (one entry per organization, never evicted)
2. If an organization is deleted and recreated, the cached UUID becomes stale
3. `DeleteOrganization` doesn't clear the cache entry

**Recommendation:** Clear the cache entry in `DeleteOrganization`:
```go
func (c *Client) DeleteOrganization(ctx context.Context, organizationName string) error {
    // ... existing code ...
    c.mu.Lock()
    delete(c.realmManagementUUIDs, organizationName)
    c.mu.Unlock()
    return nil
}
```

---

### 🟡 Important: `keycloakClient` struct has many fields but only `ID` and `ClientID` are used

`internal/idp/keycloak/types.go:837-852`

The `keycloakClient` struct defines 12 fields (`Name`, `Description`, `Enabled`, `PublicClient`, `RedirectUris`, `WebOrigins`, `StandardFlowEnabled`, `ImplicitFlowEnabled`, `DirectAccessGrantsEnabled`, `ServiceAccountsEnabled`, `Attributes`, `Protocol`) but only `ID` and `ClientID` are ever read. This is speculative — adding unused fields increases maintenance burden and gives a false impression that client management is in scope.

**Recommendation:** Trim to the fields actually used:
```go
type keycloakClient struct {
    ID       string `json:"id,omitempty"`
    ClientID string `json:"clientId,omitempty"`
}
```
If client management is added later, the struct can be expanded then.

---

### 🟡 Important: Mock client in `manager_test.go` returns hardcoded 3 roles, not matching `keycloakIdpManagerRoles`

`internal/idp/manager_test.go:414-424`

The mock `ListClientRoles` returns `manage-realm`, `manage-users`, `manage-clients` — but the actual `keycloakIdpManagerRoles` list is `manage-users`, `view-users`, `manage-identity-providers`, `view-identity-providers`, `view-realm`. The mock doesn't include `view-users`, `manage-identity-providers`, `view-identity-providers`, or `view-realm`. The `AssignIdpManagerPermissions` mock also returns a different set than what the real implementation would filter.

This means the manager tests don't accurately verify the role filtering behavior — the mock bypasses the filtering logic entirely. The real filtering is only tested in the Keycloak `client_test.go`.

**Recommendation:** Either make the mock return the full set of roles (matching what a real Keycloak realm-management client returns) and let the manager's `AssignIdpManagerPermissions` do real filtering, or acknowledge that the manager tests are testing orchestration only, not role filtering.

---

### 💡 Suggestion: Package naming — `authclient` is generic, could conflict

`internal/authclient/` — The name `authclient` is very generic for what is essentially an authenticated HTTP client. It could be confused with authentication client (OAuth, OIDC). Consider `internal/httpclient/` or `internal/apiclient/` to better convey that this is a generic REST client with bearer token auth.

---

### 💡 Suggestion: `SetHttpClient` should be `SetHTTPClient` per Go conventions

`internal/authclient/client.go:146` and `internal/idp/keycloak/client.go:792`

Go naming convention uses `HTTP` (all caps) for acronyms in exported names. `SetHttpClient` should be `SetHTTPClient`. Same applies to `httpClient` field names in builder structs — though unexported names are less strict, consistency matters.

Reference: [Go Code Review Comments — Initialisms](https://github.com/golang/go/wiki/CodeReviewComments#initialisms)

---

### 💡 Suggestion: `ListUsers` pagination should check context cancellation

`internal/idp/keycloak/client.go:956-993`

The pagination loop fetches all pages without checking `ctx.Done()`. For organizations with thousands of users, this continues making API calls even after the context is cancelled (e.g., request timeout, client disconnect).

**Recommendation:** Add at the top of the loop:
```go
if err := ctx.Err(); err != nil {
    return nil, err
}
```

---

### 💡 Suggestion: `view-identity-providers` in `keycloakIdpManagerRoles` but not tested

`internal/idp/keycloak/const.go` — The `keycloakIdpManagerRoles` list includes `view-identity-providers` and `view-realm`, but the `AssignIdpManagerPermissions` test in `client_test.go:2326` only verifies `manage-users`, `view-users`, `manage-identity-providers`, and `view-realm`. It doesn't verify `view-identity-providers` is included. The test server's `allRoles` fixture includes roles matching all 5, but the assertion doesn't check for `view-identity-providers`.

**Recommendation:** Add `Expect(roleNames).To(ContainElement("view-identity-providers"))` to the test.

---

### 💡 Suggestion: `CreateOrganization` doesn't verify the realm was actually created

`internal/idp/keycloak/client.go:840-862`

After `POST /admin/realms`, the method returns a copy of the input `org` — it doesn't verify the realm was actually created (no GET after POST, no response body parsing). Keycloak's POST to `/admin/realms` returns 201 with no body, so there's no server-assigned ID to capture. The returned `Organization.ID` is whatever the caller provided, which may be empty.

**Recommendation:** This is fine for now since the Name is the primary identifier. But consider calling `GetOrganization` after creation if you need the server-assigned ID, or document that `ID` in the returned `Organization` is not server-generated.

---

### 💡 Suggestion: Test coverage gap — no test for `DeleteOrganizationRealm` failure path in manager

`internal/idp/manager_test.go:3695-3701`

The `DeleteOrganizationRealm` test only covers the happy path. There's no test for when `DeleteOrganization` fails (e.g., realm not found, server error).

**Recommendation:** Add a failure test:
```go
It("returns an error when deletion fails", func() {
    failingMock := &mockClient{...}
    // configure mock to fail on DeleteOrganization
    ...
})
```

---

Overall: Strong foundation with clean architecture. The `idp.Client` interface is well-designed for multi-provider support. Test coverage is thorough at the Keycloak HTTP level. The critical finding is the password generation entropy. The important findings are about rollback order, cache invalidation, and mock accuracy. The suggestions are mostly about Go conventions and minor test gaps.

internal/idp/manager.go
```diff
@@ -0,0 +1,247 @@
...
+	// Note: The client ID for realm-management varies by IdP provider
+	// For Keycloak, we need to get the client ID first
+	roles, err := m.client.ListClientRoles(ctx, organizationName, "realm-management")
+	if err != nil {
+		return fmt.Errorf("failed to list realm-management roles: %w", err)
+	}
+
+	// Filter for the key management roles
+	// Common roles: manage-realm, manage-users, manage-clients, view-realm, view-users, etc.
+	managementRoleNames := []string{
```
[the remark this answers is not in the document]
DakCrowder: Makes sense - didn't have anything else in mind ty!

CrystalChun:
@tzvatot Thank you for your review! I've addressed all of the remarks from your review comment, how does it look?

tzvatot (commented):
## Review: MGMT-22988: Introduce IdP APIs and Keycloak Admin APIs

This PR introduces a well-structured IdP abstraction layer with a clean `idp.Client` interface, a Keycloak implementation, and an `OrganizationManager` that orchestrates org creation with break-glass account provisioning and rollback on failure. The architecture is solid -- the separation between generic IdP types and Keycloak-specific types is well done, the builder pattern is consistent, and tests are thorough with good coverage of happy paths, error paths, pagination, caching, and rollback scenarios.

| Category | Count |
|----------|-------|
| 🔴 Critical | 0 |
| 🟡 Important | 2 |
| 💡 Suggestion | 1 |

---

### 🟡 Important: Rollback uses potentially cancelled context

**File:** `internal/idp/manager.go:243-249`

The `rollback` method reuses the same `ctx` that was passed to `CreateOrganization`. If the original operation failed due to context cancellation or deadline exceeded, the rollback's `DeleteOrganization` call will also fail immediately, leaving an orphaned realm with no break-glass account configured.

**Recommended fix:**

```go
func (m *OrganizationManager) rollback(ctx context.Context, organizationName string, deleteOrg bool) {
	if !deleteOrg {
		return
	}

	// Use a fresh context for cleanup so rollback succeeds even if
	// the original context was cancelled or timed out.
	cleanupCtx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
	defer cancel()

	m.logger.WarnContext(ctx, "Rolling back organization creation",
		slog.String("organization", organizationName),
	)

	if err := m.client.DeleteOrganization(cleanupCtx, organizationName); err != nil {
		// ...
```

---

### 🟡 Important: Unbounded read of error response body

**File:** `internal/apiclient/client.go:244`

`io.ReadAll(response.Body)` on HTTP error responses has no size limit. A misbehaving or compromised upstream could send an arbitrarily large response body, causing excessive memory allocation. Since this is a general-purpose HTTP client that could be reused for other integrations beyond Keycloak, it's worth bounding defensively.

**Recommended fix:**

```go
bodyBytes, _ := io.ReadAll(io.LimitReader(response.Body, 1<<20)) // 1 MB max
```

---

### 💡 Suggestion: Test setup duplication in `apiclient/client_test.go`

**File:** `internal/apiclient/client_test.go`

The keycloak tests have a `createTestClient` helper, but the `apiclient` tests repeat the `NewStaticTokenSource` + `NewClient` builder chain in every single test case (~10 lines each). A similar helper would reduce boilerplate significantly.

tzvatot:
> @tzvatot Thank you for your review! I've addressed all of the remarks from your review comment, how does it look?

Thanks for addressing those. I added another smaller review this time the scope is minimal. Thanks again for looking into it.

CrystalChun:
> Thanks for addressing those. I added another smaller review this time the scope is minimal. Thanks again for looking into it.

Thank you for reviewing again! Addressed the most recent review, how does it look?

CrystalChun:
Sorry branch was out of date

akshaynadkarni (approved):
LGTM 👍

eranco74:
## Code Review: IdP APIs and Keycloak Admin Integration

### ✅ Overall Assessment

Excellent foundation for identity provider integration with clean abstraction layers and comprehensive error handling. The code demonstrates strong software engineering practices with proper rollback semantics, extensive test coverage, and provider-agnostic design.

**Strengths:**
- **Clean architecture:** Provider-agnostic `idp.Client` interface allows swapping Keycloak for Auth0/Okta/Azure AD
- **Comprehensive test coverage:** 1,433 lines of Keycloak client tests + 465 lines of manager tests
- **Robust error handling:** Rollback semantics with fresh context on cleanup (handles cancelled contexts)
- **Security-conscious:** Uses `crypto/rand` for password generation, bearer token auth, TLS support
- **All CI checks passing:** Unit tests, integration tests, format checks all green
- **Already in merge pool** (tide status: pass)

---

### 🎯 Architecture Review

#### 1. Three-Layer Design ✅

```
OrganizationManager (orchestration + rollback)
    ↓
idp.Client interface (provider-agnostic contract)
    ↓
keycloak.Client (Keycloak-specific implementation)
```

**Benefits:**
- Future IdP providers (Auth0, Okta, Azure AD) can be added without changing manager logic
- Easy to test with mock clients
- Clean separation: manager = workflow, client = HTTP calls

---

### 📋 Key Components Review

#### 1. HTTP Client (`internal/apiclient/client.go`) ✅

**Good practices:**
```go
// Line 245: Limits error response body to 1MB
bodyBytes, _ := io.ReadAll(io.LimitReader(response.Body, 1<<20))
```
✅ **Excellent:** Prevents memory exhaustion from malicious/broken servers

```go
// Line 254-255: Sets response to nil after error
response.Body.Close()
response = nil
```
✅ **Good practice:** Forces callers to check `err` before using `response` (prevents nil pointer dereference)

**Custom TLS CA pool support:**
```go
// Line 169-172
if b.caPool != nil {
    transport.TLSClientConfig = &tls.Config{
        RootCAs: b.caPool,
    }
}
```
✅ **Important:** Allows OSAC to work with self-signed Keycloak certificates in on-prem deployments

---

#### 2. Keycloak Client (`internal/idp/keycloak/client.go`) ✅

**Realm-Management UUID Caching:**
```go
// Line 693-695
realmManagementUUIDs map[string]string
mu                   sync.RWMutex
```
✅ **Smart optimization:** Keycloak requires client UUID for role operations. Caching avoids repeated lookups.

**Thread-safety:** Uses `sync.RWMutex` correctly for concurrent access to the cache.

**Error handling examples:**
```go
// Line 792-795: Detects conflict on duplicate org creation
if errors.As(err, &apiErr) && apiErr.StatusCode == http.StatusConflict {
    return nil, fmt.Errorf("organization %q already exists: %w", org.Name, err)
}
```
✅ **Good:** Translates HTTP 409 Conflict into domain-specific error

---

#### 3. Organization Manager (`internal/idp/manager.go`) ✅

**Transactional Semantics with Rollback:**
```go
// Line 113-130
defer func() {
    if err != nil {
        m.rollback(ctx, config.Name, organizationCreated)
    }
}()

// Step 1: Create organization
createdOrg, err := m.client.CreateOrganization(ctx, org)
if err != nil {
    return nil, fmt.Errorf("failed to create organization: %w", err)
}
organizationCreated = true

// Step 2: Create break-glass account
credentials, err = m.createBreakGlassAccount(ctx, config)
if err != nil {
    return nil, fmt.Errorf("failed to create break-glass account: %w", err)
}
```

✅ **Excellent pattern:** 
- Tracks state with `organizationCreated` flag
- Deferred rollback cleanup
- Each step can fail without leaving partial state

**Rollback with Fresh Context:**
```go
// Line 177-179
cleanupCtx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
defer cancel()
```
✅ **Critical detail:** Rollback uses fresh context so cleanup succeeds even if original context was cancelled/timeout. Test case explicitly verifies this (lines 1376-1402 in manager_test.go).

**Break-Glass Account Creation:**
```go
// Line 203-214: Smart defaults
username := config.BreakGlassUsername
if username == "" {
    username = fmt.Sprintf("%s-osac-break-glass", config.Name)
}

email := config.BreakGlassEmail
if email == "" {
    email = fmt.Sprintf("break-glass@%s.osac.local", config.Name)
}
```
✅ **Good UX:** Generates sensible defaults if not provided

**Secure Password Generation:**
```go
// Uses crypto/rand (not math/rand)
const charset = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789!@#$%"
const passwordLength = 24
b := make([]byte, passwordLength)
for i := range b {
    n, err := rand.Int(rand.Reader, big.NewInt(int64(len(charset))))
    if err != nil {
        return nil, fmt.Errorf("failed to generate random password: %w", err)
    }
    b[i] = charset[n.Int64()]
}
```
✅ **Cryptographically secure:** Uses `crypto/rand`, not `math/rand`
✅ **Sufficient entropy:** 24 characters from 67-char charset ≈ 143 bits of entropy

---

### 🔍 Potential Improvements (Non-Blocking)

#### 1. Password Expiration for Break-Glass Account

**Current behavior:** Break-glass account is created with a password that requires change on first login (`Temporary: true`).

**Suggestion:** Consider adding an expiration policy:
```go
RequiredActions: []string{"UPDATE_PASSWORD"},
// Add: expires after 24 hours if not changed
```

**Rationale:** Defense-in-depth if break-glass credentials are leaked before first use.

---

#### 2. Audit Logging for Admin Operations

**Current state:** Debug logs exist for operations:
```go
m.logger.InfoContext(ctx, "Creating IdP organization", 
    slog.String("organization", config.Name))
```

**Suggestion:** Consider structured audit events:
```go
m.logger.InfoContext(ctx, "IdP admin operation",
    slog.String("operation", "create_organization"),
    slog.String("organization", config.Name),
    slog.String("actor", actorFromContext(ctx)),  // Who triggered this
    slog.String("audit", "true"),  // Marks as audit event
)
```

**Rationale:** Identity operations (create org, assign admin) are security-sensitive and should be auditable.

---

#### 3. Rate Limiting for IdP Operations

**Observation:** No rate limiting on `CreateOrganization`, `CreateUser`, etc.

**Risk:** Abusive/compromised admin could spam org/user creation.

**Suggestion:** Add rate limiting at the manager or service layer:
```go
// Example with golang.org/x/time/rate
if !m.rateLimiter.Allow() {
    return nil, errors.New("rate limit exceeded for IdP operations")
}
```

**Alternative:** Rely on Keycloak's built-in rate limiting (if configured).

---

#### 4. Idempotency for CreateOrganization

**Current behavior:** Returns error if organization already exists (409 Conflict).

**Suggestion:** Make it idempotent:
```go
createdOrg, err := m.client.CreateOrganization(ctx, org)
if err != nil {
    var apiErr *apiclient.APIError
    if errors.As(err, &apiErr) && apiErr.StatusCode == http.StatusConflict {
        // Org already exists — fetch and return it instead of failing
        return m.client.GetOrganization(ctx, org.Name)
    }
    return nil, err
}
```

**Benefit:** Retry-safe, works better with declarative APIs.

**Trade-off:** Hides race conditions where two clients create the same org simultaneously.

---

### 🧪 Test Coverage Assessment

**Unit Tests:** ✅ Excellent
- `apiclient`: 259 lines (bearer auth, error handling, TLS)
- `keycloak`: 1,433 lines (HTTP integration tests with httptest server)
- `manager`: 465 lines (rollback scenarios, context cancellation)

**Test Quality Highlights:**
1. **Rollback tests verify cleanup happens even on cancelled context** (line 1376)
2. **Password generation test validates entropy** (checks 24-char length + charset regex)
3. **Mock client pattern** enables testing manager without real Keycloak
4. **HTTP integration tests** use `httptest.Server` (no external dependencies)

**Missing (non-blocking):**
- Integration test against real Keycloak instance (could be added in `it/` directory)
- Concurrent access test for realm-management UUID cache (RWMutex stress test)

---

### 🔒 Security Review

| Aspect | Assessment | Notes |
|--------|------------|-------|
| **Password Generation** | ✅ Secure | Uses `crypto/rand`, 24 chars, good charset |
| **TLS Support** | ✅ Good | Custom CA pool for self-signed certs |
| **Authentication** | ✅ Proper | Bearer tokens via `TokenSource` |
| **Error Response Limits** | ✅ Good | 1MB limit prevents DoS |
| **Context Handling** | ✅ Excellent | Fresh context for rollback |
| **Credential Storage** | ⚠️ Unspecified | Where are break-glass creds stored after creation? |

**Question on Credential Storage:**
The `BreakGlassCredentials` struct returns username/password after org creation. Where does the caller store these? 

**Recommendation:** Document whether these should be:
- Stored in a Secret Manager (Vault, AWS Secrets Manager)
- Returned to user for manual storage (break-glass SOP)
- Stored in Kubernetes Secret (if running in K8s)

---

### 📋 Summary

| Aspect | Rating | Notes |
|--------|--------|-------|
| **Architecture** | ✅ Excellent | Clean abstraction, provider-agnostic |
| **Error Handling** | ✅ Excellent | Rollback with fresh context |
| **Security** | ✅ Good | Secure password gen, TLS, bearer auth |
| **Test Coverage** | ✅ Excellent | 2,157 lines of tests, rollback scenarios |
| **Code Quality** | ✅ High | Builder patterns, proper locking, DRY |
| **CI Status** | ✅ All passing | Already in merge pool |

---

### ✅ Recommendation

**Approve and merge.** The suggested improvements are enhancements for future iterations, not blockers. The core implementation is solid and production-ready.

**Post-merge follow-ups:**
1. Document break-glass credential storage strategy
2. Consider audit logging for security-sensitive operations
3. Add rate limiting if not relying on Keycloak's built-in limits
4. Integration test against real Keycloak (in `it/` directory)

Excellent work on this IdP integration foundation! 🎯

PR #384, opened 2026-04-09
MGMT-23808: Integrate leader election into the controller

jhernand:
## Summary

- Wrap the controller's reconciler startup in a leader election loop using the `LeaderElector`
  from `internal/coordination`. Only the instance that holds the `fulfillment-controller` lease
  creates and starts reconcilers.
- Add a `--leader-identity` flag (defaults to hostname) for the holder identity used in
  leader election.
- Extract reconciler creation into a `startReconcilers` method called from the leader election
  callback, keeping the `run` method focused on infrastructure setup.
- Make the controller replica count configurable via `.Values.controller.replicas` in the Helm
  chart (defaults to 1).

## Test plan

- [x] `ginkgo run -r internal` passes (all 34 suites).
- [x] `go build ./...` succeeds.
- [x] Deployed with 2 replicas: only one instance actively reconciles while the other waits for
  the lease to expire.

## Summary by CodeRabbit

* **New Features**
  * Controller can run multiple replicas with leader election; only the leader actively performs reconciliation
  * Replica count configurable via Helm values (default: 1)
  * CLI flag added to set leader identity

* **Behavioral Changes**
  * Service reports SERVING at startup when no components are registered (health aggregation updated)

* **Tests**
  * Health initialization tests updated to expect SERVING at startup

### DIFF
--- a/charts/service/templates/controller/deployment.yaml
+++ b/charts/service/templates/controller/deployment.yaml
@@ -22,7 +22,7 @@ spec:
   selector:
     matchLabels:
       app: fulfillment-controller
-  replicas: 1
+  replicas: {{ .Values.controller.replicas | default 1 }}
   strategy:
     type: {{ if .Values.debug }}Recreate{{ else }}RollingUpdate{{ end }}
   template:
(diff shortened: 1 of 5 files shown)


jhernand:
This doesn't work well. Will resume work in another pull request later.

PR #395, opened 2026-04-14
MGMT-23659: NetworkClass ID should use UUID instead of FQN

tzvatot:
## Summary
- Add `fqn` field to NetworkClass proto (private and public) to hold the fully qualified template name
- Server always generates UUID for `id`, ignoring caller-provided values
- `fqn` is required on create and immutable on update
- Updated all tests to include `fqn` field

## Jira
[MGMT-23659](https://redhat.atlassian.net/browse/MGMT-23659)

## Related
- osac-aap counterpart PR to follow (updates template publishing to use `fqn`)

## Test plan
- [x] All 424 server unit tests pass
- [ ] Integration tests (requires kind cluster)

## Summary by CodeRabbit

* **New Features**
  * Added fully qualified name (`fqn`) field to NetworkClass for identifying source templates
  * Added customizable `VERSION` build argument to containerized builds with automatic Git-based version detection

* **Tests**
  * Updated test fixtures across multiple test suites to include `Fqn` field values
  * Added validation tests for `fqn` immutability and required field enforcement on creation


eranco74:
Why do we need the fqn? isn't it the same as implementation strategy?

tzvatot:
> Why do we need the fqn? isn't it the same as implementation strategy?

They serve different purposes:                                                                                                                                                                                                                                     
- implementation_strategy is the network backend type (e.g. "cudn", "physnet") — multiple NetworkClasses could share the same strategy with different configurations.
- fqn is the fully qualified template role name (e.g. "osac.templates.cudn_net") — it's the stable external identifier that the publish_templates playbook uses to match existing NetworkClasses for upsert (update-or-create).                                    

Before this change, the FQN was used directly as the id, which is what the ticket asks us to fix. With UUID-based IDs, we need fqn as the upsert key so the Ansible side knows which existing NetworkClass to update vs create a new one.                          
                                                                                                                                                                                                                                                                   
That said, if you think implementation_strategy could serve as the upsert key instead (assuming a 1:1 mapping), we could drop fqn and use that. What do you think?

tzvatot:
Closing in favor of #396 — dropped the `fqn` field per review feedback from @eranco74. The `implementation_strategy` field already serves as a sufficient upsert key for template publishing, making `fqn` redundant.

PR #400, opened 2026-04-14
NO-ISSUE: Move tenant computation from the service to Authorino

jhernand:
## Summary

- Moves tenant and username computation from Go code into the Authorino
  AuthConfig CEL expressions, so the `x-subject` response header already
  contains resolved values (namespace as tenant, short name as user) for
  service accounts. JWT users pass through groups and username as before.
- Merges the now-identical `JwtTenancyLogic` and
  `ServiceAccountTenancyLogic` into a single `DefaultTenancyLogic` that
  reads `subject.Tenants` directly.
- Removes the `Source` field and `SubjectSource` type from `Subject`
  since no code branches on them anymore.

## Test plan

- [x] All unit tests pass (`ginkgo run -r internal` — 35 suites)
- [x] Integration tests pass (`ginkgo run it`)

## Summary by CodeRabbit

* **Refactor**
  * Updated authentication subject model structure
  * Unified tenancy logic across authentication methods
  * Simplified and consolidated authorization behavior


internal/cmd/service/start/grpcserver/start_grpc_server_cmd.go
```diff
@@ -374,7 +367,7 @@ func (c *runnerContext) run(cmd *cobra.Command, argv []string) error {
 		}
 	default:
 		return fmt.Errorf(
-			"unknown tenancy logic '%s', valid values are 'default', 'serviceaccount', and 'guest'",
+			"unknown tenancy logic '%s', valid values are 'default' and 'guest'",
 			c.args.tenancyLogic,
```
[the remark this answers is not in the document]
jhernand: Done.

jhernand:
@CrystalChun this may be relevant for your work to integrate with identity providers.

akshaynadkarni:
@jhernand I used AI to review this PR. Here is a review summary: 

## What the PR Does Correctly

- Clean removal of the `Source` field and `SubjectSource` type with no remaining references in Go code.
- Correct renaming of `Groups` to `Tenants` throughout, including JSON tags, test assertions, and the AuthConfig response header.
- The CEL expressions correctly extract namespace (`split(":")[2]`) and SA name (`split(":")[3]`) from the `system:serviceaccount:<ns>:<name>` format.
- Using `[value]` to wrap the single tenant into a list is correct CEL syntax.
- The ternary direction flip (checking `== "jwt"` instead of `== "serviceaccount"`) is semantically equivalent given only two authnMethods are configured.
- Both the Helm template and kustomize base AuthConfig were updated consistently.
- Test restructuring is thorough: tests are now organized by behavior (assignable, default, visible) rather than by auth source, which better reflects the unified logic.
- Integration test code (`it/it_tool.go`) references to `"groups"` are correctly left unchanged since they refer to Keycloak realm configuration, not the `Subject.Groups` field.
- The `DetermineVisibleTenants` method correctly allows users with no tenants to still see shared resources (returns `SharedTenants` without error), while `DetermineAssignableTenants` properly errors, preventing object creation without tenant membership.
- Deployment Helm chart uses `--tenancy-logic=default`, which is still valid.

---

## Summary Table

| # | Type | Severity | File | Description | Link |
|---|------|----------|------|-------------|------|
| 1 | Issue | Medium | `.vscode/launch.json` | Stale `--tenancy-logic=serviceaccount` launch config will fail at startup | [L84](https://github.com/osac-project/fulfillment-service/blob/1502b0ca348f4c78c8c0738afc0f3d7343e3af09/.vscode/launch.json#L84) |
| 2 | Issue | Medium | `docs/AUTH.md` | Still documents removed `serviceaccount` tenancy logic option | [L442-L458](https://github.com/osac-project/fulfillment-service/blob/1502b0ca348f4c78c8c0738afc0f3d7343e3af09/docs/AUTH.md#L442-L458) |
| 3 | Question | Low | `authconfig.yaml` | No bounds check on `split` for malformed SA usernames (intentional trade-off?) | [L170](https://github.com/osac-project/fulfillment-service/blob/1502b0ca348f4c78c8c0738afc0f3d7343e3af09/manifests/base/grpc-server/authconfig.yaml#L170) |
| 4 | Suggestion | Low | `default_tenancy_logic.go` | Empty set allocated before error check on empty tenants | [L64](https://github.com/osac-project/fulfillment-service/blob/1502b0ca348f4c78c8c0738afc0f3d7343e3af09/internal/auth/default_tenancy_logic.go#L64) |

akshaynadkarni:
I think the first 2 are changes you can include in this PR, even though the files are currently not part of the PR. 

Comment #1 
Related to: https://github.com/osac-project/fulfillment-service/blob/10ccec05a8d0e7e393cd7280ab16d7a70190d3ec/.vscode/launch.json#L84

Since the `--tenancy-logic=serviceaccount` value was removed in this PR, so this launch configuration will fail at startup with `unknown tenancy logic 'serviceaccount', valid values are 'default' and 'guest'`.

---

Comment #2
Related to: https://github.com/osac-project/fulfillment-service/blob/10ccec05a8d0e7e393cd7280ab16d7a70190d3ec/docs/AUTH.md?plain=1#L442-L458

Suggested changes:
1. Remove the "Service Account" subsection (lines 442-458) entirely.
2. Update the "Default" heading from "Default (JWT-Authenticated Users)" to just "Default" or "Default (All Authenticated Users)".
3. Update the description under "Default" to explain that both JWT groups and service account namespaces are resolved by Authorino into a uniform `tenants` field before reaching the service.

---

I will include the remaining comments directly in the file locations.

charts/service/templates/grpc-server/authconfig.yaml
```diff
@@ -164,12 +164,13 @@ spec:
...
-              groups:
+                  auth.identity.authnMethod == "jwt"
+                    ? auth.identity.username
+                    : auth.identity.user.username.split(":")[3]
+              tenants:
                 expression: |
-                  auth.identity.authnMethod == "serviceaccount"? auth.identity.user.groups: auth.identity.groups
+                  auth.identity.authnMethod == "jwt"
+                    ? auth.identity.groups
+                    : [auth.identity.user.username.split(":")[2]]
```
akshaynadkarni, line 176: The old `ServiceAccountTenancyLogic` validated the username format with a regex (`^system:serviceaccount:([^:]*):[^:]*$`) before extracting the namespace. The new CEL expressions use `split(":")[2]` and `split(":")[3]` without format validation.

If the username from `kubernetesTokenReview` somehow doesn't contain 4 colon-separated parts, the split+index would cause a CEL runtime error. In practice, Kubernetes token review should always produce the `system:serviceaccount:<ns>:<name>` format for service accounts, so this is likely safe. But is this an intentional trade-off (rely on Kubernetes guarantees) vs. the previous explicit validation?

internal/auth/default_tenancy_logic.go
```diff
@@ -49,77 +46,47 @@ func (b *DefaultTenancyLogicBuilder) SetLogger(value *slog.Logger) *DefaultTenan
...
+		p.logger.ErrorContext(
+			ctx,
+			"Subject has no tenants",
+			slog.String("user", subject.User),
+		)
+		err = fmt.Errorf("subject must belong to at least one tenant to create objects")
 		return
 	}
-	return delegate.DetermineAssignableTenants(ctx)
+	return
```
akshaynadkarni, line 74: Minor: the method creates the result set (`collections.NewSet(subject.Tenants...)`) before checking if tenants is empty. On the error path, it returns both an empty set and a non-nil error. Callers should check `err` first and ignore `result` on error, so this is not a bug, but it does allocate an empty set unnecessarily. You could swap the order: check `len(subject.Tenants) == 0` first, then build the set. Something like: 

```suggestion
        subject := SubjectFromContext(ctx)
        if len(subject.Tenants) == 0 {
                p.logger.ErrorContext(
                        ctx,
                        "Subject has no tenants",
                        slog.String("user", subject.User),
                )
                err = fmt.Errorf("subject must belong to at least one tenant to create objects")
                return
        }
        result = collections.NewSet(subject.Tenants...)
        return
```

I don't have a strong opinion here. Just didn't know if you were intentionally initializing the empty set before the check.

charts/service/templates/grpc-server/authconfig.yaml
jhernand: This relies on the claims of the tokens issued by Kubernetes for service accounts, which will always contain the four components. The previous approach wasn't intended to validate that, only to extract the components of the claim using the regular expression. I think this doesn't deserve any more complication.

eranco74:
## Code Review: Move Tenant Computation to Authorino

### ✅ Overall Assessment

Excellent refactoring that simplifies the authentication architecture by pushing tenant/user extraction logic to Authorino's CEL expressions. This is a great example of "push complexity to the edges" — moving string parsing out of application code and into declarative configuration.

**Strengths:**
- **Massive code reduction:** Removes 552 lines (2 tenancy logic implementations + tests)
- **Simplified model:** `Subject` no longer needs `Source` field or branching logic
- **Better separation of concerns:** Authorino handles auth token parsing, service gets clean tenants list
- **All tests passing:** 35 unit test suites + integration tests (Helm + Kustomize)
- **Unified logic:** `DefaultTenancyLogic` now works for both JWT and service accounts

---

### 🎯 Architecture Improvement

**Before:**
```
Authorino → raw JWT/SA data → Go code parses:
  - JWT: groups → tenants
  - SA: system:serviceaccount:NS:NAME → extract NS as tenant
```

**After:**
```
Authorino → CEL expressions resolve tenants → Go code reads subject.Tenants directly
```

This is the right direction — declarative auth policy belongs in the auth layer, not scattered through application code.

---

### 📋 Key Changes Review

#### 1. AuthConfig CEL Expressions ✅

**File:** `charts/service/templates/grpc-server/authconfig.yaml` (lines 11-23)

The CEL expressions correctly handle both auth methods:

**JWT users:**
```yaml
user: auth.identity.username
tenants: auth.identity.groups
```

**Service accounts:**
```yaml
user: auth.identity.user.username.split(":")[3]      # Extract SA name from system:serviceaccount:ns:sa
tenants: [auth.identity.user.username.split(":")[2]]  # Extract namespace as single-tenant list
```

✅ **Correct:** Service account username format is `system:serviceaccount:<namespace>:<sa-name>`, so:
- Index [2] = namespace → becomes tenant
- Index [3] = service account name → becomes user

**Note:** This assumes service accounts always have this exact format. Kubernetes guarantees this for SA tokens, so it's safe.

---

#### 2. Simplified Subject Model ✅

**File:** `internal/auth/auth_subject.go`

Removed fields:
- ❌ `Source SubjectSource` — no longer needed (auth type determined upstream)
- ❌ `Groups []string` — renamed to `Tenants` (clearer semantics)

New structure:
```go
type Subject struct {
    User    string   `json:"user"`
    Tenants []string `json:"tenants"`
}
```

✅ **Good:** The model now represents the *resolved* identity, not the *source* of that identity. This is cleaner separation of concerns.

---

#### 3. Unified Tenancy Logic ✅

**File:** `internal/auth/default_tenancy_logic.go`

**Before:** 
- `DefaultTenancyLogic` delegated to `JwtTenancyLogic` or `ServiceAccountTenancyLogic` based on `Subject.Source`
- Each delegate had custom parsing logic

**After:**
```go
func (p *DefaultTenancyLogic) DetermineAssignableTenants(ctx context.Context) (collections.Set[string], error) {
    subject := SubjectFromContext(ctx)
    result := collections.NewSet(subject.Tenants...)
    if len(subject.Tenants) == 0 {
        return nil, fmt.Errorf("subject must belong to at least one tenant to create objects")
    }
    return result, nil
}
```

✅ **Excellent:** Now just reads `subject.Tenants` directly. The complexity moved to Authorino's CEL layer.

**Validation:** Empty tenants check prevents objects from being created without tenant isolation — critical for multi-tenancy security.

---

#### 4. Test Coverage ✅

**File:** `internal/auth/default_tenancy_logic_test.go`

The test suite was refactored to match the simplified logic:

**Removed tests:**
- ❌ Service account namespace extraction tests (now handled by Authorino)
- ❌ JWT group mapping tests (now handled by Authorino)
- ❌ Invalid source type tests (Source field removed)

**Kept/Updated tests:**
- ✅ Returns tenants from subject
- ✅ Fails if subject has no tenants (security validation)
- ✅ Visible tenants include shared tenant
- ✅ Multiple tenants handled correctly

**Coverage assessment:** The new tests cover the contract (`subject.Tenants` must be populated), but don't test the CEL expressions themselves. This is OK since:
1. Authorino's CEL engine is battle-tested
2. Integration tests verify the full auth flow
3. Manual testing on edge22 would catch CEL parsing errors

---

### 💡 Minor Issue: Outdated Help Text

**File:** `internal/cmd/service/start/grpcserver/start_grpc_server_cmd.go`

CodeRabbit flagged this (lines 369-371): The `--tenancy-logic` flag help text may still reference removed options like `serviceaccount`.

**Recommended:** Update the flag description to reflect only `default` and `guest` options (since `serviceaccount` and `jwt` logic no longer exist as separate implementations).

---

### 🔍 Potential Edge Cases

#### 1. Service Account Username Format Assumption

**AuthConfig line 17:**
```yaml
auth.identity.user.username.split(":")[3]
```

**Assumption:** Service account usernames are always `system:serviceaccount:<ns>:<sa>` (4 parts).

**Risk:** If Kubernetes ever changes this format or uses different SA token formats, this will break.

**Mitigation:** 
- This is a stable Kubernetes API contract (since v1.0)
- If concerned, could add validation in Authorino to check array length before indexing
- Integration tests will catch this if it ever breaks

**Recommendation:** Document this assumption in the AuthConfig YAML as a comment.

---

#### 2. Empty Tenants Handling for Visibility

**File:** `internal/auth/default_tenancy_logic.go` (lines 322-346)

`DetermineVisibleTenants` allows empty tenants (returns only `shared`), but `DetermineAssignableTenants` fails on empty tenants.

**Current behavior:**
```go
// Listing: User with no tenants can see 'shared' tenant resources
DetermineVisibleTenants() → ["shared"]  // OK

// Creating: User with no tenants cannot create resources
DetermineAssignableTenants() → error  // Fails
```

**Question:** Is there a valid scenario where a user has no tenants but should still call List operations?

**Recommendation:** This seems intentional (read-only access to shared resources), but worth confirming this is the desired behavior. If not, both methods should enforce tenants presence.

---

### 📋 Summary

| Aspect | Assessment |
|--------|------------|
| **Architecture** | ✅ Excellent — pushes complexity to auth layer |
| **Code Quality** | ✅ Massive simplification (-552 lines) |
| **Test Coverage** | ✅ Good — unit tests cover contract, integration tests cover E2E |
| **Security** | ✅ Tenant isolation validation preserved |
| **CI** | ✅ All checks passing |

---

### ✅ Recommendation

**Approve with minor suggestions:**

1. **Update `--tenancy-logic` help text** (CodeRabbit finding)
2. **Add comment to AuthConfig** documenting the service account username format assumption
3. **Verify empty tenants behavior** is intentional for `DetermineVisibleTenants`

None of these are blocking — the core refactoring is sound and well-tested. Great work on simplifying the auth layer! 🎯

eranco74 (commented):
/lgtm

internal/auth/default_tenancy_logic.go
jhernand: I will change this. However sets are passed by value here (and should always be) so the moment the variable is declared an empty set will be allocated, so the result received by the user, even in case of error, will always be a valid set, maybe empty.

jhernand: Done. Changed the code and added a unit test to verify that when an error is returned the result is also an empty set.

jhernand:
@akshaynadkarni this was merged before addressing your comments. I tried to address them here: https://github.com/osac-project/fulfillment-service/pull/408 .

PR #403, opened 2026-04-14
MGMT-23740: Default NetworkClass with auto-swap and VN auto-population

wgordon17:
## Summary
- Adds `is_default` boolean field to NetworkClass proto (field 9) with auto-swap enforcement
- Makes VirtualNetwork `network_class` optional; server populates from default when omitted
- Adds CLI table rendering YAMLs for NetworkClass and VirtualNetwork

Implements [MGMT-23740](https://redhat.atlassian.net/browse/MGMT-23740)

Follow-up: [MGMT-23869](https://redhat.atlassian.net/browse/MGMT-23869) — AAP playbook `is_default` support
Follow-up: [MGMT-23870](https://redhat.atlassian.net/browse/MGMT-23870) — osac-operator CRD and protobuf regeneration

## Summary by CodeRabbit

* **New Features**
  * NetworkClass can be marked as the default; public API shows which is default.
  * Virtual Networks created without a network_class now auto-populate from the active default.

* **Enhancements**
  * network_class is optional on create but remains immutable after set.
  * DB now enforces at most one active default NetworkClass.

* **Bug Fixes**
  * Improved validation and error messages when no default exists or defaults are inconsistent.


eranco74:
## Code Review: Default NetworkClass with Auto-Swap and VN Auto-Population

### ✅ Overall Assessment

Excellent implementation of default NetworkClass semantics with thoughtful edge-case handling and comprehensive test coverage. The auto-swap mechanism ensures the "only one default" invariant, and the VN auto-population provides great UX while maintaining backward compatibility.

Great work on this feature! The auto-population UX improvement is a nice touch. 

One question about the fallbackk  strategy
In case there's only one network_class would it be the default? or the user will get an error if they try to create a VN without passing network_class (note that this is likely the common case we are aiming at - single network class and the user not having to worry abou it)

wgordon17:
@eranco74 While we _can_ ship with either an implicit or explicit assumption of `is_default` when there's only a single NC, I think this would be considered a _surprising_ side effect for most admins (granted for _users_, the surprise would be getting an error for not specifying a NC).

Shipping an auto-selected is_default NC _could_ be something we address with our CUDN template in osac-app (https://github.com/osac-project/osac-aap/blob/main/collections/ansible_collections/osac/templates/roles/cudn_net/meta/osac.yaml) instead of encoding the behavior into the API-layer. WDYT?

eranco74:
/lgtm
please follow-up with updating the cudn-net template with is_default: true

PR #407, opened 2026-04-14
fix: propagate node_set size changes to ClusterOrder for ready clusters

trewest:
The cluster reconciler skipped all Kubernetes work when the cluster state was not PROGRESSING, so node_set size updates on READY clusters were saved to the database but never patched on the ClusterOrder CR. Allow the reconciler to sync spec changes for both PROGRESSING and READY states.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cluster synchronization logic to correctly handle additional cluster states, expanding the conditions under which cluster reconciliation proceeds.

* **Tests**
  * Added test coverage for scenarios where cluster node counts change while the cluster is in ready state.

### DIFF
--- a/internal/controllers/cluster/cluster_reconciler_function.go
+++ b/internal/controllers/cluster/cluster_reconciler_function.go
@@ -159,8 +159,12 @@ func (t *task) update(ctx context.Context) error {
 		return err
 	}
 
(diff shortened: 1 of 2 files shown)


jhernand (approved):
Looks good.

PR #410, opened 2026-04-14
NO-ISSUE: Support universal tenant sets from Authorino

jhernand:
## Summary

- Update the Authorino `AuthConfig` to emit `["*"]` as the tenant list for admin
  accounts, enabling them to access all tenants without enumerating each one.
- Change the `Subject.Tenants` field from `[]string` to `collections.Set[string]` and
  add custom JSON marshal/unmarshal logic that interprets `"*"` as the universal set.
- Adapt the `DefaultTenancyLogic` so that universal tenant sets propagate correctly
  through assignable, default, and visible tenant determination, skipping the shared
  tenant union for universal sets.

## Test plan

- [x] Unit tests for `Subject` JSON marshalling/unmarshalling (round-trips, universal
  marker, error on mixed `*` with specific tenants).
- [x] Unit tests for `DefaultTenancyLogic` covering universal tenant sets in
  assignable, default, and visible tenant determination.
- [x] Updated interceptor tests to use `collections.Set` instead of string slices.

### DIFF
--- a/internal/auth/auth_subject.go
+++ b/internal/auth/auth_subject.go
@@ -13,11 +13,80 @@ language governing permissions and limitations under the License.
 
 package auth
 
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/osac-project/fulfillment-service/internal/collections"
+)
+
 // Subject represents an entity, such as person or a service account.
 type Subject struct {
 	// User is the name of the user.
-	User string `json:"user"`
+	User string
 
-	// Tenants are the identifiers of the tenants that the subject belongs to.
+	// Tenants is the set of tenants that the subject belongs to. It may be the universal set if the subject has
+	// access to all tenants.
+	Tenants collections.Set[string]
+}
+
+// subjectJson is the JSON representation of a subject used internally for serialization and deserialization.
+type subjectJson struct {
+	User    string   `json:"user"`
 	Tenants []string `json:"tenants"`
 }
+
+// UnmarshalJSON implements custom JSON unmarshalling for Subject. It interprets the special value '*' in the tenants
+// array as meaning all tenants. This special value must not be combined with specific tenant names, that will be
+// considered as an error.
+func (s *Subject) UnmarshalJSON(data []byte) error {
+	var subject subjectJson
+	if err := json.Unmarshal(data, &subject); err != nil {
+		return err
+	}
+	s.User = subject.User
+	s.Tenants = collections.NewSet[string]()
+	universal := false
+	for _, tenant := range subject.Tenants {
+		if tenant == universalMarker {
+			universal = true
+			break
+		}
+	}
+	if universal {
+		if len(subject.Tenants) > 1 {
+			return fmt.Errorf(
+				"the universal marker '%s' cannot be combined with specific tenant names",
+				universalMarker,
+			)
+		}
+		s.Tenants = AllTenants
+	} else {
+		s.Tenants = collections.NewSet(subject.Tenants...)
+	}
+	return nil
+}
+
+// MarshalJSON implements custom JSON marshalling for Subject.
+func (s *Subject) MarshalJSON() (data []byte, err error) {
+	var tenants []string
+	if s.Tenants.Universal() {
+		tenants = []string{
+			universalMarker,
+		}
+	} else if s.Tenants.Finite() {
+		tenants = s.Tenants.Inclusions()
+	} else {
+		err = fmt.Errorf("the tenant set is infinite")
+		return
+	}
+	data, err = json.Marshal(subjectJson{
+		User:    s.User,
+		Tenants: tenants,
+	})
+	return
+}
+
+// universalMarker is the special value used in the subject's tenant list to indicate that the subject has access to all
+// tenants.
+const universalMarker = "*"
--- a/charts/service/templates/grpc-server/authconfig.yaml
+++ b/charts/service/templates/grpc-server/authconfig.yaml
@@ -66,8 +66,9 @@ spec:
         authnMethod:
           value: jwt
   authorization:
-    "fulfillment-api":
+    default:
       opa:
+        allValues: true
         rego: |
           import rego.v1
 
@@ -101,7 +102,8 @@ spec:
             input.auth.identity.authnMethod == "jwt"
           }
 
-          # Function to check if an account is an admin account:
+          # Check if an account is an admin account:
+          default is_admin = false
           is_admin if {
             subject_name in admin_service_accounts
           }
@@ -110,7 +112,8 @@ spec:
             group in admin_groups
           }
 
-          # Function to check if an account is a client account:
+          # Check if an account is a client account:
+          default is_client = false
           is_client if {
             not is_admin
           }
@@ -171,6 +174,8 @@ spec:
                     : auth.identity.user.username.split(":")[3]
               tenants:
                 expression: |
-                  auth.identity.authnMethod == "jwt"
-                    ? auth.identity.groups
-                    : [auth.identity.user.username.split(":")[2]]
+                  auth.authorization.default.is_admin
+                    ? ["*"]
+                    : auth.identity.authnMethod == "jwt"
+                      ? auth.identity.groups
+                      : [auth.identity.user.username.split(":")[2]]
--- a/internal/auth/auth_subject_test.go
+++ b/internal/auth/auth_subject_test.go
@@ -1,0 +1,151 @@
+/*
+Copyright (c) 2026 Red Hat Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with the
+License. You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific
+language governing permissions and limitations under the License.
+*/
+
+package auth
+
+import (
+	"encoding/json"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	"github.com/osac-project/fulfillment-service/internal/collections"
+)
+
+var _ = Describe("Subject", func() {
+	Describe("JSON unmarshalling", func() {
+		It("Parses user and specific tenants", func() {
+			var subject Subject
+			err := json.Unmarshal(
+				[]byte(`{
+					"user": "alice",
+					"tenants": ["t1", "t2"]
+				}`),
+				&subject,
+			)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(subject.User).To(Equal("alice"))
+			Expect(subject.Tenants.Equal(collections.NewSet("t1", "t2"))).To(BeTrue())
+		})
+
+		It("Parses asterisk as universal set", func() {
+			var subject Subject
+			err := json.Unmarshal(
+				[]byte(`{
+					"user": "admin",
+					"tenants": ["*"]
+				}`),
+				&subject,
+			)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(subject.User).To(Equal("admin"))
+			Expect(subject.Tenants.Universal()).To(BeTrue())
+		})
+
+		It("Rejects asterisk mixed with specific tenants", func() {
+			var subject Subject
+			err := json.Unmarshal(
+				[]byte(`{
+					"user": "admin",
+					"tenants": ["*", "t1"]
+				}`),
+				&subject,
+			)
+			Expect(err).To(HaveOccurred())
+			Expect(err.Error()).To(ContainSubstring("cannot be combined"))
+		})
+
+		It("Parses empty tenants as empty set", func() {
+			var subject Subject
+			err := json.Unmarshal(
+				[]byte(`{
+					"user": "alice",
+					"tenants": []
+				}`),
+				&subject,
+			)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(subject.Tenants.Empty()).To(BeTrue())
+		})
+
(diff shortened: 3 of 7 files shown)


jhernand:
@CrystalChun this may be relevant for your authentication/authorization work.

eranco74:
## Code Review: Support Universal Tenant Sets from Authorino

### ✅ Overall Assessment

Excellent implementation of universal tenant access for admin accounts using set theory semantics. The design cleanly separates admin authorization logic (Authorino) from tenant enumeration (Go code), with proper type safety via custom JSON marshalling.

**Strengths:**
- **Elegant abstraction:** Uses `collections.Set[string]` with universal set support
- **Type-safe serialization:** Custom JSON marshal/unmarshal interprets `"*"` as universal marker
- **Comprehensive tests:** 151 lines of new Subject tests + updated tenancy logic tests
- **Authorino integration:** Admin check in CEL expressions, emits `["*"]` for admins
- **All CI checks passing:** Unit tests, integration tests, format checks green

---

### 🎯 Architecture Review

#### Before: Admin Enumeration Problem

Admins had to enumerate all tenants they could access:
```go
Subject{
    User: "admin",
    Tenants: ["tenant-1", "tenant-2", ..., "tenant-N"]  // Must list all N tenants
}
```

**Problems:**
- Doesn't scale to thousands of tenants
- Requires keeping list in sync as tenants are added/removed
- No semantic difference between "admin with specific tenants" vs "admin with all tenants"

#### After: Universal Set Semantics

Admins get universal tenant access:
```go
Subject{
    User: "admin",
    Tenants: AllTenants  // Universal set (∀ tenants)
}
```

**Benefits:**
- ✅ Scales to unlimited tenants (set theory: universal set contains all elements)
- ✅ No enumeration needed (CEL emits `["*"]`, unmarshalled to `AllTenants`)
- ✅ Clear semantics (universal set is distinct from finite set)

---

### 📋 Key Changes Review

#### 1. Authorino AuthConfig CEL Logic ✅

**File:** `charts/service/templates/grpc-server/authconfig.yaml` (lines 40-47)

**Admin detection:**
```yaml
# Lines 21-28: Define is_admin predicate
default is_admin = false
is_admin {
  subject_name in admin_service_accounts
}
is_admin {
  group in admin_groups
}
```
✅ **Good:** Explicit `default is_admin = false` prevents undefined behavior

**Tenant assignment:**
```yaml
# Lines 39-47
tenants:
  expression: |
    auth.authorization.default.is_admin
      ? ["*"]
      : auth.identity.authnMethod == "jwt"
        ? auth.identity.groups
        : [auth.identity.user.username.split(":")[2]]
```

✅ **Correct logic:**
- Admin accounts → `["*"]` (universal set marker)
- JWT users → groups (finite set)
- Service accounts → namespace (single-element set)

**Authorino configuration:**
```yaml
# Lines 10-12
authorization:
  default:
    opa:
      allValues: true  # Required to expose is_admin in CEL expressions
```
✅ **Important:** `allValues: true` makes OPA policy results available in response metadata

---

#### 2. Custom JSON Marshalling ✅

**File:** `internal/auth/auth_subject.go` (lines 82-130)

**Unmarshalling `["*"]` → Universal Set:**
```go
// Lines 84-110
func (s *Subject) UnmarshalJSON(data []byte) error {
    var subject subjectJson
    if err := json.Unmarshal(data, &subject); err != nil {
        return err
    }
    s.User = subject.User
    s.Tenants = collections.NewSet[string]()
    universal := false
    for _, tenant := range subject.Tenants {
        if tenant == universalMarker {  // universalMarker = "*"
            universal = true
            break
        }
    }
    if universal {
        if len(subject.Tenants) > 1 {
            return fmt.Errorf(
                "the universal marker '%s' cannot be combined with specific tenant names",
                universalMarker,
            )
        }
        s.Tenants = AllTenants  // Universal set constant
    } else {
        s.Tenants = collections.NewSet(subject.Tenants...)
    }
    return nil
}
```

✅ **Excellent validation:** Rejects `["*", "tenant-1"]` (ambiguous semantics)

**Marshalling Universal Set → `["*"]`:**
```go
// Lines 113-130
func (s *Subject) MarshalJSON() (data []byte, err error) {
    var tenants []string
    if s.Tenants.Universal() {
        tenants = []string{universalMarker}
    } else if s.Tenants.Finite() {
        tenants = s.Tenants.Inclusions()
    } else {
        err = fmt.Errorf("the tenant set is infinite")
        return
    }
    // ...
}
```

✅ **Good:** Handles three set types (universal, finite, infinite)
⚠️ **Note:** The `infinite` case is defensive coding (can't happen in current code, but future-proof)

---

#### 3. Tenancy Logic Updates ✅

**File:** `internal/auth/default_tenancy_logic.go` (lines 61-89)

**Assignable tenants (for object creation):**
```go
// Lines 61-71
func (p *DefaultTenancyLogic) DetermineAssignableTenants(ctx context.Context) (collections.Set[string], error) {
    subject := SubjectFromContext(ctx)
    result := subject.Tenants  // Direct assignment (was: collections.NewSet(subject.Tenants...))
    if result.Empty() {
        p.logger.ErrorContext(ctx, "Subject has no tenants", slog.String("user", subject.User))
        return nil, fmt.Errorf("subject must belong to at least one tenant to create objects")
    }
    return result, nil
}
```

✅ **Correct:** If `subject.Tenants` is universal set, returns universal set (admin can create in any tenant)

**Visible tenants (for list queries):**
```go
// Lines 87-95
func (p *DefaultTenancyLogic) DetermineVisibleTenants(ctx context.Context) (collections.Set[string], error) {
    subject := SubjectFromContext(ctx)
    result := subject.Tenants
    if result.Finite() {  // Only union with SharedTenants if finite
        result = SharedTenants.Union(result)
    }
    return result, nil
}
```

✅ **Smart optimization:** Universal set already includes `SharedTenants`, so skip union
- Finite set: `{tenant-a, tenant-b} ∪ {shared} = {tenant-a, tenant-b, shared}`
- Universal set: `∀ ∪ {shared} = ∀` (no-op)

---

### 🧪 Test Coverage Assessment

**New tests in `auth_subject_test.go`:** ✅ Excellent (151 lines)

**Unmarshalling tests:**
```go
It("Parses asterisk as universal set")          // ["*"] → AllTenants
It("Rejects asterisk mixed with specific tenants")  // ["*", "t1"] → error
It("Parses empty tenants as empty set")         // [] → empty set
It("Parses missing tenants as empty set")       // omitted field → empty set
```

**Marshalling tests:**
```go
It("Marshals universal set as asterisk")        // AllTenants → ["*"]
It("Marshals specific tenants")                 // {t1} → ["t1"]
It("Round-trips specific tenants")              // Marshal→Unmarshal preserves set
It("Round-trips universal set")                 // Marshal→Unmarshal preserves universal
```

**Tenancy logic tests:**
```go
// Lines 345-357: DetermineAssignableTenants
It("Returns universal set when tenants is universal")

// Lines 383-393: DetermineDefaultTenants  
It("Returns universal set when tenants is universal")

// Lines 416-427: DetermineVisibleTenants
It("Returns universal set when tenants is universal")  // Skips SharedTenants union
```

✅ **Coverage:** All three set types (empty, finite, universal) tested across all methods

---

### 🔍 Potential Issues

#### 1. ⚠️ Breaking Change: `Subject.Tenants` Type Changed

**Before:**
```go
type Subject struct {
    User    string   `json:"user"`
    Tenants []string `json:"tenants"`
}
```

**After:**
```go
type Subject struct {
    User    string
    Tenants collections.Set[string]
}
```

**Impact:** Any code outside this PR that directly accesses `subject.Tenants` will break.

**Recommendation:** 
- Search codebase for `subject.Tenants` access patterns
- Verify all callers are updated (or already go through `DetermineAssignableTenants` etc.)

**Grep check:**
```bash
git grep 'subject\.Tenants' | grep -v _test.go
```

If there are external callers, they need migration:
```go
// Old code
for _, tenant := range subject.Tenants { ... }

// New code  
if subject.Tenants.Finite() {
    for _, tenant := range subject.Tenants.Inclusions() { ... }
} else if subject.Tenants.Universal() {
    // Handle all-tenants case
}
```

---

#### 2. ⚠️ Security: Admin Detection Relies on OPA Policy

**Observation:** Admin status determined by OPA Rego:
```rego
is_admin {
  subject_name in admin_service_accounts
}
is_admin {
  group in admin_groups
}
```

**Risk:** If OPA policy is misconfigured (e.g., `admin_groups = []`), no admins exist.

**Mitigation (current):**
- OPA policy is version-controlled in AuthConfig YAML
- Changes go through PR review

**Suggestion:** Add integration test that verifies admin accounts get `["*"]`:
```go
It("Admin service account receives universal tenant set", func() {
    // Create token for admin SA listed in admin_service_accounts
    // Call API
    // Verify subject.Tenants.Universal() == true
})
```

---

#### 3. 💡 Minor: Infinite Set Case is Unreachable

**Code:**
```go
// Line 120-122
} else {
    err = fmt.Errorf("the tenant set is infinite")
    return
}
```

**Observation:** `collections.Set[string]` only supports finite and universal sets (no infinite-but-not-universal).

**Question:** When would a set be infinite but not universal?

**Recommendation:** 
- If this is defensive coding, add comment: `// Defensive: should never happen with current Set implementation`
- If unreachable, simplify:
  ```go
  } else {
      // Universal case already handled above, finite case handled above
      panic("unreachable: set must be universal or finite")
  }
  ```

---

### 📋 Summary

| Aspect | Assessment | Notes |
|--------|------------|-------|
| **Architecture** | ✅ Excellent | Clean use of set theory for admin access |
| **Type Safety** | ✅ Excellent | Custom JSON marshal prevents invalid states |
| **Authorino Integration** | ✅ Good | CEL expressions detect admins, emit `["*"]` |
| **Test Coverage** | ✅ Excellent | 151 lines, all edge cases covered |
| **Performance** | ✅ Good | Skips `SharedTenants` union for universal sets |
| **CI Status** | ✅ All passing | Unit + integration tests green |

---

### ✅ Recommendation

**Approve with verification:**

1. **Check for breaking changes:**
   ```bash
   git grep 'subject\.Tenants' | grep -v _test.go | grep -v 'subject\.Tenants\.'
   ```
   Verify no direct slice access outside this PR.

2. **Add integration test** (optional but recommended):
   - Verify admin SA receives `["*"]` from Authorino
   - Verify admin can list/create resources in any tenant

3. **Document the infinite set case** or remove it if unreachable.

The core implementation is solid — universal sets are the right abstraction for admin access. Great work! 🎯

tzvatot (commented):
## Review: NO-ISSUE: Support universal tenant sets from Authorino

Clean design for representing admin "all tenants" access. The `collections.Set` integration is well done, the OPA policy changes are correct (adding `default` values and `allValues: true` is the right Authorino pattern), and the JSON round-trip logic handles edge cases properly. However, there's a critical security concern with the zero-value behavior of `Set[string]` when used as a tenant permission boundary.

| Category | Count |
|----------|-------|
| 🔴 Critical | 1 |
| 🟡 Important | 0 |
| 💡 Suggestion | 0 |

---

### 🔴 Critical: Zero-value `Subject.Tenants` silently grants universal access

**File:** `internal/auth/auth_subject.go` / `internal/collections/sets.go`

The zero value of `collections.Set[string]` is the universal set (`finite: false`, `items: nil`), because `Universal()` returns `!s.finite && len(s.items) == 0`. This means any `Subject` created without explicitly initializing `Tenants` silently has access to all tenants:

```go
// Old behavior (fail-safe): Tenants is nil []string
// → DetermineAssignableTenants checks len(subject.Tenants) == 0 → error "at least one tenant"
subject := &Subject{User: "bob"}

// New behavior (fail-open): Tenants is zero-value Set = universal set
// → DetermineAssignableTenants checks result.Empty() → false → returns universal set → all-tenant access
subject := &Subject{User: "bob"}
```

The JSON deserialization path is safe (it always calls `collections.NewSet[string]()` first), but any programmatic construction of `Subject` that omits `Tenants` flips from "denied" to "full access". This violates the principle of least privilege — the default for a permission boundary should be "nothing", not "everything".

**Recommended fix — use a constructor that enforces initialization:**

```go
func NewSubject(user string, tenants collections.Set[string]) *Subject {
    return &Subject{
        User:    user,
        Tenants: tenants,
    }
}
```

And add a defensive check at the top of `DetermineAssignableTenants`:

```go
func (p *DefaultTenancyLogic) DetermineAssignableTenants(ctx context.Context) (result collections.Set[string], err error) {
    subject := SubjectFromContext(ctx)
    result = subject.Tenants
    // Guard against zero-value Set being the universal set
    if result.Universal() && result == (collections.Set[string]{}) {
        err = fmt.Errorf("subject tenants not initialized")
        return
    }
    // ...
}
```

Alternatively, consider whether `Set[T]`'s zero value should be the empty set instead of the universal set, since that's safer for permission use cases. This would require changing the `Set` struct to default `finite` to `true`, but that's a broader change to evaluate.

jhernand:
> The zero value of `collections.Set[string]` is the universal set (`finite: false`, `items: nil`),

@tzvatot alter ego, that was fixed in https://github.com/osac-project/fulfillment-service/pull/404 .

internal/auth/auth_subject.go
```diff
@@ -13,11 +13,80 @@ language governing permissions and limitations under the License.
...
+	data, err = json.Marshal(subjectJson{
+		User:    s.User,
+		Tenants: tenants,
+	})
+	return
+}
+
+// universalMarker is the special value used in the subject's tenant list to indicate that the subject has access to all
+// tenants.
+const universalMarker = "*"
```
CrystalChun, line 92: Sorry may be a naïve question, but is this meant for the cloud provider admin who can create and see all tenants (orgs)?

jhernand: Yes, in general `["*"]` indicates that the user has visibility of all the tenants, and that is currently done only for the `admin` users. Any other user will receive a finite list of tenants extracted from the `groups` claim of the JWT. The objective of this is to be able to change all this logic by just configuring Authorino and not having to change the service.

CrystalChun: Thank you for the explanation!

PR #414, opened 2026-04-15
MGMT-23769: Provide up front validation and handling of template prov…

DakCrowder:
Adds validation and initialization of compute instances using fetched template spec default values.

The gist of this is we fetch the template (was already occurring), but the template we get back now includes spec defaults.  We take the compute instance request body, populate fields on the request body from the spec defaults (such as cpu cores, memory etc.) and then validate the combined payload.  The result is users can get additional feedback on fields they must define if the template doesn't define them, and we populate the initial values from the spec defaults if the user doesn't provide specific values.

For example, if we had a template that had NO defaults defined, the user can now see they need to define fields such as cores, memory etc. at creation time rather than having to debug why their VM won't spin up later on.
```
osac create computeinstance --name test --template test-no-defaults
Error: failed to create compute instance: rpc error: code = InvalidArgument desc = the following required spec fields are missing: boot_disk, cores, image, memory_gib, run_strategy
```

Additionally if users specify a template with the defaults defined those defaults will be used to populate the computinstance.
```
# Populates all values from the template defined defaults
osac create computeinstance --name test --template osac.templates.ocp_virt_vm
```

User values always have precedence - so in a case like the following the user passed cores and memory will override the template default, while the other values will be sourced from the template.
```
# Populates cores and memory from the passed user value, and the rest of the values from template defined defaults
osac create computeinstance --name test --template osac.templates.ocp_virt_vm --cores 4 --memory-gib 16
``` 

Additionally if a template does NOT define a value (e.g. lets say cores is NOT specified in the default spec values), the user then must provide that value.
```
# If a template does not define cores and a user does NOT pass them we fail
osac create computeinstance --name test --template no-cores-defined-in-default-spec
Error: failed to create compute instance: rpc error: code = InvalidArgument desc = the following required spec fields are missing: cores
# If a template does not define cores and a user does pass them it succeeds
osac create computeinstance --name test --template no-cores-defined-in-default-spec --cores 4
``` 

Requires https://github.com/osac-project/osac-aap/pull/246 to fetch the default fields from templates

## Summary by CodeRabbit

* **New Features**
  * Templates can include spec defaults (cores, memory, image, boot disk, run strategy) that are applied when creating instances.

* **Improvements**
  * Create/update flow applies template defaults and validates required fields; user-provided spec values take precedence. Validation aggregates missing-field errors and rejects invalid run strategy values.

* **Tests**
  * Added unit and integration tests covering defaulting, validation, persistence semantics, and error messages.


internal/utils/spec_defaults.go
```diff
@@ -0,0 +1,106 @@
...
+	grpccodes "google.golang.org/grpc/codes"
+	grpcstatus "google.golang.org/grpc/status"
+	"google.golang.org/protobuf/proto"
+
+	privatev1 "github.com/osac-project/fulfillment-service/internal/api/osac/private/v1"
+)
+
+// validRunStrategies contains the run strategy values accepted by the Kubernetes ComputeInstance CRD.
+// Note: these values are case-sensitive as currently no normalization is performed.
+var validRunStrategies = []string{"Always", "Halted"}
```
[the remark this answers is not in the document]
DakCrowder: The CRD this maps to does not implement all of those values currently - see https://github.com/osac-project/osac-operator/blob/main/config/crd/bases/osac.openshift.io_computeinstances.yaml#L149

internal/testing/database.go
```diff
@@ -119,7 +119,7 @@ func MakeDatabaseServer() *DatabaseServer {
 	)
 	handle, err := sql.Open("pgx", url)
 	Expect(err).ToNot(HaveOccurred())
-	Eventually(handle.Ping, 10, 1).ShouldNot(HaveOccurred())
+	Eventually(handle.Ping, 30, 1).ShouldNot(HaveOccurred())
```
DakCrowder, line 122: My local environment (on my poor little thinkpad) was consistently timing out here running tests, bumped this up due to hitting it.

akshaynadkarni:
@DakCrowder 
Nice work on the validation layer, the `ApplySpecDefaults` / `ValidateRequiredSpecFields` split is clean and well-tested. I had a question about the end-to-end flow though.

I see the tests at L650-675 assert that template defaults are intentionally **not** stored on the object ("the template is the source of truth at render time"). I understand the reasoning, but I'm wondering how this works with the current CRD and controller.

When the controller reconciles, [`addExplicitFields()`](https://github.com/osac-project/fulfillment-service/blob/1185cb26dc0f3fd2b742d414962d2cf8b42c5231/internal/controllers/computeinstance/computeinstance_reconciler_function.go#L539-L579) only sets fields on the K8s CR where `Has*()` returns true. Since defaults aren't persisted, those checks return false, and the CR is created without `cores`, `memoryGiB`, etc. The CRD marks all five as [required](https://github.com/osac-project/osac-operator/blob/main/config/crd/bases/osac.openshift.io_computeinstances.yaml#L195-L201), so the K8s API server would reject it before the osac-operator/AAP can apply its own defaults from `create_validate.yaml`.

The ticket ([MGMT-23769](https://redhat.atlassian.net/browse/MGMT-23769)) has two expectations: (1) defaults populated automatically from the template, and (2) clear error when info is missing. This PR addresses (2) well. For (1), is the plan to also update the controller to merge template defaults in `buildSpec()`, or to relax the CRD to make those fields optional? Just want to make sure I'm not missing a companion change.

internal/servers/private_compute_instances_server.go
```diff
@@ -238,36 +268,33 @@ func (s *PrivateComputeInstancesServer) validateTemplate(ctx context.Context, vm
 			slog.String("template_id", templateID),
 			slog.Any("error", err),
 		)
-		return grpcstatus.Errorf(
+		return nil, grpcstatus.Errorf(
 			grpccodes.Internal,
 			"failed to retrieve template '%s'",
 			templateID,
 		)
```
akshaynadkarni, line 282: `fetchTemplate` maps all DAO errors to `grpccodes.Internal` here. When the template doesn't exist, the DAO returns `*dao.ErrNotFound`, which is a client error (user referenced an invalid template). The pattern in [`validateNetworkReferences` L344-349](https://github.com/osac-project/fulfillment-service/blob/1185cb26dc0f3fd2b742d414962d2cf8b42c5231/internal/servers/private_compute_instances_server.go#L344-L349) shows how to check for this:

```go
var notFoundErr *dao.ErrNotFound
if errors.As(err, &notFoundErr) {
    return nil, grpcstatus.Errorf(grpccodes.InvalidArgument,
        "template '%s' does not exist", templateID)
}
```

Without this, users get a 500 instead of a 400 when referencing a non-existent template. Also, the nil check at L277-284 appears to be dead code since the DAO returns `ErrNotFound` before it could return a nil object.

internal/utils/spec_defaults.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+	if !spec.HasImage() {
+		missing = append(missing, "image")
+	}
+	if !spec.HasBootDisk() {
+		missing = append(missing, "boot_disk")
+	}
+	if !spec.HasRunStrategy() {
+		missing = append(missing, "run_strategy")
+	}
```
akshaynadkarni, line 109: Minor: `ValidateRequiredSpecFields` checks `HasImage()` and `HasBootDisk()` but not whether the inner fields are valid. For example, `image: {}` (empty message) would pass `HasImage()` but has no `source_ref` or `source_type`. The CRD requires both ([`sourceRef` and `sourceType` are required on the image object](https://github.com/osac-project/osac-operator/blob/main/config/crd/bases/osac.openshift.io_computeinstances.yaml#L118-L120)), so this would still fail downstream. Not necessarily blocking, but worth noting as a gap if "upfront validation" is the goal.

internal/servers/private_compute_instances_server_test.go
```diff
@@ -634,6 +646,194 @@ var _ = Describe("Private compute instances server", func() {
 			Expect(err).To(HaveOccurred())
 			Expect(response).To(BeNil())
 		})
+
+		It("Validates spec defaults from template without storing them", func() {
+			createTemplate("defaults-template")
+
+			// Create a compute instance without any spec fields — validation should pass
+			// because template defaults cover all required fields, but defaults should NOT
+			// be stored on the object (the template is the source of truth at render time):
```
akshaynadkarni: Curious about the design intent here. The test asserts defaults are NOT stored, with the comment "the template is the source of truth at render time." Could you help me understand how the controller handles this case? Looking at [`addExplicitFields()`](https://github.com/osac-project/fulfillment-service/blob/1185cb26dc0f3fd2b742d414962d2cf8b42c5231/internal/controllers/computeinstance/computeinstance_reconciler_function.go#L539-L579), it only populates fields where `Has*()` is true, so the K8s CR would be created without these spec fields. Since the [CRD requires them](https://github.com/osac-project/osac-operator/blob/main/config/crd/bases/osac.openshift.io_computeinstances.yaml#L195-L201), would the CR creation fail?

DakCrowder: Yes good observation / ty for raising.  I had been waffling on the right approach here but I was missing a piece of the puzzle - which is the vendored mass open cloud aap arg spec has additional fields defined as template_parameters so I was seeing things populated in different ways that i was expecting.

After some more thought, I think there is a simpler option here - inspired by [this comment](https://github.com/osac-project/osac-aap/pull/246#discussion_r3101737612) on the associated aap pr I wonder if the defaults should just be template parameters with some well-known name or flag to indicate they are spec fields, we use the info defined (e.g. required, default value etc.) from the template parameter field definition to perform the validation, and then save them as a part of that block (continues behavior as-is of the  template parameters are saved).

DakCrowder: I guess one difference / potential issue there is the template parameters only support a "flat" structure in comparison to nested spec fields like the image / boot_disk.

internal/utils/spec_defaults.go
DakCrowder: Added some additional checks to provide better validation as highlighted

internal/servers/private_compute_instances_server_test.go
DakCrowder: To follow up on this - the code is now taking a snapshot of template default values _at creation time_ (which is how we handle template_parameters) and store them - ensuring the fields are populated before the CR creation runs in the reconciler.

internal/servers/private_compute_instances_server.go
```diff
@@ -189,7 +195,7 @@ func (s *PrivateComputeInstancesServer) Update(ctx context.Context,
 		}
 	}
 	if hasMaskPrefix(mask, "spec.template", "spec.template_parameters") {
```
akshaynadkarni: Changing `spec.template` is allowed on update, but only `fetchAndValidateTemplate` runs here (template parameter validation). `applySpecDefaults` is not called, so the new template's spec defaults won't be applied for any fields the user hasn't explicitly set. The old template's defaults remain on the CI.

Two options:
1. Call `applySpecDefaults` here too, so the new template's defaults take effect on update.
2. Block template changes post-creation (make `spec.template` immutable after create), if switching templates on a running CI isn't intended to be supported.

Either way, the current behavior (silently keeping the old template's defaults after a template change) seems like it could surprise users.

internal/utils/spec_defaults.go
```diff
@@ -0,0 +1,146 @@
...
+	if !spec.HasCores() && defaults.HasCores() {
+		spec.SetCores(defaults.GetCores())
+	}
+	if !spec.HasMemoryGib() && defaults.HasMemoryGib() {
+		spec.SetMemoryGib(defaults.GetMemoryGib())
+	}
+	if !spec.HasRunStrategy() && defaults.HasRunStrategy() {
+		spec.SetRunStrategy(defaults.GetRunStrategy())
+	}
+	if !spec.HasImage() && defaults.HasImage() {
```
akshaynadkarni: The defaulting for `image` and `boot_disk` is all-or-nothing: if the user provides any part of the nested object (e.g., `--image "quay.io/my-image:latest"` sets `source_ref`), `HasImage()` returns true and the entire default is skipped, including `source_type`.

This means `osac create computeinstance --name test --template osac.templates.ocp_virt_vm --image "quay.io/my-image:latest"` fails with `"the following required image fields are missing: image.source_type"`, even though the template default has `source_type: "registry"` (the only valid value per the CRD enum).

A field-level merge for nested objects would be more user-friendly:

```go
if defaults.HasImage() {
    if !spec.HasImage() {
        spec.SetImage(proto.Clone(defaults.GetImage()).(*privatev1.ComputeInstanceImage))
    } else {
        img := spec.GetImage()
        defImg := defaults.GetImage()
        if img.GetSourceType() == "" && defImg.GetSourceType() != "" {
            img.SetSourceType(defImg.GetSourceType())
        }
        if img.GetSourceRef() == "" && defImg.GetSourceRef() != "" {
            img.SetSourceRef(defImg.GetSourceRef())
        }
    }
}
```

Perhaps worth considering since `source_type` only has one valid value today.

internal/servers/private_compute_instances_server.go
DakCrowder: I was punting on update until I could get more clarity BUT after looking at clusters there is a method ```validateTemplateImmutability``` that (as named) ensures the corresponding fields on clusters cannot be changed on objects.  Additionally, the [compute_instace_type.proto](https://github.com/osac-project/fulfillment-service/blob/main/proto/public/osac/public/v1/compute_instance_type.proto#L57) indicates the template and template parameters should be immutable.  Given this I think matching behavior can be introduced here aligned with option **2** above.  If we want to change this behavior later then we can take it on more holistically alongside clusters / how we view templates as a whole.

DakCrowder: Updated to add the same behavior and ensures immutability as the proto def indicates

akshaynadkarni (approved):
Code changes LGTM. 

Prior to merging, can you update the PR description? It will be good to include any test results etc. so it's clear what the new behavior will be. Thanks.

PR #417, opened 2026-04-16
MGMT-23785: Create organizations server

CrystalChun:
# Description
Initial creation of the Organizations server. 

# Testing 
Unit testing
```
go test -v ./internal
```

# Additional details

Actually using this relies on deploying Keycloak which will be added in a different PR.

Future PRs to fully manage Organizations through the identity provider (Keycloak for our use-case), we'll use the identity provider client to perform CRUD operations on Organizations (realms). This will allow Organizations to be managed directly in the identity provider, while providing an API client from OSAC to manage them.

Future PR to refactor of the BreakGlassCredentials: Creates separate identity type to define BreakGlassCredentials so they can be returned as part of the response when an Organization is created as opposed to stored in the DB where it's vulnerable.

Eventually should be able to test with

Organization APIs are only accessible by the admin user.

```sh
# Regular user
export token=$(kubectl create token -n osac client)
host=$(kubectl get route -n osac fulfillment-api -o json | jq -r '.spec.host')

# Should be able to list things normally
~/go/bin/grpcurl -insecure -H "Authorization: Bearer ${token}" ${host}:443 osac.public.v1.ClusterTemplates/List
# shouldn't be able to get orgs
~/go/bin/grpcurl -insecure -H "Authorization: Bearer ${token}" ${host}:443 osac.public.v1.Organizations/Get

# Testing organizations (admin access only)
export ADMIN_TOKEN=$(oc create token admin -n osac --duration=24h)

~/go/bin/grpcurl -insecure \
-H "Authorization: Bearer ${ADMIN_TOKEN}" \
${host}:443 \
osac.public.v1.Organizations/List
```

## Summary by CodeRabbit

# Release Notes

* **New Features**
  * Added Organizations API for managing organizations in the service.
  * Integrated optional Identity Provider support (Keycloak) for organization administration.
  * Break-glass credentials now provided during organization creation for administrative access.

* **Documentation**
  * Added Identity Provider integration guide with Keycloak setup instructions.
  * Documented Organizations API usage and configuration options.

### DIFF
--- a/internal/database/migrations/28_create_organizations_tables.up.sql
+++ b/internal/database/migrations/28_create_organizations_tables.up.sql
@@ -1,0 +1,54 @@
+--
+-- Copyright (c) 2026 Red Hat Inc.
+--
+-- Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+-- the License. You may obtain a copy of the License at
+--
+--   http://www.apache.org/licenses/LICENSE-2.0
+--
+-- Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+-- an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+-- specific language governing permissions and limitations under the License.
+--
+
+-- Create the organizations tables:
+--
+-- This migration establishes the database schema for Organization resources following the generic schema pattern.
+--
+
+create table organizations (
+  id text not null primary key,
+  name text not null default '',
+  creation_timestamp timestamp with time zone not null default now(),
+  deletion_timestamp timestamp with time zone not null default 'epoch',
+  finalizers text[] not null default '{}',
+  creators text[] not null default '{}',
+  tenants text[] not null default '{}',
+  labels jsonb not null default '{}'::jsonb,
+  annotations jsonb not null default '{}'::jsonb,
+  data jsonb not null,
+  version integer not null default 0
+);
+
+create table archived_organizations (
+  id text not null,
+  name text not null default '',
+  creation_timestamp timestamp with time zone not null,
+  deletion_timestamp timestamp with time zone not null,
+  archival_timestamp timestamp with time zone not null default now(),
+  finalizers text[] not null default '{}',
+  creators text[] not null default '{}',
+  tenants text[] not null default '{}',
+  labels jsonb not null default '{}'::jsonb,
+  annotations jsonb not null default '{}'::jsonb,
+  data jsonb not null,
+  version integer not null default 0
+);
+
+create index organizations_by_name on organizations (name);
+
+create index organizations_by_owner on organizations using gin (creators);
+
+create index organizations_by_tenant on organizations using gin (tenants);
+
+create index organizations_by_label on organizations using gin (labels);
--- a/internal/servers/private_organizations_server_test.go
+++ b/internal/servers/private_organizations_server_test.go
@@ -1,0 +1,181 @@
+/*
+Copyright (c) 2026 Red Hat Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with the
+License. You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific
+language governing permissions and limitations under the License.
+*/
+
+package servers
+
+import (
+	"context"
+
+	"github.com/jackc/pgx/v5/pgxpool"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	privatev1 "github.com/osac-project/fulfillment-service/internal/api/osac/private/v1"
+	"github.com/osac-project/fulfillment-service/internal/database"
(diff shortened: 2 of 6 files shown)


CrystalChun:
/retitle [MGMT-23785](https://redhat.atlassian.net/browse/MGMT-23785): Create Organizations server and hook in Organization management through Keycloak

internal/database/migrations/28_create_organizations_tables.up.sql
DakCrowder, line 1: I think we should get in the habit of also including the golang-migrate supported down migration files - it looks like it hasn't been consistently enforced (and based off my cursory looking we don't have anything that runs the down migrations) but might make troubleshooting in the future easier and allow for the migrate cli to easily switch between versions if we need to debug at some point.

Additionally, it looks like migrations versions currently in `main` go from 1-17, there is no 18, then it goes up to 27.  Has 18 been somehow previously reserved for this?  If not I think it might be "more" correct to go to 28 but not 100% sure how the tool handles this.

internal/cmd/service/start/grpcserver/start_grpc_server_cmd.go
```diff
@@ -96,6 +99,38 @@ func Cmd() *cobra.Command {
 		"default",
 		"Type of tenancy logic to use. Valid values are 'default' and 'guest'.",
 	)
+	flags.StringVar(
+		&runner.args.keycloakURL,
+		"keycloak-url",
```
DakCrowder: A couple comments here related to the keycloak flags:

1. Are we planning on supporting a certain range of identity providers in the future, or only keycloak?  I'm wondering if the `keycloak-` prefix on these could be removed or replaced with something more generic.   Not a big deal though as my assumption is this isn't really a public api and we could change later on.
2.  One thought is some of these might be better suited as env vars or files.  The flags couple the credentials with the initial invocation while the others would allow for creds to be swapped out dynamically if a user wished down the line.  And would allow for users to easily mount the secret or inject to the container.

internal/cmd/service/start/grpcserver/start_grpc_server_cmd.go
```diff
@@ -783,6 +823,97 @@ func (c *runnerContext) run(cmd *cobra.Command, argv []string) error {
...
+			Build()
+		if err != nil {
+			return fmt.Errorf("failed to create Keycloak token store: %w", err)
+		}
+
+		// Create a token source for Keycloak authentication using password flow
+		keycloakIssuer := fmt.Sprintf("%s/realms/%s", c.args.keycloakURL, c.args.keycloakRealm)
+		keycloakTokenSource, err := oauth.NewTokenSource().
+			SetLogger(c.logger).
+			SetFlow(oauth.PasswordFlow).
```
DakCrowder: Is it a requirement to use the password flow, or could a client credentials grant accomplish the same outcomes?  iirc keycloak exposes service accounts w/ client id / client secrets meant for machine to machine authentication.  That feels more "standard" to me for this use case rather than a user account, however I might be missing some additional context wrt capabilities each provides.  We appear to have an oauth credentials flow also defined in the repo so I don't _think_ it would be a huge lift but would definitely affect user facing documentation / behavior as they would have to manage that service account.

internal/cmd/service/start/grpcserver/start_grpc_server_cmd.go
CrystalChun: Very great points! Thank you for bringing this up! I agree that this should be more generic in the event that there are other identity providers, and good idea to use a environment variables so there can be dynamic changes.

I've made the changes to reflect this, how does it look?

internal/database/migrations/28_create_organizations_tables.up.sql
CrystalChun: Thank you for letting me know! I've added a down-migration file for organizations, but let me ask the team channel about the numbering. I'll adjust the numbers depending on the answer!

internal/cmd/service/start/grpcserver/start_grpc_server_cmd.go
DakCrowder: Looks great ty!

internal/database/migrations/28_create_organizations_tables.up.sql
CrystalChun: Thank you again for bringing this up! I've moved it to number 28 based off of the discussion.

internal/servers/private_organizations_server_test.go
```diff
@@ -0,0 +1,562 @@
...
+	return nil
+}
+
+func (m *mockClient) GetUserOrganizationRoles(ctx context.Context, organization, userID string) ([]*idp.Role, error) {
+	return nil, nil
+}
+
+func (m *mockClient) GetUserClientRoles(ctx context.Context, organization, userID, clientID string) ([]*idp.Role, error) {
+	return nil, nil
+}
```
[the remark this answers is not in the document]
CrystalChun: This is already fixed

jhernand:
I have two major suggestions on the approach, which you may just ignore:

1. We currently use command line flags to configure the services, not environment variables. I'd very much prefer to keep doing that unless it is impossible to do it. Flags for some things and environment variables for other things would be really confusing.

2. For other things, when we need to integrate with other systems (HypeShift, KubeVirt, etc) we use the reconciler mode. Can we use that same model for organizations? You could extend the controller component so that it watches for changes in organizations and talks to Keycloak to do whatever is needed. That would make for a more consistent approach, and would be more reliable in the sense that the reconciliation can be periodically retried.

CrystalChun:
/retitle [MGMT-23785](https://redhat.atlassian.net/browse/MGMT-23785): Create organizations server

jhernand (approved):

PR #418, opened 2026-04-16
MGMT-23786: Add users and access key APIs

CrystalChun:
Extending the API of the identity provider to allow OSAC users to create/manage other users and access keys within their tenant is an important component of multi-tenancy. This PR only defines the APIs of these objects, and a later change will be included to create a server for the APIs and hook them into the identity provider's APIs

New APIs

```
  Public API (osac.public.v1)                                                   
                                                                                
  Users Service:                                                                
  - List - List users in an organization (supports filter, offset, limit, order)
  - Get - Retrieve a specific user                                              
  - Create - Create a new user                                                
  - Update - Update an existing user (with field mask and optimistic locking)   
  - Delete - Delete a user                                                      
                                                                                
  AccessKeys Service:                                                           
  - List - List access keys for a user                                          
  - Get - Retrieve a specific access key                                        
  - Create - Create a new access key (returns credentials on creation)
  - Disable - Disable an access key (prevents authentication)                   
  - Enable - Re-enable a disabled access key                                    
  - Delete - Delete an access key                                               
                                                                                
  Private API (osac.private.v1)                                                 
                                                                                
  Users Service:                                                                
  - List - List users (no order parameter)                                      
  - Get - Retrieve a specific user                                              
  - Create - Create a new user                                                
  - Update - Update an existing user                                            
  - Delete - Delete a user                                                      
                                                                                
  AccessKeys Service:                                                           
  - List - List access keys                                                     
  - Get - Retrieve a specific access key                                      
  - Create - Create a new access key (returns credentials)
  - Disable - Disable an access key                       
  - Enable - Enable an access key                                               
  - Delete - Delete an access key
```

Assisted-by: Claude

## Summary by CodeRabbit

* **New Features**
  * Added user management APIs and data models for create, retrieve, list, update, delete, and signaling of users.
  * Added access key management APIs and data models for create (returns credentials at creation), retrieve, list, enable, disable, delete, and signaling of access keys.

* **Tests**
  * Expanded tests to include "user"/"users" in reflected type expectations.


proto/public/osac/public/v1/user_type.proto
```diff
@@ -0,0 +1,94 @@
...
+// UserCondition represents a condition of a user.
+message UserCondition {
+  // Type of the condition.
+  string type = 1;
+
+  // Status of the condition (True, False, Unknown).
+  string status = 2;
+
+  // Last time the condition was updated.
+  string last_transition_time = 3;
```
akshaynadkarni: The `last_transition_time` field in `UserCondition` is declared as `string`, but the established pattern in this codebase uses `google.protobuf.Timestamp` for all timestamp fields. For example, `ClusterCondition.last_transition_time` in `cluster_type.proto` is `google.protobuf.Timestamp` (line 225), and `Metadata.creation_timestamp`/`deletion_timestamp` also use `Timestamp`.

The same applies to `AccessKeyStatus.last_used_time` (also `string` in both public and private protos).

Using `Timestamp` provides type safety, automatic parsing/formatting in generated code, and consistent serialization. Since the server implementation is coming in a follow-up PR, changing this now avoids a breaking change later.

```suggestion
  google.protobuf.Timestamp last_transition_time = 3;
```

Would need to add `import "google/protobuf/timestamp.proto";` to both `user_type.proto` and `access_key_type.proto` (public and private).

proto/public/osac/public/v1/user_type.proto
```diff
@@ -0,0 +1,94 @@
...
+  repeated UserCondition conditions = 2;
+}
+
+// UserCondition represents a condition of a user.
+message UserCondition {
+  // Type of the condition.
+  string type = 1;
+
+  // Status of the condition (True, False, Unknown).
+  string status = 2;
```
akshaynadkarni: `UserCondition.status` is `string` with a comment saying it holds "True, False, Unknown". The codebase already defines a `ConditionStatus` enum in `condition_status_type.proto` with `CONDITION_STATUS_TRUE`, `CONDITION_STATUS_FALSE`, and `CONDITION_STATUS_UNSPECIFIED` values. All existing condition types (e.g., `ClusterCondition`) use this enum.

Using the enum provides compile-time validation and consistency with the existing condition pattern across the codebase.

```suggestion
  ConditionStatus status = 2;
```

Would need to add `import "osac/public/v1/condition_status_type.proto";` to the user type proto. The private proto can import the public enum (same as it already does for `Metadata`).

proto/public/osac/public/v1/user_type.proto
```diff
@@ -0,0 +1,94 @@
...
+}
+
+// UserStatus contains the observed state of the user.
+message UserStatus {
+  // The user's current state in the identity provider.
+  string phase = 1;
+
+  // Additional conditions about the user's state.
+  repeated UserCondition conditions = 2;
+}
```
akshaynadkarni, line 67: `UserStatus.phase` is a `string`, while all existing resource types use enums for state/phase (e.g., `ClusterState` enum for `ClusterStatus.state`, `ComputeInstanceState` for compute instances). Similarly, `UserCondition.type` is `string`, while existing conditions use enums like `ClusterConditionType`.

Is this intentional because the phase/condition type values haven't been defined yet, or should enums like `UserState` and `UserConditionType` be introduced now? If the values aren't known yet, it might be worth adding a TODO comment, since converting from `string` to `enum` later is a potentially breaking change (different wire representation for JSON).

Same question applies to `AccessKeyStatus.phase`.

proto/private/osac/private/v1/users_service.proto
```diff
@@ -0,0 +1,100 @@
...
+    option (google.api.http) = {delete: "/api/private/v1/organizations/{organization_id}/users/{id}"};
+  }
+
+  rpc Update(UsersUpdateRequest) returns (UsersUpdateResponse) {
+    option (google.api.http) = {
+      patch: "/api/private/v1/organizations/{object.spec.organization_id}/users/{object.id}"
+      body: "object"
+      response_body: "object"
+    };
+  }
```
akshaynadkarni, line 115: Both private Users and AccessKeys services are missing the `Signal` RPC that every other private service defines. Signal is needed on the delete path since Delete is a soft delete and external controllers call Signal to notify the fulfillment-service when finalizer cleanup is complete.

Is the omission intentional, or should Signal be added now to keep the interface complete?

proto/public/osac/public/v1/user_type.proto
```diff
@@ -0,0 +1,94 @@
...
+  // Status of the condition (True, False, Unknown).
+  string status = 2;
+
+  // Last time the condition was updated.
+  string last_transition_time = 3;
+
+  // Reason for the condition's last transition.
+  string reason = 4;
+
+  // Human-readable message about the transition.
```
akshaynadkarni, line 83: `UserCondition.reason` and `message` are plain `string`, but every other condition type in the codebase uses `optional string` for these fields. Is this intentional?

```suggestion
  // Reason for the condition's last transition.
  optional string reason = 4;

  // Human-readable message about the transition.
  optional string message = 5;
```

proto/public/osac/public/v1/user_type.proto
CrystalChun: Thanks so much for letting me know, Akshay! I didn't realize this was the standard in the repo (still a bit new), but very much appreciate the information! 

I've modified both access keys and users to use the `Timestamp` type for the `last_used_time` field.

proto/public/osac/public/v1/user_type.proto
CrystalChun: Great to know! Thank you Akshay!!

I've modified this to use the predefined `ConditionStatus`.

proto/private/osac/private/v1/users_service.proto
CrystalChun: This is great to know, thank you so much for the explanation! I've added the `Signal` RPC to each of these.

proto/public/osac/public/v1/user_type.proto
CrystalChun: Not intentional, just unaware 😅 Thank you for letting me know! I've changed them to be optional!

proto/public/osac/public/v1/user_type.proto
CrystalChun: Thanks for pointing that out! The phase/condition hasn't entirely been defined yet, would it be better to leave the `phase` field out for now? Would it also be a breaking change if it's left out now and added later?

akshaynadkarni: Keeping `string phase` is fine for now. Since we haven't released yet, we can always change the type later without breaking anything.

proto/private/osac/private/v1/access_keys_service.proto
```diff
@@ -0,0 +1,129 @@
...
+      post: "/api/private/v1/organizations/{object.spec.organization_id}/users/{object.spec.user_id}/access-keys"
+      body: "object"
+      response_body: "object"
+    };
+  }
+
+  rpc Disable(AccessKeysDisableRequest) returns (AccessKeysDisableResponse) {
+    option (google.api.http) = {
+      post: "/api/private/v1/organizations/{organization_id}/users/{user_id}/access-keys/{id}:disable"
+      body: "*"
```
akshaynadkarni, line 111: (question): Do we need a `body` here?

proto/private/osac/private/v1/access_keys_service.proto
```diff
@@ -0,0 +1,129 @@
...
+      response_body: "object"
+    };
+  }
+
+  rpc Enable(AccessKeysEnableRequest) returns (AccessKeysEnableResponse) {
+    option (google.api.http) = {
+      post: "/api/private/v1/organizations/{organization_id}/users/{user_id}/access-keys/{id}:enable"
+      body: "*"
```
akshaynadkarni, line 119: Same comment as [above](https://github.com/osac-project/fulfillment-service/pull/418/changes#r3113756538)

Just curious.

akshaynadkarni (approved):
Thanks @CrystalChun for addressing all the comments. And it's ok to not know about what already exists in the repo. I feel the same way. It's all good 😄. 

I have left a couple of questions, but overall these changes LGTM.

PR #419, opened 2026-04-17
NO-ISSUE: Display identifier if object doesn't have a name

jhernand:
## Summary

- When rendering table columns that use the lookup feature, if the referenced object
  doesn't have a name, the cell was left empty. This change falls back to displaying
  the identifier itself instead of an empty cell.

## Test plan

- Run `fulfillment-cli get computeinstances` (or any resource with lookup columns) where
  a referenced object has no name and verify the identifier is displayed instead of a blank.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a display issue where table cells could appear blank when lookup results were unavailable. The system now properly displays the original key as a fallback instead of showing empty values.

### DIFF
--- a/internal/rendering/table_renderer.go
+++ b/internal/rendering/table_renderer.go
@@ -481,6 +481,9 @@ func (r *TableRenderer) lookupName(ctx context.Context, messageFullName protoref
 	object := listResult.Items[0]
 	metadata := helper.GetMetadata(object)
 	result = metadata.GetName()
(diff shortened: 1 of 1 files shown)


jhernand:
@trewest this fixes the issue where you didn't see the template associated to clusters.

jhernand:
Replaced by https://github.com/osac-project/fulfillment-service/pull/420 .

PR #421, opened 2026-04-17
MGMT-23918: Fix Update without FieldMask clearing optional spec fields

zszabo-rh:
## Summary

- Add `GenericMapper.CopyUpdate()` — a new mapping method with correct Update semantics: absent fields are preserved (like Merge), while present list and map fields are replaced entirely (like Copy)
- Switch all 9 public server Update paths from `Copy()` to `CopyUpdate()`
- Update Cluster label-clearing test to use FieldMask (proto3 cannot distinguish empty from absent)

## Context

Public server `Update` methods used `GenericMapper.Copy()` to map incoming public objects onto existing private objects. `Copy()` calls `to.Clear()` for any field where `!from.Has()`, silently erasing optional spec fields (cores, memory_gib, run_strategy, image, boot_disk, etc.) not included in the update request.

Neither `Copy` nor `Merge` is correct for Updates:
- **Copy**: clears absent optional fields (data loss)
- **Merge**: appends to lists instead of replacing (data corruption)

`CopyUpdate` combines both: `continue` for absent fields (Merge behavior), `copyList` for present lists (Copy behavior).

## Affected servers

All public servers with an Update path: ComputeInstances, Clusters, SecurityGroups, Subnets, VirtualNetworks, ClusterTemplates, ComputeInstanceTemplates, HostTypes, NetworkClasses.

## Test plan

- [x] 5 GenericMapper unit tests covering: preserve absent optionals, replace present lists, preserve absent lists, overwrite present optionals, mixed scenario
- [x] ComputeInstance server integration test: create with optional fields, update without them, verify preservation through DB round-trip
- [x] Full server test suite: 431/431 passing
- [x] `go build ./...` clean

## Behavioral change

Updates without FieldMask no longer clear absent map/list fields. To explicitly clear a collection field, use FieldMask. This aligns with proto3 semantics where empty collections are indistinguishable from absent fields.

Generated-By: Claude Code (Anthropic)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed partial update behavior across multiple resource types. When updating resources without specifying all fields, unspecified fields are now preserved instead of being cleared. This applies to Clusters, Compute Instances, Host Types, Network Classes, Security Groups, Subnets, Virtual Networks, and related templates.

* **Tests**
  * Added test coverage validating correct preservation of optional fields during partial updates.

### DIFF
--- a/internal/servers/cluster_templates_server.go
+++ b/internal/servers/cluster_templates_server.go
@@ -272,8 +272,8 @@ func (s *ClusterTemplatesServer) Update(ctx context.Context,
 	}
 	existingPrivateClusterTemplate := getResponse.GetObject()
 
-	// Map the public changes to the existing private object (preserving private data):
-	err = s.inMapper.Copy(ctx, publicClusterTemplate, existingPrivateClusterTemplate)
+	// Map the public changes to the existing private object, preserving absent fields:
+	err = s.inMapper.CopyUpdate(ctx, publicClusterTemplate, existingPrivateClusterTemplate)
 	if err != nil {
 		s.logger.ErrorContext(
 			ctx,
--- a/internal/servers/clusters_server.go
+++ b/internal/servers/clusters_server.go
@@ -334,7 +334,7 @@ func (s *ClustersServer) Update(ctx context.Context,
 		}
 		privateCluster = getResponse.GetObject()
 	}
-	err = s.inMapper.Copy(ctx, publicCluster, privateCluster)
+	err = s.inMapper.CopyUpdate(ctx, publicCluster, privateCluster)
 	if err != nil {
 		s.logger.ErrorContext(
 			ctx,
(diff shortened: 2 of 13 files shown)


jhernand (changes requested):
I think we should not do this. The semantics of updates without field masks are exactly those: replace everything, including things that are empty. When that isn't desired the user should use the field mask.

zszabo-rh:
> I think we should not do this. The semantics of updates without field masks are exactly those: replace everything, including things that are empty. When that isn't desired the user should use the field mask.

@wgordon17 what's your take on Juan's proposal?

wgordon17:
@zszabo-rh, Juan is right. I messed up with my implementation for Subnet CIDRs, assuming that the client would do the _wrong_ thing. This is a manufactured bug, and I should have reviewed more thoroughly 😞

zszabo-rh:
update works as expected

PR #425, opened 2026-04-20
MGMT-23147: Rename AVAILABLE to READY condition in ComputeInstance proto

obochan-rh:
- Rename COMPUTE_INSTANCE_CONDITION_TYPE_AVAILABLE to READY (enum value 2 preserved)
- Document Provisioned condition reason values in public proto
- Update generated pb.go files

## Summary by CodeRabbit

* **Documentation**
  * Updated instance condition docs: "READY" now documents reachability via IP, adds documented reason "AsExpected", and clarifies provisioning reason values for TRUE/FALSE scenarios.

* **API Changes**
  * Renamed compute instance condition value from "AVAILABLE" to "READY" (numeric value unchanged) and updated example/description to use "READY".


proto/public/osac/public/v1/compute_instance_type.proto
```diff
@@ -325,6 +328,12 @@ enum ComputeInstanceConditionType {
   // infrastructure resources are allocated.
   //
   // When `status` is `FALSE`, infrastructure provisioning is in progress.
+  //
+  // Possible `reason` values:
+  // - `TenantNotReady`: The parent tenant is not yet in a ready state.
+  // - `WaitingForVM`: VirtualMachine has not been created yet.
+  // - `ProvisioningStorage`: DataVolumes are being created for boot and additional disks.
+  // - `InfrastructureReady`: All infrastructure resources provisioned successfully.
   COMPUTE_INSTANCE_CONDITION_TYPE_PROVISIONED = 5;
```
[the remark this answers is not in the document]
avishayt: I agree - I think the InfrastructureReady reason is confusing.

avishayt (approved):

adriengentil:
Can be worth to update https://github.com/osac-project/enhancement-proposals/tree/main/enhancements/computeinstance-phase-condition-expansion if we want to go to this directory.

Also, as the enhancement says:
```
Note: Available = True indicates VM infrastructure readiness but does not guarantee that the guest OS has fully booted or that applications inside the VM are ready to serve traffic.
```
whereas "ready" may convey the idea that the OS is up an running. I'm not against this change per se but we need to think about it.

proto/public/osac/public/v1/compute_instance_type.proto
```diff
@@ -294,10 +294,19 @@ enum ComputeInstanceConditionType {
   // provisioning or after a spec update).
   COMPUTE_INSTANCE_CONDITION_TYPE_CONFIGURATION_APPLIED = 1;
 
-  // Indicates that the compute instance is available.
+  // Indicates that the compute instance is ready and accessible.
   //
-  // Currently there are no `reason` values defined.
-  COMPUTE_INSTANCE_CONDITION_TYPE_AVAILABLE = 2;
+  // Mirrors the KubeVirt VirtualMachine Ready condition (virt-launcher readiness probe).
```
jhernand: I think we should not refer the user to KubeVirt documetnation, as that is an internal implementation detail. We should be able to replace KubeVirt with some other implementation and keep this documentation unchanged.

proto/public/osac/public/v1/compute_instance_type.proto
```diff
@@ -325,6 +334,14 @@ enum ComputeInstanceConditionType {
...
   //
   // When `status` is `FALSE`, infrastructure provisioning is in progress.
+  //
+  // Possible `reason` values when `status` is `TRUE`:
+  // - `InfrastructureReady`: All infrastructure resources provisioned successfully.
+  //
+  // Possible `reason` values when `status` is `FALSE`:
+  // - `TenantNotReady`: The parent tenant is not yet in a ready state.
+  // - `WaitingForVM`: VirtualMachine has not been created yet.
+  // - `ProvisioningStorage`: DataVolumes are being created for boot and additional disks.
```
jhernand: Same here, the "parent tenant" concept and the fact that we are waiting for a KubeVirt VM are implementation details that should not be presented to the user.

avishayt (approved):

eranco74 (approved):
/approve

PR #427, opened 2026-04-21
NO-ISSUE: use private API types for User and AccessKey protos

akshaynadkarni:
## Summary

User and AccessKey private proto definitions were importing `Metadata` and
`ConditionStatus` from `osac.public.v1` instead of `osac.private.v1`. This
created a transitive dependency on the public API module for any consumer
generating code from the private API (e.g., osac-operator), which is
inconsistent with all other private proto types that correctly use
`osac.private.v1` for these shared types.

## Testing

```
$ go build ./...
# passes

$ go run github.com/onsi/ginkgo/v2/ginkgo run -r --skip-package=it
Ginkgo ran 44 suites in 1m27s
Test Suite Passed

$ gofmt -s -l .
# no output (clean)
```

## Ticket

N/A (no-issue bugfix)

<br>

<small>Assisted-by: Cursor/Claude</small>

## Summary by CodeRabbit

* **Refactor**
  * Updated internal data structures to reference private-scoped type definitions across key system components, improving code isolation and reducing external dependencies while strengthening architectural boundaries.


akshaynadkarni:
cc: @CrystalChun 
I missed this in the review. We shouldn't create a dependency between private and public types.

adriengentil (approved):

CrystalChun:
> cc: @CrystalChun I missed this in the review. We shouldn't create a dependency between private and public types.

Thanks so much for catching this and handling it 🙏

PR #430, opened 2026-04-21
NO_ISSUE: resolve network_class by implementation_strategy

omer-vishlitzky:
## Summary
- Allow `VirtualNetwork.spec.network_class` to be resolved by `implementation_strategy` (e.g., `cudn_net`) in addition to UUID
- The [proto documentation](https://github.com/osac-project/fulfillment-service/blob/main/proto/private/osac/private/v1/virtual_network_type.proto#L68-L69) explicitly states this field "references the `implementation_strategy` field of a NetworkClass resource", but the validation only did a direct ID lookup

## Problem
Creating a VirtualNetwork with `network_class: "cudn_net"` returns:
```
Code: InvalidArgument
Message: network_class 'cudn_net' does not exist
```
even though a NetworkClass with `implementation_strategy: cudn_net` exists and is in READY state.

## Fix
`validateNetworkClassReference()` now tries ID lookup first (fast path, backwards compatible with UUID usage), then falls back to filtering by `implementation_strategy`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation and resolution of network class references to provide clearer errors when a referenced network class is missing or invalid.
  * Enhanced diagnostic messages for readiness and capability mismatches, and more robust lookup behavior to reduce false negatives and aid troubleshooting.

### DIFF
--- a/internal/servers/private_virtual_networks_server.go
+++ b/internal/servers/private_virtual_networks_server.go
@@ -16,6 +16,7 @@ package servers
 import (
 	"context"
 	"errors"
+	"fmt"
 	"log/slog"
 	"net"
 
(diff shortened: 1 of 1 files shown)


akshaynadkarni (approved):
LGTM

PR #431, opened 2026-04-21
MGMT-24090: the created VM doesn't match the provided computeinstance…

achuzhoy:
… name

Creating a computeinstance will result in VM spawned with the same name as the computeinstance.

Co-Authored-By: Claude Opus 4.6

## Summary by CodeRabbit

* **Bug Fixes**
  * ComputeInstance resource creation now uses explicitly provided names when available, falling back to auto-generated names when not specified.

### DIFF
--- a/internal/controllers/computeinstance/computeinstance_reconciler_function.go
+++ b/internal/controllers/computeinstance/computeinstance_reconciler_function.go
@@ -196,7 +196,11 @@ func (t *task) update(ctx context.Context) error {
 		object = &unstructured.Unstructured{}
 		object.SetGroupVersionKind(gvks.ComputeInstance)
 		object.SetNamespace(t.hubNamespace)
-		object.SetGenerateName(objectPrefix)
+		if name := t.computeInstance.GetMetadata().GetName(); name != "" {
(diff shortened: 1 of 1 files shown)


achuzhoy:
Successfully tested this patch on my setup.

internal/controllers/computeinstance/computeinstance_reconciler_function.go
```diff
@@ -196,7 +196,11 @@ func (t *task) update(ctx context.Context) error {
 		object = &unstructured.Unstructured{}
 		object.SetGroupVersionKind(gvks.ComputeInstance)
 		object.SetNamespace(t.hubNamespace)
-		object.SetGenerateName(objectPrefix)
+		if name := t.computeInstance.GetMetadata().GetName(); name != "" {
+			object.SetName(name)
```
adriengentil, line 200: name in fulfillment-service is not unique, it's just a way for users to customize the name that is displayed, and I don't think it should be used internally inside OSAC (maybe that name can also be updated?).

See original issue: https://github.com/osac-project/issues/issues/220

PR #436, opened 2026-04-22
MGMT-24097: Fix exposed BreakGlassCredentials

CrystalChun:
# Description
Previously, the break glass credentials were exposed when an organization is created and were returned in every operation as part of the Organization object.

This ensures that the credentials are only shown during create as intended, and not exposed or returned during other operations.

Includes refactoring the credentials into its own type.

Assisted-by: Claude

# Testing

Tested with 

```
go test -v ./internal/servers
```

## Summary by CodeRabbit

* **New Features**
  * Break glass emergency credentials are now supported for organizations. Administrators can configure emergency access credentials during organization creation, with comprehensive security protections ensuring credentials are transmitted securely, stored safely, and handled with appropriate lifecycle management safeguards.


jhernand:
If we do this how will we avoid storing the credentials in plain text in the database? I think you already considered that, and you plan to implement it accessing Keycloak directly from the gRPC server. The problem with that is that it couples the gRPC server and Keycloak: the gRPC server will have to know the address of Keycloak, the authentication details, etc. This is a general problem, not only affecting this use case, but other use cases where there are secrets involved. I think we need a general solution for that, similar to the secret concept in Kubernetes. If we had that then we could reference a secret, and the interaction with the backend would be hidden behind that secret. Would you consider moving in that direction?

adriengentil:
> If we do this how will we avoid storing the credentials in plain text in the database? 

Can't we just store a hash? as I understand it, we need to communicate this password only once through a secured channel.

jhernand:
I regret my previous comment: it is too ambitious. But I think we can do the following to avoid storing the credentials in the database for a long period of time:

- Create the credentials in the gRPC server when the object is created, and store them in the database. This way the user will see them when she creates the object.

- In the controller take the credentials from the database, use them to configure the IDP and then remove them from the database. The controller will be called inmediatelly (usually) so the credentials will be in the database only for a very short period of time.

Does that look reasonable @CrystalChun ?

If we do that then I think we don't need to discuss the details in this pull request. I would however suggest a few other less important changes:

1. Instead of adding a "break_glass_credentials" field to the create request, add it to the "Organization" type. Otherwise the credentials can't be saved to the database automatically, even if it is for a very short period of time.

2. Try to minimize the fields of the "AccountCredentials" type, to the very minimum. I think the very minimum is the user name and the password, everything else isn't strictly needed.

3. Consider renaming "AccountCredentials" to "BreakGlassCredentials".

CrystalChun:
Thanks for taking a look and for your guidance, Juan!

> Create the credentials in the gRPC server when the object is created, and store them in the database. This way the user will see them when she creates the object.

Sorry, by `object` are you referring to an Organization getting created? 

> Instead of adding a "break_glass_credentials" field to the create request, add it to the "Organization" type. Otherwise the credentials can't be saved to the database automatically, even if it is for a very short period of time.

I've removed it from the create request and added it directly to the type. I think this means we need to update the idp manager too? https://github.com/osac-project/fulfillment-service/blob/a133623c59bc62b507b86fd748f93ce6c49a128a/internal/idp/manager.go#L122

> Try to minimize the fields of the "AccountCredentials" type, to the very minimum. I think the very minimum is the user name and the password, everything else isn't strictly needed.

Good suggestion, removed the excess fields!

> Consider renaming "AccountCredentials" to "BreakGlassCredentials".

Renamed it, thank you!

jhernand (approved):

jhernand:
> Thanks for taking a look and for your guidance, Juan!
> 
> > Create the credentials in the gRPC server when the object is created, and store them in the database. This way the user will see them when she creates the object.
> 
> Sorry, by `object` are you referring to an Organization getting created?

Yes, that is what I mean.

> > Instead of adding a "break_glass_credentials" field to the create request, add it to the "Organization" type. Otherwise the credentials can't be saved to the database automatically, even if it is for a very short period of time.
> 
> I've removed it from the create request and added it directly to the type. I think this means we need to update the idp manager too?

The IDP manger isn't directly exposed to the user of the API, if I understand correctly, so we can keep it outside of this conversation/review.

jhernand (approved):

PR #438, opened 2026-04-23
MGMT-24113: Remove Access Key Type

CrystalChun:
# Description
After evaluation, this type is longer needed. Removes all files associated with access keys.

# Testing

Tested with 

```
go test -v ./internal/servers
```

Assisted-by: Claude

/cc @jhernand 

## Release Notes

* **New Features**
  * Introduced Access Keys API with create, read, update, delete, and list operations
  * Added enable/disable functionality for granular access key lifecycle management
  * Implemented secure credential generation with hashed storage for enhanced security


proto/private/osac/private/v1/access_key_type.proto
```diff
@@ -42,6 +42,12 @@ message AccessKeySpec {
 
   // Whether the access key is enabled.
   bool enabled = 3;
+
+  // The access key ID (public identifier).
+  string access_key_id = 4;
```
jhernand: What is the purpose of this? And more in general what is the purpose of access keys? Will users be allowed to use the API sending these access keys in an HTTP header?

proto/private/osac/private/v1/access_keys_service.proto
```diff
@@ -121,6 +136,14 @@ service AccessKeys {
     };
   }
 
+  rpc Update(AccessKeysUpdateRequest) returns (AccessKeysUpdateResponse) {
+    option (google.api.http) = {
+      patch: "/api/private/v1/organizations/{object.spec.organization_id}/users/{object.spec.user_id}/access-keys/{object.id}"
```
jhernand: I'd very much prefer to avoid nested REST collections. We already have mechanism to restrict the visible objects, and to search, no need to use the URL structure for that. I mean that we should have a top level `/api/private/v1/access_keys` endpoint, and there are mechanisms to search by organization, or user, or any other field of the object. For example, you can use `search=this.organization == "my-org"` without needing a deeply nested URL structure.

proto/private/osac/private/v1/access_key_type.proto
jhernand: Sorry to comment when this has already been merged, but in other parts of the API we don't use `whatever_id` to refer to other objects, we use `whatever` instead. So this should be just `user` if it s referring to a user, `organization` if it is referring to an organization, etc.

jhernand: Where is this actually used? I guess in some part of the .proto that was already merged, maybe as an additional parameter of an operation? I'd suggest to avoid that, and make the credentials part of the `AccessKey` type. That is to avoid as much as possible non standard operations, or non standard parameters in operations, for consistency.

proto/private/osac/private/v1/access_keys_service.proto
```diff
@@ -121,6 +136,14 @@ service AccessKeys {
...
+      body: "object"
+      response_body: "object"
+    };
+  }
+
   rpc Delete(AccessKeysDeleteRequest) returns (AccessKeysDeleteResponse) {
     option (google.api.http) = {delete: "/api/private/v1/organizations/{organization_id}/users/{user_id}/access-keys/{id}"};
```
jhernand: Same, try to avoid nested URL structure. A top level collection is enough. Actually it is better, because it allows search across organizations and users, which is impossible with the nested URL structure.

proto/private/osac/private/v1/access_keys_service.proto
jhernand: The identifier of an access key should be unique, so there is no need to to have the organization or user identifier in this request (also won't be possible to implement that easily with the current database access objects). If  it is necessary to search by organization or by user it can be done using the `search` parameter, for example `filter=this.spec.organization == "my-org" && this.spec.user == "123"`.

jhernand: Better remove the `user_id` and `organization_id` fields. That kind of search (and more) is already possible with the `filter` parameter.

jhernand: There should not be disable or enable operations. Enabling or disabling an access key (if really needed) should be done changing the `spec.enabled` field.

I also wonder if it is really necessary to have this enable/disable mechanism when there is already a mechanism to delete an access key, which I assume will have the same effect than disabling it.

jhernand (approved):

PR #439, opened 2026-04-23
NO-ISSUE: Remove `SystemTenancyLogic` in favor of `DefaultTenancyLogic`

jhernand:
## Summary

- Remove `SystemTenancyLogic` and use `DefaultTenancyLogic` for both public and private APIs,
  eliminating the separate tenancy logic that was previously used for the private API.
- Update `DefaultTenancyLogic.DetermineDefaultTenants` to return the shared tenant when the
  subject has access to all tenants (universal set), since an infinite set cannot be stored as
  the tenants of an object.
- Replace `SystemTenancyLogic` and `SystemAttributionLogic` usages in tests with mocks,
  and add auth subject context where needed.

## Test plan

- [x] Unit tests updated for `DefaultTenancyLogic` to verify shared tenant behavior
- [x] Verify `ginkgo run -r internal` passes
- [x] Verify integration tests pass

## Summary by CodeRabbit

* **Refactor**
  * Consolidated tenancy logic handling and simplified system architecture.
  * Updated default tenant selection behavior for improved consistency.

* **Tests**
  * Enhanced test infrastructure with improved mocking and context injection for authentication scenarios.


CrystalChun (approved):
/lgtm

PR #440, opened 2026-04-23
MGMT-24114: Update Organization type to include status fields

CrystalChun:
# Description

New status fields are added to capture the current state of the Organization when the controller is reconciling it.

# Testing

Tested with 

```
go test -v ./internal/servers
```

Assisted-by: Claude

/cc @jhernand 

## Summary by CodeRabbit

* **New Features**
  * Organizations now include synchronization status tracking with state indicators (pending, synced, failed) and detailed status messages for system health monitoring.


jhernand (approved):

jhernand:
@CrystalChun as a follow up to this, I'd suggest to move everything but `id` and `metadata` (only `description`, I think) to a new `spec` field. That will be more in line with other objects and will give us room for adding more "spec" fields in the future if needed.

CrystalChun:
> @CrystalChun as a follow up to this, I'd suggest to move everything but `id` and `metadata` (only `description`, I think) to a new `spec` field. That will be more in line with other objects and will give us room for adding more "spec" fields in the future if needed.

Good suggestion! Opened https://github.com/osac-project/fulfillment-service/pull/442

PR #443, opened 2026-04-24
MGMT-24097: Move break glass credentials to organization status

jhernand:
## Summary

- Moves the `break_glass_credentials` field from `OrganizationSpec` (public API)
  and the top-level `Organization` message (private API) into `OrganizationStatus`
  in both APIs, since these credentials are server-generated and belong in status.
- Adds `break_glass_credentials` to the private `OrganizationStatus` so the private
  server can populate this information.
- Removes the redundant `optional` modifier from the `BreakGlassCredentials` message
  field (message types in proto3 always have presence tracking).
- Renumbers the private `OrganizationStatus` field tags so that `break_glass_credentials`
  uses tag 1 in both public and private views.

Backwards compatibility is not a concern because Organizations are not yet used in
any production environment.

## Test plan

- [x] `buf lint` passes
- [x] `buf generate` succeeds
- [x] All 49 unit test suites pass (`ginkgo run -r internal`)


CrystalChun (approved):
/lgtm

PR #452, opened 2026-04-27
NO-ISSUE: Make users service consistent with other services

jhernand:
## Summary

- Adds user type definitions (proto) with credentials support for both the public and private
  APIs, including `UserCredentials`, `UserSpec`, `UserStatus`, and `UserCondition` messages.
- Makes the users service proto definitions consistent with other services by adding proper
  request/response types for all RPCs (List, Get, Create, Update, Delete, Signal).
- Adds `pruneCredentials` to both private and public users servers to prevent credential data
  from leaking in API responses. The private server only prunes in `List` (preserving
  credentials in `Get` and `Update` for controller use), while the public server prunes in
  all responses (`Create`, `List`, `Get`, `Update`).
- Adds tests verifying the credential pruning behavior in both servers.

## Test plan

- [x] Unit tests pass: `ginkgo run -r internal` (528 passed)
- [x] Private server tests verify `List` prunes credentials
- [x] Private server tests verify `Get` and `Update` preserve credentials for controllers
- [x] Public server tests verify `Create`, `List`, `Get`, and `Update` all prune credentials


proto/public/osac/public/v1/user_type.proto
```diff
@@ -53,8 +53,25 @@ message UserSpec {
...
+  //
+  // SECURITY: This field contains sensitive data:
+  // - Only transmitted over TLS
+  // - Never returned in responses
+  // - Clear from memory after use
+  optional string password = 1;
+
+  // Whether the password is temporary and must be changed on first login. Only applicable when password is provided.
+  bool temporary_password = 2;
 }
```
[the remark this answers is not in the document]
jhernand: The pruning of credentials from responses is addressed in the second commit in https://github.com/osac-project/fulfillment-service/pull/452 .

internal/servers/users_server.go
```diff
@@ -160,7 +160,10 @@ func (s *UsersServer) List(ctx context.Context,
 		publicItems[i] = publicItem
 	}
 
-	// Build and return the response:
+	// Prune credentials and build the response:
+	for _, item := range publicItems {
+		s.pruneCredentials(item)
+	}
```
CrystalChun, line 166: Wouldn't the credentials have already been pruned in the private server's list function? 
https://github.com/osac-project/fulfillment-service/pull/452/changes/a171f13a2f8794c4acb08fffd58255531e17607f#diff-aad73f4200cca01caa69a0a2b4018ef299b50090684c75a9c15fa45f818e9bb8R113-R115

jhernand: Yes, this is just an additional safety measure: if we changed the private server to not prune the results for the public API would be pruned anyhow.

CrystalChun: Gotcha, nice!

CrystalChun (approved):
/lgtm

PR #454, opened 2026-04-27
MGMT-24062: Wire organization and user servers, events and CLI support

jhernand:
## Summary

- Creates and registers public and private gRPC servers for organizations and users during
  service startup.
- Extends the `Event` protobuf message with `Organization` and `User` payload variants and
  updates `GenericServer` and `EventsServer` to set and extract metadata from those payloads,
  enabling event distribution for these resources.
- Adds table rendering definitions for organizations and users so the CLI `get` command can
  display them, including a lookup column on users that resolves the organization reference.

## Test plan

- [x] Verify `ginkgo run -r internal` passes.
- [x] Verify the gRPC server starts and registers organization and user endpoints.
- [x] Verify `osac get organizations` and `osac get users` render the expected table output.

Related: https://redhat.atlassian.net/browse/MGMT-24062


CrystalChun (approved):
/lgtm

PR #456, opened 2026-04-27
MGMT-23763: Block IP pool deletion when allocated PublicIPs exist

SiddarthR56:
PrivatePublicIPPoolsServer.Delete now checks for allocated PublicIP objects before proceeding. If any PublicIP references the pool via spec.pool, the call is rejected with FailedPrecondition including the pool ID and remaining IP count. 

Added two unit tests cover the happy and rejection paths (both pass), Manully tested the below 4 scenarios:
1.  Create pool, delete immediately (no IPs) → 200 OK, deletionTimestamp set 
2.  Allocate 2 IPs from pool, attempt delete → FailedPrecondition: cannot delete public IP pool '…': 2 public IP(s) are still allocated from it 
3.  Delete both IPs, retry pool delete → 200 OK 
4.  IPs on pool-B don't block pool-A delete; pool-B still blocked 

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevents deletion of IP pools that still have public IPs allocated; deletion now fails with an explicit error if allocations exist. Lookup failures are surfaced as errors to avoid unintended deletions.

* **Tests**
  * Added tests covering deletion rejection when a pool has allocated public IPs and successful deletion when unreferenced.

### DIFF
--- a/internal/servers/private_public_ip_pools_server.go
+++ b/internal/servers/private_public_ip_pools_server.go
@@ -164,7 +164,9 @@ func (s *PrivatePublicIPPoolsServer) Delete(ctx context.Context,
 	return
 }
 
-// checkNoAllocatedIPs returns a FailedPrecondition error when at least one PublicIP still references the pool
+// checkNoAllocatedIPs returns a FailedPrecondition error when at least one PublicIP still
+// references this pool. Any database error is treated as a hard failure so that a transient
+// connectivity issue cannot silently bypass the referential-integrity check and orphan IPs.
 func (s *PrivatePublicIPPoolsServer) checkNoAllocatedIPs(ctx context.Context, poolID string) error {
 	filter := fmt.Sprintf("this.spec.pool == %q", poolID)
 	listResponse, err := s.publicIPDAO.List().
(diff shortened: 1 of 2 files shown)


akshaynadkarni:
@SiddarthR56 Can you please add a PR description and capture what tests you ran with the results?

jhernand:
~~I think this kind of validation should go in the private API server, not the public one. Otherwise admin users will be able to bypass it via the private API, and potentially break the system.~~ It is already in the private server, sorry.

internal/servers/private_public_ip_pools_server.go
```diff
@@ -92,6 +97,17 @@ func (b *PrivatePublicIPPoolsServerBuilder) Build() (result *PrivatePublicIPPool
...
+		SetMetricsRegisterer(b.metricsRegisterer).
+		Build()
+	if err != nil {
+		err = fmt.Errorf("failed to create public IP DAO: %w", err)
+		return
+	}
+
 	generic, err := NewGenericServer[*privatev1.PublicIPPool]().
 		SetLogger(b.logger).
 		SetService(privatev1.PublicIPPools_ServiceDesc.ServiceName).
```
akshaynadkarni, line 113: Is the soft-fail here intentional?

Currently if the DAO query errors (e.g., DB connectivity issue), pool deletion proceeds anyway (`return nil`), which could orphan allocated PublicIPs.

An alternative is hard-fail, returning the error so the delete is blocked until the check can actually run:

```go
if err \!= nil {
    return grpcstatus.Errorf(grpccodes.Internal,
        "failed to verify allocated public IPs for pool '%s': %v", poolID, err)
}
```

For referential integrity checks, hard-fail is usually safer since it prevents data inconsistency at the cost of blocking deletion during transient failures. Curious about the reasoning if soft-fail is intentional here.

akshaynadkarni (commented):
Review comment on error handling approach.

akshaynadkarni (changes requested):
Please check [the comment](https://github.com/osac-project/fulfillment-service/pull/456/changes#r3154573385) and add a PR description containing test output.

akshaynadkarni (approved):
Changes LGTM

PR #457, opened 2026-04-27
MGMT-23761: Add CIDR validation

SiddarthR56:
Added input validation, capacity tracking, and immutability enforcement to the PublicIPPool server.
- Validates CIDR format, prefix length, and IP-family consistency on create
- Rejects pools whose CIDRs overlap any existing pool (AlreadyExists)
- Computes status.total and status.available from the CIDRs at creation time; ignores any caller-supplied status fields
- Enforces spec.cidrs and spec.ip_family as immutable after creation; metadata-only updates are always allowed

Added unit tests to verify the following functionality (All Pass)

1. CIDR format — accepts valid IPv4/IPv6; rejects malformed strings, out-of-range prefixes, and bare IPs; error includes the correct field index
2. IP-family consistency — rejects mismatched CIDRs; accepts multiple same-family CIDRs
3. Cross-pool overlap — rejects exact matches, subsets, and supersets of existing CIDRs; accepts non-overlapping; fails if any one CIDR among many overlaps
4. Capacity on create — correct status.total for /24→254, /28→14, /30→2, /31→2, /32→1, IPv6 /128→1, /64→MaxInt64, and multi-CIDR sums; available==total and allocated==0 on a new pool
5. Immutability — rejects changes to cidrs or ip_family; allows identical spec resend and metadata-only updates

## Summary by CodeRabbit

* **New Features**
  * IP pools now automatically calculate and track total and available capacity based on provided CIDR ranges.
  * Enhanced validation prevents creation of pools with invalid or overlapping CIDR specifications.

* **Bug Fixes**
  * Stricter enforcement of IP family and CIDR immutability rules on pool updates with improved error messaging.

### DIFF
--- a/internal/servers/private_public_ip_pools_server.go
+++ b/internal/servers/private_public_ip_pools_server.go
@@ -201,10 +201,14 @@ func (s *PrivatePublicIPPoolsServer) validateCreate(ctx context.Context,
 		}
 	}
 
+	if err := validateNoCIDRSelfOverlap(cidrs); err != nil {
+		return err
+	}
+
 	return s.validateNoPoolCIDROverlap(ctx, pool)
 }
 
-// validateUpdate validates a PublicIPPool update request. Only immutability of spec fields is checked 
+// validateUpdate validates a PublicIPPool update request. Only immutability of spec fields is checked
 func validateUpdate(newPool *privatev1.PublicIPPool, existing *privatev1.PublicIPPool) error {
 	if newPool == nil {
 		return grpcstatus.Errorf(grpccodes.InvalidArgument, "public IP pool is mandatory")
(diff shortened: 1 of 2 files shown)


akshaynadkarni:
@SiddarthR56 Can you please add a description to this PR? Also capture what tests you ran etc.

internal/servers/private_public_ip_pools_server.go
```diff
@@ -125,16 +130,239 @@ func (s *PrivatePublicIPPoolsServer) Get(ctx context.Context,
...
+		}
+
+		for _, existing := range listResponse.GetItems() {
+			for _, existingCIDRStr := range existing.GetSpec().GetCidrs() {
+				_, existingNet, err := net.ParseCIDR(existingCIDRStr)
+				if err != nil {
+					continue
+				}
+				for _, newCIDR := range newCIDRs {
+					if newCIDR.net.Contains(existingNet.IP) || existingNet.Contains(newCIDR.net.IP) {
```
akshaynadkarni: Can you call the existing [`cidrsOverlap`](https://github.com/osac-project/fulfillment-service/blob/ad13762c5c4f7887e9d07e0892fa88bdf23da953/internal/servers/private_subnets_server.go#L431) helper instead of reimplementing the overlap logic [inline here](https://github.com/osac-project/fulfillment-service/blob/ad13762c5c4f7887e9d07e0892fa88bdf23da953/internal/servers/private_public_ip_pools_server.go#L289)? It's in the same package so you can call it directly. Maybe in the future we can create a common helper function. 
That would also let you remove the [`parsedCIDR` struct and pre-parsing loop](https://github.com/osac-project/fulfillment-service/blob/ad13762c5c4f7887e9d07e0892fa88bdf23da953/internal/servers/private_public_ip_pools_server.go#L262-L269), since `cidrsOverlap` handles parsing internally. This would also fix the [suppressed `net.ParseCIDR` error on L268](https://github.com/osac-project/fulfillment-service/blob/ad13762c5c4f7887e9d07e0892fa88bdf23da953/internal/servers/private_public_ip_pools_server.go#L268): right now if parsing fails, `ipNet` is nil and [`newCIDR.net.Contains()` on L289](https://github.com/osac-project/fulfillment-service/blob/ad13762c5c4f7887e9d07e0892fa88bdf23da953/internal/servers/private_public_ip_pools_server.go#L289) would panic. `cidrsOverlap` [handles parse errors properly](https://github.com/osac-project/fulfillment-service/blob/ad13762c5c4f7887e9d07e0892fa88bdf23da953/internal/servers/private_subnets_server.go#L434-L438).

(*Below :point_down:  is some new functionality you can add in this PR or in a future PR. If you decide to add this functionality in a future PR, then please create a task and assign it to yourself.*)

[`validateNoPoolCIDROverlap`](https://github.com/osac-project/fulfillment-service/blob/ad13762c5c4f7887e9d07e0892fa88bdf23da953/internal/servers/private_public_ip_pools_server.go#L259) checks the new pool's CIDRs against existing pools, but we're not checking whether CIDRs within the same pool overlap each other. For example, a pool created with `cidrs: ["10.0.0.0/24", "10.0.0.0/25"]` would pass validation even though the /25 is a subset of the /24. This would also cause [`calculatePoolCapacity`](https://github.com/osac-project/fulfillment-service/blob/ad13762c5c4f7887e9d07e0892fa88bdf23da953/internal/servers/private_public_ip_pools_server.go#L308) to double-count addresses in the overlapping range. Can you also add a check that CIDRs within the same pool don't overlap each other, [between the format validation loop and the cross-pool check](https://github.com/osac-project/fulfillment-service/blob/ad13762c5c4f7887e9d07e0892fa88bdf23da953/internal/servers/private_public_ip_pools_server.go#L198-L204)?

akshaynadkarni (approved):
LGTM

PR #464, opened 2026-04-28
NO-ISSUE: Enable Keycloak organizations

jhernand:
## Summary

This enables the Keycloak organizations feature and grants the
`osac-controller` service account the permissions it needs to manage
them.

The realm configuration is updated to set `organizationsEnabled` to
`true`, and the `osac-controller` service account is granted the
following roles from the `realm-management` client:

- `manage-realm` - Manage the realm configuration, including organizations.
- `manage-users` - Create, update and delete users.
- `view-realm` - View the realm configuration.
- `view-users` - View users.

The documentation and integration test tooling are updated accordingly.

## Summary by CodeRabbit

* **Documentation**
  * Updated Keycloak setup documentation across multiple deployment guides to reflect expanded service account authorization requirements for proper controller operation.

* **Configuration**
  * Enabled Keycloak organizations functionality in the realm configuration.
  * Updated controller service account role mappings to include additional realm management and viewing permissions.

### DIFF
--- a/it/it_tool.go
+++ b/it/it_tool.go
@@ -958,7 +958,10 @@ func (t *Tool) deployKeycloak(ctx context.Context) error {
 		ClientId:    controllerClientId,
 		ClientRoles: map[string][]string{
 			"realm-management": {
+				"manage-realm",
 				"manage-users",
+				"view-realm",
+				"view-users",
 			},
 		},
 	})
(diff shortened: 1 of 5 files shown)


it/it_tool.go
```diff
@@ -954,6 +956,11 @@ func (t *Tool) deployKeycloak(ctx context.Context) error {
 		Name:        "OSAC controller",
 		Description: "Service account for the OSAC controller",
 		ClientId:    controllerClientId,
+		ClientRoles: map[string][]string{
+			"realm-management": {
+				"manage-realm",
```
CrystalChun, line 961: I think we'd want our admin account to also manage users too, right? Also maybe we should add `view-realm` so the admin can see which organizations (realms) have been created

jhernand: Yes, you are right, I will add those roles.

CrystalChun (approved):
/lgtm

PR #465, opened 2026-04-28
MGMT-24187: Add auth type

CrystalChun:
# Description

The auth type will be used to allow users to log in and obtain a JWT to access the services. There is also a logout functionality that can disable a token.

# Testing

- [x] `buf lint` passes
- [x] `buf generate` succeeds
- [x] All 520 unit test suites pass (`go test -v ./internal/servers`)

/cc @jhernand 

Assisted-by: Claude

## Summary by CodeRabbit

* **New Features**
  * Added authentication service supporting login, token refresh, and logout.
  * Organization-scoped authentication available via both public and private endpoints.
  * Token-based access with access and refresh tokens plus absolute expiration timestamps.
  * Exposed as REST POST endpoints for login, refresh, and logout to simplify integration.


jhernand:
The OAuth specification already defines the [token endpoint](https://datatracker.ietf.org/doc/html/rfc6749#section-3.2) that is used for generating access tokens, almost identical to what is defined here. All OAuth servers, including Keycloak, implement it. So why do we need to reimplement it?

CrystalChun:
> The OAuth specification already defines the [token endpoint](https://datatracker.ietf.org/doc/html/rfc6749#section-3.2) that is used for generating access tokens, almost identical to what is defined here. All OAuth servers, including Keycloak, implement it. So why do we need to reimplement it?

I think the goal is to be able to provide this endpoint through our APIs rather than having them go to a separate endpoint. Specifically, it would allow external tests such as the ISV validation to use our APIs to validate.

Do you think this is unnecessary to meet the requirements from the ISV test suite? If so, I can close it and move on.

jhernand:
> Do you think this is unnecessary to meet the requirements from the ISV test suite? If so, I can close it and move on.

I think it is perfectly fine to tell users that to get tokens they have to go to a different endpoint. We already have a gRPC/REST endpoint that tells the user that:

```
$ curl --silent --insecure https://fulfillment-api.osac.svc.cluster.local:8000/api/fulfillment/v1/capabilities | jq
{
  "authn": {
    "trusted_token_issuers": [
      "https://keycloak.keycloak.svc.cluster.local:8000/realms/osac"
    ]
  }
}
```

This tells the user:

_To get a JSON web token to talk to me go to https://keycloak.keycloak.svc.cluster.local:8000/realms/osac and use the standard OAuth/OIDC protocol to obtain an access token._

That is already used by the OSAC CLI so that the user doesn't have to explicitly provide these details.

If we wanted to improve this we could implement [RFC9728](https://datatracker.ietf.org/doc/html/rfc9728), which would be similar to what you are proposing here, but more standard.

CrystalChun:
/close

Unneeded

PR #466, opened 2026-04-28
MGMT-23734/MGMT-23900: PublicIP pool validation, state machine and capacity tracking

akshaynadkarni:
## Summary

Add business logic validation to the PublicIP server for [MGMT-23734](https://redhat.atlassian.net/browse/MGMT-23734)/[MGMT-23900](https://redhat.atlassian.net/browse/MGMT-23900):

- **Pool validation on Create**: reject nonexistent pool, non-READY pool, exhausted pool (available == 0)
- **State machine enforcement on Update**: only valid transitions accepted (PENDING->ALLOCATED, ALLOCATED->ATTACHED, ATTACHED->RELEASING, RELEASING->ALLOCATED)
- **Atomic capacity tracking**: Create decrements pool.status.available and increments pool.status.allocated in the same DB transaction; Delete reverses
- **Delete constraint**: reject deletion when state is ATTACHED ("detach from ComputeInstance first")
- **Pool immutability on Update**: reject changes to spec.pool after creation

All validation and capacity updates run within the request's existing database transaction.

### Unit Tests

```bash
# Build
go build ./...

# Run all server tests (577 specs)
go run github.com/onsi/ginkgo/v2/ginkgo run internal/servers
# Ran 577 of 577 Specs in 31s
# SUCCESS\! -- 577 Passed | 0 Failed | 0 Pending | 0 Skipped

# Full internal test suite (no regressions)
go run github.com/onsi/ginkgo/v2/ginkgo run -r internal
# SUCCESS\! -- 577 Passed | 0 Failed | 0 Pending | 0 Skipped
```

33 new test cases covering:
- Pool validation: nonexistent pool, non-READY pool, exhausted pool, valid pool (4 tests)
- Capacity tracking: decrement on Create, increment on Delete (2 tests)
- State machine: 4 valid transitions + 3 invalid transitions + UNSPECIFIED no-op (8 tests)
- Delete constraint: reject ATTACHED, allow ALLOCATED, allow PENDING (3 tests)
- Pool immutability: reject change, allow same value (2 tests)
- Existing tests updated to create pools before PublicIP creation (14 tests fixed)

### E2E Tests (edge-22, osac-devel namespace)

Tested against the cluster with all components running the PR branch image
(`mgmt-23734-e2e`). API calls via `curl` to the REST gateway.

| # | Test | Requirement | Result |
|---|------|-------------|--------|
| 1 | Create PublicIP from nonexistent pool | VAL-01 | PASS: `pool 'nonexistent-pool-id-12345' does not exist` (code 3) |
| 2 | Create PublicIP from READY pool + verify capacity | VAL-01, VAL-03 | PASS: PublicIP created, pool allocated 1->2, available 5->4 |
| 3 | Invalid state transition (PENDING -> ATTACHED) | VAL-02 | PASS: `invalid state transition from PUBLIC_IP_STATE_PENDING to PUBLIC_IP_STATE_ATTACHED` (code 9) |
| 4 | Valid state transition (PENDING -> ALLOCATED) | VAL-02 | PASS: state changed to ALLOCATED |
| 5 | Delete PublicIP in ATTACHED state | VAL-04 | PASS: `cannot delete PublicIP: detach from ComputeInstance first` (code 9) |
| 6 | Pool immutability on Update | D-09 | PASS: `field 'spec.pool' is immutable and cannot be changed from '019dc144...' to 'some-other-pool-id'` (code 3) |
| 7 | Delete in non-ATTACHED state + capacity restore | VAL-03, VAL-04 | PASS: deleted, pool allocated 2->1, available 4->5 |

**7/7 E2E tests passed.**

E2E environment: fulfillment-service `mgmt-23734-e2e`, osac-operator `mgmt-23734-e2e`,
OCP 4.20.0, Keycloak OAuth for controller auth, Authorino+OPA for API authorization.

## Related PRs

- [PR #456](https://github.com/osac-project/fulfillment-service/pull/456): Pool-side referential integrity (blocks pool deletion when allocated PublicIPs exist, Siddarth). Complementary, no merge dependency.

## Ticket

- [MGMT-23734](https://redhat.atlassian.net/browse/MGMT-23734): PublicIP Backend (parent story)
- [MGMT-23900](https://redhat.atlassian.net/browse/MGMT-23900): Add pool validation, state machine, capacity tracking, and unit tests (subtask)

<br>

<small>Assisted-by: Cursor/Claude</small>

## Summary by CodeRabbit

* **New Features**
  * Enforces pool existence, READY state, and available capacity on PublicIP create; updates pool capacity counters atomically
  * Validates allowed PublicIP state transitions and enforces immutability of a PublicIP's pool

* **Bug Fixes**
  * Blocks deletion of PublicIPs in ATTACHED or RELEASING states
  * Maps concurrent capacity update conflicts to an abort error

* **Tests**
  * Expanded tests using real pool fixtures covering capacity, transitions, immutability, and error cases

### DIFF
--- a/internal/servers/field_mask.go
+++ b/internal/servers/field_mask.go
@@ -1,0 +1,42 @@
+/*
+Copyright (c) 2026 Red Hat Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with the
+License. You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific
+language governing permissions and limitations under the License.
+*/
+
+package servers
+
+import (
+	"strings"
+
+	"google.golang.org/protobuf/types/known/fieldmaskpb"
+)
+
+// updateIncludesField reports whether a field (or any of its children) is being
+// updated. When mask is nil or empty the caller is doing a full object
+// replacement, so every field is considered updated and the function returns
+// true. This matches GenericServer.Update semantics: a nil mask replaces the
+// entire object, while a non-nil mask applies only the listed fields.
+//
+// Use this to guard field-specific validation in Update hooks so that
+// validation only runs when the field is actually part of the update.
+func updateIncludesField(mask *fieldmaskpb.FieldMask, prefixes ...string) bool {
+	if mask == nil || len(mask.GetPaths()) == 0 {
+		return true
+	}
+	for _, path := range mask.GetPaths() {
+		for _, prefix := range prefixes {
+			if path == prefix || strings.HasPrefix(path, prefix+".") {
+				return true
+			}
+		}
+	}
+	return false
+}
--- a/internal/servers/field_mask.go
+++ b/internal/servers/field_mask.go
@@ -33,7 +33,9 @@ func updateIncludesField(mask *fieldmaskpb.FieldMask, prefixes ...string) bool {
 	}
 	for _, path := range mask.GetPaths() {
 		for _, prefix := range prefixes {
-			if path == prefix || strings.HasPrefix(path, prefix+".") {
+			if path == prefix ||
+				strings.HasPrefix(path, prefix+".") ||
+				strings.HasPrefix(prefix, path+".") {
 				return true
 			}
 		}
--- a/internal/servers/private_public_ips_server.go
+++ b/internal/servers/private_public_ips_server.go
@@ -156,6 +156,9 @@ func (s *PrivatePublicIPsServer) Get(ctx context.Context,
 	return
 }
 
+// Create validates the pool reference, creates the PublicIP, and updates pool capacity counters.
+// All operations share the gRPC interceptor's database transaction: if any step fails, the
+// entire transaction rolls back (pool capacity, PublicIP creation, and validation are atomic).
 func (s *PrivatePublicIPsServer) Create(ctx context.Context,
 	request *privatev1.PublicIPsCreateRequest) (response *privatev1.PublicIPsCreateResponse, err error) {
 	publicIP := request.GetObject()
@@ -179,7 +182,9 @@ func (s *PrivatePublicIPsServer) Create(ctx context.Context,
 		return
 	}
 
-	// Update pool capacity: decrement available, increment allocated:
+	// Update pool capacity: decrement available, increment allocated.
+	// Concurrent allocations are serialized by the pool's resource_version
+	// (optimistic locking). A version conflict returns Aborted for client retry.
 	err = s.updatePoolCapacity(ctx, poolID, 1)
 	if err != nil {
 		return
@@ -188,9 +193,12 @@ func (s *PrivatePublicIPsServer) Create(ctx context.Context,
 	return
 }
 
+// Update validates pool immutability and state machine transitions before persisting.
+// The Get and Update share the gRPC interceptor's transaction. GenericServer.Update uses
+// optimistic locking (resource_version), so a concurrent state change between Get and
+// Update causes a version conflict rather than a silent overwrite.
 func (s *PrivatePublicIPsServer) Update(ctx context.Context,
 	request *privatev1.PublicIPsUpdateRequest) (response *privatev1.PublicIPsUpdateResponse, err error) {
-	// Get existing object for immutability and state machine validation:
 	id := request.GetObject().GetId()
 	if id == "" {
 		err = grpcstatus.Errorf(grpccodes.InvalidArgument, "object identifier is mandatory")
@@ -225,6 +233,8 @@ func (s *PrivatePublicIPsServer) Update(ctx context.Context,
 	return
 }
 
+// Delete rejects deletion of ATTACHED PublicIPs and restores pool capacity on success.
+// All operations share the gRPC interceptor's transaction for atomicity.
 func (s *PrivatePublicIPsServer) Delete(ctx context.Context,
 	request *privatev1.PublicIPsDeleteRequest) (response *privatev1.PublicIPsDeleteResponse, err error) {
 	// Fetch the existing PublicIP to check state and get pool ID:
--- a/internal/servers/private_public_ips_server.go
+++ b/internal/servers/private_public_ips_server.go
@@ -55,9 +55,9 @@ var _ privatev1.PublicIPsServer = (*PrivatePublicIPsServer)(nil)
 type PrivatePublicIPsServer struct {
 	privatev1.UnimplementedPublicIPsServer
 
-	logger           *slog.Logger
-	generic          *GenericServer[*privatev1.PublicIP]
-	publicIPPoolDao  *dao.GenericDAO[*privatev1.PublicIPPool]
+	logger          *slog.Logger
+	generic         *GenericServer[*privatev1.PublicIP]
+	publicIPPoolDao *dao.GenericDAO[*privatev1.PublicIPPool]
 }
 
 // NewPrivatePublicIPsServer creates a builder that can then be used to configure and create a PrivatePublicIPsServer.
--- a/internal/servers/private_public_ips_server.go
+++ b/internal/servers/private_public_ips_server.go
@@ -25,7 +25,17 @@ import (
 	privatev1 "github.com/osac-project/fulfillment-service/internal/api/osac/private/v1"
 	"github.com/osac-project/fulfillment-service/internal/auth"
 	"github.com/osac-project/fulfillment-service/internal/database"
+	"github.com/osac-project/fulfillment-service/internal/database/dao"
 )
+
+// validPublicIPTransitions defines the allowed state transitions for PublicIP resources.
+// The key is the current state and the value is the list of valid target states.
+var validPublicIPTransitions = map[privatev1.PublicIPState][]privatev1.PublicIPState{
+	privatev1.PublicIPState_PUBLIC_IP_STATE_PENDING:   {privatev1.PublicIPState_PUBLIC_IP_STATE_ALLOCATED},
+	privatev1.PublicIPState_PUBLIC_IP_STATE_ALLOCATED: {privatev1.PublicIPState_PUBLIC_IP_STATE_ATTACHED},
+	privatev1.PublicIPState_PUBLIC_IP_STATE_ATTACHED:  {privatev1.PublicIPState_PUBLIC_IP_STATE_RELEASING},
+	privatev1.PublicIPState_PUBLIC_IP_STATE_RELEASING: {privatev1.PublicIPState_PUBLIC_IP_STATE_ALLOCATED},
+}
 
 // PrivatePublicIPsServerBuilder contains the data and logic needed to create a PrivatePublicIPsServer.
 type PrivatePublicIPsServerBuilder struct {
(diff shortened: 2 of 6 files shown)


internal/servers/private_public_ips_server_test.go
```diff
@@ -336,4 +383,482 @@ var _ = Describe("Private public IPs server", func() {
...
+
+			// Update without setting state (UNSPECIFIED is the zero value):
+			object.GetMetadata().Name = "updated-name"
+			object.GetStatus().SetState(privatev1.PublicIPState_PUBLIC_IP_STATE_UNSPECIFIED)
+			resp, err := publicIPsServer.Update(ctx, privatev1.PublicIPsUpdateRequest_builder{
+				Object: object,
+			}.Build())
+			Expect(err).ToNot(HaveOccurred())
+			Expect(resp.GetObject().GetMetadata().GetName()).To(Equal("updated-name"))
+		})
```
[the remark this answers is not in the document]
akshaynadkarni: @coderabbitai The GenericServer's Update does a full object replacement: it stores exactly what the client sends, including zero-value fields. This is consistent across all resources, not specific to PublicIP. If the client sends UNSPECIFIED state, that's what gets persisted. The test verifies that validation is skipped for UNSPECIFIED, which is the intended behavior. State preservation on UNSPECIFIED would be a new requirement and a behavior change, not a test gap.

internal/servers/private_public_ips_server.go
```diff
@@ -142,19 +169,113 @@ func (s *PrivatePublicIPsServer) Create(ctx context.Context,
...
+	getRequest.SetId(id)
+	var getResponse *privatev1.PublicIPsGetResponse
+	err = s.generic.Get(ctx, getRequest, &getResponse)
+	if err != nil {
+		return
+	}
+
+	existingPublicIP := getResponse.GetObject()
+
+	// Reject delete when in ATTACHED state:
```
DakCrowder: Just want to confirm, we're ok with deleting when the state is RELEASING?

akshaynadkarni: Good catch! 
I've updated the code to also block deletion in `RELEASING` state. A PublicIP in `RELEASING` is still bound to a
 ComputeInstance while the controller processes the detach; deleting mid-release could leave the ComputeInstance referencing a deleted IP. Delete is now rejected for both `ATTACHED` and `RELEASING`.

internal/servers/private_public_ips_server.go
```diff
@@ -142,19 +169,114 @@ func (s *PrivatePublicIPsServer) Create(ctx context.Context,
...
+	poolID := publicIP.GetSpec().GetPool()
+	err = s.validatePoolReference(ctx, poolID)
+	if err != nil {
+		return
+	}
+
+	// Create the PublicIP:
 	err = s.generic.Create(ctx, request, &response)
+	if err != nil {
+		return
```
eranco74, line 182: **TOCTOU between validation and capacity update**

`validatePoolReference` checks `available > 0`, and then `updatePoolCapacity` later decrements the counter. Between these two calls, a concurrent request could also pass validation against the same pool.

In practice this is safe because optimistic locking on the pool's `resource_version` will cause the second `updatePoolCapacity` to conflict, and the negative-capacity guard below also catches it. But the early check in `validatePoolReference` is then somewhat redundant -- its main value is producing a clearer error message ("no available capacity" vs an opaque conflict/abort error).

Worth adding a brief comment noting that the real serialization point is the capacity update (optimistic locking), and this early check is for user-facing error quality.

internal/servers/private_public_ips_server.go
```diff
@@ -180,3 +302,111 @@ func (s *PrivatePublicIPsServer) validatePublicIP(ctx context.Context,
...
+	// Check pool is in READY state:
+	if pool.GetStatus().GetState() != privatev1.PublicIPPoolState_PUBLIC_IP_POOL_STATE_READY {
+		return grpcstatus.Errorf(grpccodes.FailedPrecondition,
+			"pool '%s' is not in READY state (current state: %s)",
+			poolID, pool.GetStatus().GetState().String())
+	}
+
+	// Check pool has available capacity:
+	if pool.GetStatus().GetAvailable() <= 0 {
+		return grpcstatus.Errorf(grpccodes.FailedPrecondition,
```
eranco74, line 340: **Misleading error message when `newAllocated < 0`**

When `newAllocated < 0` (e.g., releasing from a pool showing 0 allocated), the error "no available capacity" is confusing -- the actual issue is an inconsistent allocation count, not exhausted capacity.

Consider something more descriptive:
```go
"pool '%s' capacity is inconsistent (allocated=%d, available=%d after delta=%d)", poolID, newAllocated, newAvailable, delta
```

internal/servers/private_public_ips_server_test.go
```diff
@@ -336,4 +383,483 @@ var _ = Describe("Private public IPs server", func() {
...
+			// Delete the PublicIP:
+			_, err = publicIPsServer.Delete(ctx, privatev1.PublicIPsDeleteRequest_builder{
+				Id: publicIPID,
+			}.Build())
+			Expect(err).ToNot(HaveOccurred())
+
+			// Verify pool capacity was restored:
+			poolResp, err = publicIPPoolDao.Get().SetId(poolID).Do(ctx)
+			Expect(err).ToNot(HaveOccurred())
+			pool := poolResp.GetObject()
```
eranco74, line 593: **Nit: duplicate helper function**

`createPublicIPInState` (line 498) and `createPublicIPWithState` (line 639) do the exact same thing -- create a PublicIP via the server then set state via DAO. Consider keeping just one and sharing it across the "State machine enforcement" and "Delete constraint" describe blocks.

internal/servers/private_public_ips_server.go
```diff
@@ -142,19 +169,114 @@ func (s *PrivatePublicIPsServer) Create(ctx context.Context,
...
+	var getResponse *privatev1.PublicIPsGetResponse
+	err = s.generic.Get(ctx, getRequest, &getResponse)
+	if err != nil {
+		return
+	}
+
+	existingPublicIP := getResponse.GetObject()
+
+	// Check pool immutability:
+	if err = validateImmutableFieldsPublicIP(request.GetObject(), existingPublicIP); err != nil {
```
eranco74: **No `FieldMask` check for state transition validation**

Other servers in this codebase (e.g., `private_clusters_server.go` with `isFieldInMask()`) check whether a field is actually included in the update mask before validating it. Here, if a client sends an `UpdateMask` that doesn't include `status.state` but the request object happens to carry a state value, the validation would fire.

The `UNSPECIFIED` skip and `newState != existingState` checks mitigate this for most cases, but full `FieldMask` support would be more robust. Not blocking, but worth noting for future hardening.

eranco74 (commented):
Nice work -- well-structured, follows existing codebase conventions, and thorough test coverage. A few suggestions below.

internal/servers/private_public_ips_server.go
akshaynadkarni: Added a comment clarifying that the availability check in validatePoolReference is for user-facing error quality, and the real serialization point is updatePoolCapacity's optimistic locking via resource_version.

internal/servers/private_public_ips_server.go
akshaynadkarni: Split the bounds check into two cases: "no available capacity" for exhausted pools (newAvailable < 0), and a diagnostic message with actual values for allocation count inconsistencies (newAllocated < 0).

internal/servers/private_public_ips_server_test.go
akshaynadkarni: Consolidated into a single createPublicIPInState helper in the outer scope, parameterized on the server instance. Both inner definitions removed.

internal/servers/private_public_ips_server.go
akshaynadkarni: Implemented. Pool immutability and state machine validation now only run when the respective fields are in the UpdateMask. Created a shared `updateIncludesField` helper in `field_mask.go` that any server in the package can reuse. Nil mask returns true (full replacement means all fields are being updated), matching GenericServer.Update semantics.

internal/servers/private_public_ips_server.go
```diff
@@ -25,8 +26,18 @@ import (
...
 )
 
+// validPublicIPTransitions defines the allowed state transitions for PublicIP resources.
+// The key is the current state and the value is the list of valid target states.
+var validPublicIPTransitions = map[privatev1.PublicIPState][]privatev1.PublicIPState{
+	privatev1.PublicIPState_PUBLIC_IP_STATE_PENDING:   {privatev1.PublicIPState_PUBLIC_IP_STATE_ALLOCATED},
+	privatev1.PublicIPState_PUBLIC_IP_STATE_ALLOCATED: {privatev1.PublicIPState_PUBLIC_IP_STATE_ATTACHED},
+	privatev1.PublicIPState_PUBLIC_IP_STATE_ATTACHED:  {privatev1.PublicIPState_PUBLIC_IP_STATE_RELEASING},
+	privatev1.PublicIPState_PUBLIC_IP_STATE_RELEASING: {privatev1.PublicIPState_PUBLIC_IP_STATE_ALLOCATED},
+}
```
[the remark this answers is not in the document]
akshaynadkarni: Addressed via FieldMask-aware validation. The Update method now checks `updateIncludesField(mask, "status.state")` before running state validation. When UpdateMask is present and doesn't include `status.state`, GenericServer preserves the existing state (partial update), so validation is safely skipped. When UpdateMask is nil (full replacement), `updateIncludesField` returns true and validation runs normally. The shared `updateIncludesField` helper lives in `field_mask.go` and is reusable by any server.

internal/servers/field_mask.go
```diff
@@ -0,0 +1,42 @@
...
+// Use this to guard field-specific validation in Update hooks so that
+// validation only runs when the field is actually part of the update.
+func updateIncludesField(mask *fieldmaskpb.FieldMask, prefixes ...string) bool {
+	if mask == nil || len(mask.GetPaths()) == 0 {
+		return true
+	}
+	for _, path := range mask.GetPaths() {
+		for _, prefix := range prefixes {
+			if path == prefix || strings.HasPrefix(path, prefix+".") {
+				return true
```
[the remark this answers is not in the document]
akshaynadkarni: Fixed. `updateIncludesField` now also checks `strings.HasPrefix(prefix, path+".")`, so a mask containing `"spec"` correctly matches `"spec.pool"`.

internal/servers/private_public_ips_server.go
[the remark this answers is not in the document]
akshaynadkarni: Agreed. Removed the `newState != UNSPECIFIED` bypass. UNSPECIFIED is now rejected by `validatePublicIPStateTransition` as an invalid transition target for any source state. Added a test that verifies UNSPECIFIED is rejected on full replacement, and a separate test that verifies partial updates (UpdateMask excluding `status.state`) skip validation correctly.

internal/servers/private_public_ips_server.go
```diff
@@ -180,3 +305,118 @@ func (s *PrivatePublicIPsServer) validatePublicIP(ctx context.Context,
...
+	}
+
+	newPool := newPublicIP.GetSpec().GetPool()
+	existingPool := existingPublicIP.GetSpec().GetPool()
+
+	// Only check if the request explicitly sets a pool value that differs from the existing one:
+	if newPool != "" && newPool != existingPool {
+		return grpcstatus.Errorf(grpccodes.InvalidArgument,
+			"field 'spec.pool' is immutable and cannot be changed from '%s' to '%s'",
+			existingPool, newPool)
```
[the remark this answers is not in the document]
akshaynadkarni: Fixed. Removed the `newPool != ""` guard. With the FieldMask check ensuring we only enter this block when `spec.pool` is part of the update, an empty value is a deliberate clear attempt and should be rejected as an immutability violation.

DakCrowder (approved):

PR #470, opened 2026-04-29
NO-ISSUE: Sort node requests deterministically

tzvatot:
## Summary

- Sort NodeRequest slice by map key before building ClusterOrder spec
- `GetNodeSets()` returns a map — iterating it produces non-deterministic order
- Without sorting, identical cluster configurations generate different patches when multiple node sets exist, causing unnecessary hub writes

Follow-up from CodeRabbit review on #467.

## Test plan

- [x] `ginkgo run internal/controllers/cluster/` — 9/9 pass

Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved consistency in cluster node processing by ensuring deterministic ordering of node requests during reconciliation operations.

[Review Change Stack](https://app.coderabbit.ai/change-stack/osac-project/fulfillment-service/pull/470)

### DIFF
--- a/internal/controllers/cluster/cluster_reconciler_function.go
+++ b/internal/controllers/cluster/cluster_reconciler_function.go
@@ -334,10 +334,16 @@ func (t *task) addExplicitFields(spec *osacv1alpha1.ClusterOrderSpec) {
 }
 
 func (t *task) prepareNodeRequests() []osacv1alpha1.NodeRequest {
-	var nodeRequests []osacv1alpha1.NodeRequest
-	for _, nodeSet := range t.cluster.GetSpec().GetNodeSets() {
-		nodeRequest := t.prepareNodeRequest(nodeSet)
-		nodeRequests = append(nodeRequests, nodeRequest)
+	nodeSets := t.cluster.GetSpec().GetNodeSets()
+	keys := make([]string, 0, len(nodeSets))
+	for key := range nodeSets {
+		keys = append(keys, key)
+	}
+	slices.Sort(keys)
+
+	nodeRequests := make([]osacv1alpha1.NodeRequest, 0, len(keys))
+	for _, key := range keys {
+		nodeRequests = append(nodeRequests, t.prepareNodeRequest(nodeSets[key]))
 	}
 	return nodeRequests
 }


eranco74:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

internal/controllers/cluster/cluster_reconciler_function.go
```diff
@@ -334,10 +334,16 @@ func (t *task) addExplicitFields(spec *osacv1alpha1.ClusterOrderSpec) {
...
+	keys := make([]string, 0, len(nodeSets))
+	for key := range nodeSets {
+		keys = append(keys, key)
+	}
+	slices.Sort(keys)
+
+	nodeRequests := make([]osacv1alpha1.NodeRequest, 0, len(keys))
+	for _, key := range keys {
+		nodeRequests = append(nodeRequests, t.prepareNodeRequest(nodeSets[key]))
 	}
```
adriengentil, line 347: can't we implement the sort interface `ByKey` for node requests so it can be re-used? https://pkg.go.dev/sort#pkg-overview (see example)

tzvatot: The sorting here is on map keys (node set names from `GetNodeSets()`), not on `NodeRequest` fields directly. A `sort.Interface` on `NodeRequest` (e.g. by `ResourceClass`) would look like:                                                                                                                                    
```                                                                                                                                                                                                                                                                                                                       
  type ByResourceClass []osacv1alpha1.NodeRequest
                                                                                                                                                                                                                                                                                                                       
  func (a ByResourceClass) Len() int           { return len(a) }                                                                                                                                                                                                                                                       
  func (a ByResourceClass) Less(i, j int) bool { return a[i].ResourceClass < a[j].ResourceClass }
  func (a ByResourceClass) Swap(i, j int)      { a[i], a[j] = a[j], a[i] }                                                                                                                                                                                                                                             
```                                                                                                                                                                                                                                                                                                                       
  Which would simplify prepareNodeRequests to:                                                                                                                                                                                                                                                                         
```                                                                                                                                                                                                                                                                                                                       
  func (t *task) prepareNodeRequests() []osacv1alpha1.NodeRequest {
        var nodeRequests []osacv1alpha1.NodeRequest
        for _, nodeSet := range t.cluster.GetSpec().GetNodeSets() {                                                                                                                                                                                                                                                    
                nodeRequests = append(nodeRequests, t.prepareNodeRequest(nodeSet))                                                                                                                                                                                                                                     
        }                                                                                                                                                                                                                                                                                                              
        sort.Sort(ByResourceClass(nodeRequests))                                                                                                                                                                                                                                                                       
        return nodeRequests                                                                                                                                                                                                                                                                                            
  }
```
                                                                                                                                                                                                                                                                                                                       
However, `NodeRequest` is defined in `osac-operator/api/v1alpha1`, so we'd define `ByResourceClass` locally in this package. The tradeoff: this assumes node set map keys and `ResourceClass` produce the same ordering - which they do today since `prepareNodeRequest` maps `nodeSet.GetHostType()` → `ResourceClass`. 
But sorting by map key is arguably more correct since that's the actual source of non-determinism.

adriengentil: We can have a  dedicated package with such tooling. Alternatively, we can also sort the list of `NodeRequest` by key instead.

tzvatot: Both options have tradeoffs:

1. Dedicated package with sort tooling: This feels like over-engineering for a single `slices.Sort(keys)` call. We'd be creating a package to house one sort type that's used in one place. If more sort patterns emerge later, sure, but right now it's premature abstraction.

2. Sort the `NodeRequest` slice by key: This is more interesting but has a problem: `NodeRequest` only has `ResourceClass` and `NumberOfNodes`. There's no key field that maps back to the node set name. We'd be sorting by `ResourceClass`, which happens to work today but conflates two different things - the map key (source of non-determinism) and the resource class (a domain field). If two node sets had the same `ResourceClass` but different sizes, the ordering would be ambiguous.

I'd lean toward keeping the current approach (sorting map keys) as the most correct solution since the non-determinism comes from map iteration, not from the `NodeRequest` struct. The current code is 4 lines, clear, and sorts at the source of the problem.

adriengentil: >since the non-determinism comes from map iteration, not from the NodeRequest struct

I don't understand, the key comes from NodeRequest struct

>sorts at the source of the problem

I won't argue more, but I value the fact that we can structure the code is a way it can be re-used.

adriengentil (approved):

internal/controllers/cluster/cluster_reconciler_function.go
tzvatot: > the key comes from NodeRequest struct

In my implementation, the key comes from the node set name. i.e. "master", "worker" etc. Using `NodeRequest` means the iteration will use the host type (from `ResourceClass`). So, for example, "acme_small", "zeta_large", etc. The result will be a different ordering - still valid though. 

I can make a followup PR to create sorting that we can re-use later on. Do you know if any existing examples that may benefit from it? perhaps I can use that as reference, and fix those too.

PR #476, opened 2026-04-30
Add PublicIP and PublicIPAttachment CLI commands

SiddarthR56:
## Description
Added three osac CLI subcommands for managing PublicIP resources using the public tenant-facing API, following the existing Subnet CLI pattern.

- osac create publicip — allocates a public IP from an existing PublicIPPool; --pool is required, --name is optional
- osac describe publicip ID_OR_NAME — displays tabwriter-formatted details (ID, Name, Pool, Compute Instance, Address, State, Message)
- osac delete publicip ID_OR_NAME... — deletes one or more public IPs by ID or name; accepts multiple arguments
- osac create publicipattachment – Attaches an allocated public IP to acompute instance by setting `spec.compute_instance` via a targeted field-mask
  update. Accepts the public IP and compute instance by ID or name.
- osac delete publicipattachment – Detaches a public IP from its compute  instance by clearing `spec.compute_inst

## Automated — unit tests
$ go run github.com/onsi/ginkgo/v2/ginkgo run \
  internal/cmd/cli/describe/publicip \
  internal/cmd/cli/describe \
  internal/cmd/cli/create
[Describe PublicIP Suite]  6/6 specs    SUCCESS!     PASS
[Describe command]        13/13 specs    SUCCESS!     PASS
[Create command]           8/8 specs    SUCCESS!     PASS
Ginkgo ran 3 suites in 2.929s   ✅ Test Suite Passed

### Createed a PublicIP:
$ ./osac create publicip --name test-ip --pool 019ddf22-7b95-700c-80ee-0d302cb8019c
Created public IP 'test-ip' (ID: 019ddf23-4186-75a8-8f5e-8d1a79cabbdc).

### Describe a PublicIP:  
 $ ./osac describe publicip test-ip
ID:                019ddf23-4186-75a8-8f5e-8d1a79cabbdc
Name:              test-ip
Pool:              019ddf22-7b95-700c-80ee-0d302cb8019c
Compute Instance:  -
Address:           -
State:             -
Message:           -

### Delete Multiple PublicIPs:
$  ./osac delete publicip test-ip test-ip-2
Deleted public IP 'test-ip'.
Deleted public IP 'test-ip-2'.

### Create a publicIP attachemnt
$ ./osac create publicipattachment --publicip 019dfd74-29c5-7305-b163-5853bbb03e5c  --compute-instance 019dfe1f-45de-7450-858f-5f9292521db4
Attached public IP '019dfd74-29c5-7305-b163-5853bbb03e5c' to compute instance '019dfe1f-45de-7450-858f-5f9292521db

### Delelte a publicIP attachement
$ ./osac delete publicipattachment 019dfd74-29c5-7305-b163-5853bbb03e5c
Detached public IP '019dfd74-29c5-7305-b163-5853bbb03e5c' from its compute instance.

## Release Notes

* **New Features**
  * Added public IP management with `create`, `describe`, and `delete` commands
  * Added ability to attach and detach public IPs from compute instances

* **Improvements**
  * Delete operations now process all items before reporting failures, improving error resilience


internal/cmd/cli/create/publicip/create_publicip_cmd.go
```diff
@@ -0,0 +1,123 @@
...
+	c.logger = logging.LoggerFromContext(ctx)
+	c.console = terminal.ConsoleFromContext(ctx)
+
+	cfg, err := config.Load(ctx)
+	if err != nil {
+		return err
+	}
+	if cfg.Address == "" {
+		return fmt.Errorf("there is no configuration, run the 'login' command")
+	}
```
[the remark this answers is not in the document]
SiddarthR56: config.Load uses named return values and assigns cfg = &Config{} before every successful return (lines 78 and 91 of config.go), so (nil, nil) is impossible. The existing if err != nil guard is sufficient.

internal/cmd/cli/delete/publicip/delete_publicip_cmd.go
```diff
@@ -0,0 +1,132 @@
...
+	c.logger = logging.LoggerFromContext(ctx)
+	c.console = terminal.ConsoleFromContext(ctx)
+
+	cfg, err := config.Load(ctx)
+	if err != nil {
+		return err
+	}
+	if cfg.Address == "" {
+		return fmt.Errorf("there is no configuration, run the 'login' command")
+	}
```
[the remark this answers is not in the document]
SiddarthR56: config.Load uses named return values and assigns cfg = &Config{} before every successful return (lines 78 and 91 of config.go), so (nil, nil) is impossible. The existing if err != nil guard is sufficient.

internal/cmd/cli/describe/publicip/describe_publicip_cmd.go
```diff
@@ -0,0 +1,145 @@
...
+		return fmt.Errorf("failed to create gRPC connection: %w", err)
+	}
+	defer conn.Close()
+
+	client := publicv1.NewPublicIPsClient(conn)
+
+	filter := fmt.Sprintf(`this.id == %[1]q || this.metadata.name == %[1]q`, ref)
+	listResponse, err := client.List(ctx, publicv1.PublicIPsListRequest_builder{
+		Filter: &filter,
+	}.Build())
```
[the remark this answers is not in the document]
SiddarthR56: config.Load uses named return values and assigns cfg = &Config{} before every successful return (lines 78 and 91 of config.go), so (nil, nil) is impossible. The existing if err != nil guard is sufficient.

internal/cmd/cli/delete/delete_cmd.go
```diff
@@ -153,27 +155,34 @@ func (c *runnerContext) run(cmd *cobra.Command, args []string) error {
 		}
 	}
 
-	// Delete each resolved object:
+	// Delete each resolved object. Attempt all deletions and report errors at the end
+	var hadErrors bool
```
SiddarthR56, line 159: Following the kubectl/oc pattern, a failure on one item does
not prevent the remaining items from being attempted. All errors are
reported inline and the command exits with a non-zero code if any deletion
failed.

internal/cmd/cli/create/publicip/create_publicip_cmd.go
```diff
@@ -0,0 +1,110 @@
...
+	flags := result.Flags()
+	flags.StringVarP(
+		&runner.args.name,
+		"name",
+		"n",
+		"",
+		"Name of the public IP.",
+	)
+	flags.StringVar(
```
DakCrowder, line 49: Can we mark this flag as required using cobra here so users will see that demarcation in the -h output?

internal/cmd/cli/delete/delete_cmd.go
```diff
@@ -44,6 +45,7 @@ func Cmd() *cobra.Command {
 		Short: "Delete objects",
 		RunE:  runner.run,
 	}
+	result.AddCommand(publicip.Cmd())
```
DakCrowder: Why is publicip the only resource type registered here as a subcommand?  Based off my cursory looking we use reflection to route delte commands, can we do the same for publicip or is there some specific logic we need that differs?

internal/cmd/cli/create/publicip/create_publicip_cmd.go
```diff
@@ -0,0 +1,110 @@
...
+		&runner.args.pool,
+		"pool",
+		"",
+		"ID of the parent PublicIPPool to allocate the address from.",
+	)
```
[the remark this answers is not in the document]
SiddarthR56: attaching a compute instance at creation time (via --compute-instance flag) is future scope. To attach a public IP to a compute instance

internal/cmd/cli/delete/delete_cmd.go
akshaynadkarni: @SiddarthR56 The `delete/publicip/` package has a full implementation but is never imported or registered in `delete_cmd.go`. It's unreachable dead code. `osac delete publicip` currently falls through to the generic reflection-based handler, which already handles name/ID resolution, continue-on-error, and template-based error messages for all other resources (cluster, subnet, etc.).

I'd suggest removing the `delete/publicip/` directory. Every other resource uses the generic handler for deletion, and it works fine for PublicIP too. The only typed delete subcommand (`publicipattachment`) exists because it's not a real delete, it's an Update RPC that clears a field, which reflection can't handle. PublicIP deletion is a standard Delete RPC, so the generic path covers it.

internal/cmd/cli/create/publicipattachment/create_publicipattachment_cmd.go
```diff
@@ -0,0 +1,122 @@
...
+
+	conn, err := cfg.Connect(ctx, cmd.Flags())
+	if err != nil {
+		return fmt.Errorf("failed to create gRPC connection: %w", err)
+	}
+	defer conn.Close()
+
+	client := publicv1.NewPublicIPsClient(conn)
+
+	getResponse, err := client.Get(ctx, publicv1.PublicIPsGetRequest_builder{
```
akshaynadkarni: Both `create publicipattachment` and `delete publicipattachment` pass the user's input directly to `client.Get()`, which only accepts UUIDs (the `PublicIPsGetRequest` message only has an `id` field). If a user passes a name like `my-ip` instead of a UUID, the call will fail with "not found".

The help examples are misleading because they show name-based usage:
```
osac create publicipattachment --publicip my-ip --compute-instance my-vm
osac delete publicipattachment my-ip
```

Other commands like `describe publicip` and the generic `delete` handler support both names and UUIDs by doing a List+filter resolution step first (e.g., `this.id == "ref" || this.metadata.name == "ref"`), then using the resolved ID for the Get call. Could you add the same resolution step here for consistency?

The same issue applies to `delete publicipattachment` at [line 74](https://github.com/osac-project/fulfillment-service/blob/7d54f37a139a17273c5fd605e851f1513c5d55ca/internal/cmd/cli/delete/publicipattachment/delete_publicipattachment_cmd.go#L74).

internal/cmd/cli/create/publicip/create_publicip_cmd.go
```diff
@@ -0,0 +1,108 @@
...
+	result.MarkFlagRequired("pool") //nolint:errcheck
+	// Note: attaching a compute instance at creation time (via --compute-instance flag) is future
+	// scope. To attach a public IP to a compute instance, use 'osac attach publicip'.
```
akshaynadkarni: The comment references `osac attach publicip`, but the actual command added in this PR is `osac create publicipattachment`.

```suggestion
	// Note: attaching a compute instance at creation time (via --compute-instance flag) is future
	// scope. To attach a public IP to a compute instance, use 'osac create publicipattachment'.
```

internal/cmd/cli/create/create_cmd_test.go
```diff
@@ -55,7 +57,7 @@ var _ = Describe("Create command", func() {
 				subcommandNames = append(subcommandNames, subcmd.Name())
 			}
 
-			Expect(subcommandNames).To(ContainElements("cluster", "computeinstance", "hub", "virtualnetwork", "subnet", "securitygroup"))
+			Expect(subcommandNames).To(ContainElements("cluster", "computeinstance", "hub", "publicip", "virtualnetwork", "subnet", "securitygroup"))
```
akshaynadkarni: `publicipattachment` is registered as a subcommand in `create_cmd.go` but isn't included in this assertion. `ContainElements` is a subset check so the test passes, but adding it would ensure the registration isn't accidentally removed.

```suggestion
			Expect(subcommandNames).To(ContainElements("cluster", "computeinstance", "hub", "publicip", "publicipattachment", "virtualnetwork", "subnet", "securitygroup"))
```

akshaynadkarni (approved):
LGTM

PR #482, opened 2026-05-03
OSAC-474: remove --include-deleted flag, always show deleting objects

eranco74:
## Summary

- Remove the `--include-deleted` flag entirely from `get`, `describe`, and `edit` commands — it was a temporary workaround from before archive tables existed
- Always show all non-archived objects in CLI output (no more filtering by `deletion_timestamp`)
- Always display a `DELETING` column in table output so users can see objects being torn down
- Keep the "Deleted" message in the delete command (consistent with `oc delete` convention)

## Jira

[OSAC-474](https://redhat.atlassian.net/browse/OSAC-474)

## Context

The `--include-deleted` flag and `!has(this.metadata.deletion_timestamp)` filter were introduced as a temporary workaround before archive tables existed (see fulfillment-cli PR #17). Now that fully-deleted objects go to `archived_*` tables, the workaround is obsolete — if an object is in the main table, it's still active and should be visible.

This caused user confusion: `osac delete subnet` said "Deleted" and `osac get subnet` showed nothing, but creating a new subnet with the same CIDR failed because the old one was still being torn down.

The agreed approach (from [Jira discussion](https://redhat.atlassian.net/browse/OSAC-474)):
1. Remove the flag entirely — archive tables handle truly-deleted objects now
2. Keep the "Deleted" message as-is (consistent with `oc delete namespace`)
3. Show deletion state visually in table output (DELETING column)
4. Don't change CIDR overlap validation (it's correct)

## Test plan

- [x] `gofmt -s -w .` — no changes
- [x] `go build ./...` — compiles
- [x] `ginkgo run -r internal` — passes (pre-existing auth test env issue excluded)
- [x] New unit tests for DELETING column: header always present, dash for active objects, timestamp for deleting objects
- [ ] `osac get subnet` shows subnets in DELETING state with DELETING column
- [ ] `osac delete subnet <id>` prints "Deleted subnet '...'"

### DIFF
--- a/internal/cmd/cli/annotate/annotate_cmd.go
+++ b/internal/cmd/cli/annotate/annotate_cmd.go
@@ -129,7 +129,7 @@ func (c *runnerContext) run(cmd *cobra.Command, args []string) error {
 	}
 
 	// Find the object by identifier or name:
-	object, err := c.helper.FindObject(ctx, ref, false, c.console)
+	object, err := c.helper.FindObject(ctx, ref, c.console)
 	if err != nil {
 		return err
 	}
--- a/internal/cmd/cli/console/connect/resolve.go
+++ b/internal/cmd/cli/console/connect/resolve.go
@@ -27,7 +27,7 @@ import (
 // ResolveInstance resolves a name or ID to a compute instance ID.
 func ResolveInstance(ctx context.Context, conn *grpc.ClientConn, key string) (string, error) {
 	client := publicv1.NewComputeInstancesClient(conn)
-	ci, err := lookup.Find(key, "compute instance", lookup.FindOptions{}, func(filter string, limit int32) ([]*publicv1.ComputeInstance, error) {
+	ci, err := lookup.Find(key, "compute instance", func(filter string, limit int32) ([]*publicv1.ComputeInstance, error) {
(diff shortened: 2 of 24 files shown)


tzvatot (commented):
## Review

Clean, well-scoped UX fix that aligns CLI messaging with the async reality of deletion. The rename from "deleted" to "deleting" and the default change to `true` are both correct decisions — users should see objects that are still being torn down. The PR description is excellent.

| Category | Count |
|----------|-------|
| 🔴 Critical | 0 |
| 🟡 Important | 0 |
| 💡 Suggestion | 1 |

---

### 💡 Leftover variable name `notDeletedFilter`

**File:** `internal/cmd/cli/get/get_cmd.go` — around the filter logic

```go
const notDeletedFilter = "!has(this.metadata.deletion_timestamp)"
```

Everything else was renamed from "deleted" to "deleting" but this local constant kept the old name. For consistency: `notDeletingFilter`.

jhernand (changes requested):
As I explained in the [ticket](https://redhat.atlassian.net/browse/MGMT-24095?focusedCommentId=16868392) I think that this flag should be completely removed, and that the message doesn't need to be changed.

jhernand (approved):

jhernand (approved):

PR #486, opened 2026-05-05
MGMT-22992: Add role and role binding reconciler skeletons

jhernand:
## Summary

- Add `role` and `role_binding` payloads to both the private and public `Event` messages so that the reconciler watch/LIST sync mechanisms and public event streaming work correctly.
- Register both types in `GenericServer.setPayload` for NOTIFY and add corresponding `extractMetadata` cases in the public events server.
- Wire skeleton reconcilers in `start_controller_cmd` that add the controller finalizer, set default status, and log the reconciled object. The actual reconciliation logic will be added in later patches.

## Test plan

- [x] `buf lint` passes.
- [x] `go build ./...` succeeds.
- [x] Existing unit tests pass (`ginkgo run -r internal`).
- [x] Verify that role and role binding CRUD operations generate events picked up by the new reconcilers.

Related: https://redhat.atlassian.net/browse/MGMT-22992

## Summary by CodeRabbit

* **New Features**
  * Added automatic reconciliation and synchronization capabilities for Role and RoleBinding resources to maintain consistent state across the system.
  * Extended event notifications to include Role and RoleBinding resource events, enabling real-time visibility into role and role binding lifecycle changes including creation, updates, and deletions.


CrystalChun (approved):
/lgtm

PR #495, opened 2026-05-06
OSAC-116: Add public ip address field to ComputeInstance proto status

DakCrowder:
Modifies ComputeInstance status in two critical ways:
1. Adds a new public_ip_address field
2. Renames the existing ip_address field to internal_ip_address.  This is a breaking change for clients, not actually breaking on the wire

Assisted-By: Claude

## Summary by CodeRabbit

* **New Features**
  * Compute instance views now show separate INTERNAL IP ADDRESS and PUBLIC IP ADDRESS columns (replacing the previous EXTERNAL IP).
  * Instance status now exposes distinct internal and public IP fields for clearer visibility.

* **Tests**
  * End-to-end and scenario test data updated to use separate internal and public IP values.


proto/public/osac/public/v1/compute_instance_type.proto
akshaynadkarni:
```suggestion
  // instance is reachable via the Internal IP address in `status.internal_ip_address`).
```

akshaynadkarni:
```suggestion
  // has passed. The instance is reachable via the Internal IP address in `status.internal_ip_address`.
```

internal/testing/compute_instance_scenario.go
akshaynadkarni, line 1: Are you planning to add `PublicIPAddress` support in a follow-up PR?

DakCrowder: Should be simple enough to add now

akshaynadkarni (approved):
LGTM

PR #497, opened 2026-05-06
MGMT-22992: Create admin client in the master Keycloak realm

jhernand:
## Summary

Creates an `admin` service account client in the Keycloak `master` realm during
integration test setup, so it can be used in the future to configure Keycloak via
the admin API instead of importing the OSAC realm from a JSON file.

The Keycloak Helm chart now accepts configurable `admin.username` and
`admin.password` values. After deploying Keycloak, the integration test tool
connects to the admin API to create the client and assign it the `admin` realm
role.

Secret management in the integration tests is also unified: `IT_CLIENT_SECRET`
has been renamed to `IT_SECRET`, and the single secret is now used everywhere
passwords or credentials are needed. The secret is persisted to a Kubernetes
`Secret` in the `default` namespace so that re-runs against an existing cluster
reuse the same value.

## Test plan

- [x] Run integration tests with `IT_KEEP_KIND=true ginkgo run --label-filter=setup it`
- [x] Verify the `admin` client is created in the Keycloak master realm
- [x] Verify the secret is persisted to `default/random` in the cluster
- [x] Verify re-runs reuse the same secret

Related: https://redhat.atlassian.net/browse/MGMT-22992

### DIFF
--- a/README.md
+++ b/README.md
@@ -316,16 +316,26 @@ To clean up a preserved cluster manually:
 $ kind delete cluster --name fulfillment-service-it
 ```
 
-### Client secret for service accounts
+### Secret for passwords and credentials
 
-The integration tests automatically create two Keycloak service account clients: `osac-admin` and
-`osac-controller`. By default, a random secret is generated for them. If you want to use a known
-secret, for example to log in with the CLI afterwards, you can set the `IT_CLIENT_SECRET`
-environment variable:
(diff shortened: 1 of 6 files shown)


jhernand:
@CrystalChun now that we moved to using organizations instead of realms this is less necessary, but I think it will be useful anyhow.

CrystalChun (approved):
/lgtm

PR #498, opened 2026-05-06
OSAC-478: Add Organization-scoped authorization for Admins

CrystalChun:
# Description

Currently, Keycloak Organizations don't have fine-grained access control, so organizational access will need to be managed through OPA in order to ensure admins only have management over their Organization.

## Key changes
- Extracts the Organization of a request in the interceptor to pass along to OPA.

# Testing

- `go build ./...` passes
- All unit tests pass `ginkgo run -r ./internal`

/cc @jhernand 

## Summary by CodeRabbit

* **New Features**
  * Added organization-scoped authorization so org roles can be enforced.
  * Organization admins and IdP managers can now manage users (create, read, update, delete) within their organization.
  * JWT handling enhanced to read organization claims for authorization.
  * Authentication responses now include organization/tenant information for context-aware access.

### DIFF
--- a/charts/service/templates/grpc-server/authconfig.yaml
+++ b/charts/service/templates/grpc-server/authconfig.yaml
@@ -90,6 +90,18 @@ spec:
             "admins",
           }
 
+          # Tenant admin roles - users with these roles can manage users in their tenant
+          tenant_admin_roles := {
+            "tenant-admin",
+            "tenant-user-manager",
+          }
+
+          # Tenant IdP manager roles - users with these roles can manage IdP config and assign roles
+          tenant_idp_manager_roles := {
+            "tenant-admin",
+            "tenant-idp-manager",
+          }
+
           # Get the gRPC method:
           grpc_method := input.context.request.http.path
 
@@ -109,6 +121,38 @@ spec:
             input.auth.identity.authnMethod == "jwt"
           }
 
+          # Get the subject's tenant(s) from JWT claims or service account namespace
+          # For JWT users, this comes from the "organization" scope which is in defaultClientScopes.
+          # The organization claim is required for tenant admins and IdP managers.
+          # For regular users without organization claim, fall back to groups.
+          default subject_tenants = []
+          subject_tenants = input.auth.identity.organization if {
+            input.auth.identity.authnMethod == "jwt"
+            input.auth.identity.organization
+          }
+          subject_tenants = input.auth.identity.organizations if {
+            input.auth.identity.authnMethod == "jwt"
+            input.auth.identity.organizations
+          }
+          # Fallback to groups for JWT users without organization claim
+          subject_tenants = subject_groups if {
+            input.auth.identity.authnMethod == "jwt"
+            not input.auth.identity.organization
+            not input.auth.identity.organizations
+          }
+          # For service accounts, use groups as tenants
+          subject_tenants = subject_groups if {
+            input.auth.identity.authnMethod == "serviceaccount"
+          }
+
+          # Get the subject's realm roles from JWT
+          default subject_realm_roles = []
+          subject_realm_roles = input.auth.identity.realm_access.roles if {
+            input.auth.identity.authnMethod == "jwt"
+            input.auth.identity.realm_access
+            input.auth.identity.realm_access.roles
+          }
+
           # Check if an account is an admin account:
           default is_admin = false
           is_admin if {
@@ -122,10 +166,38 @@ spec:
             group in admin_groups
           }
 
-          # Check if an account is a client account:
+          # Check if an account is a tenant admin (can manage users AND IdP in their tenant):
+          default is_tenant_admin = false
+          is_tenant_admin if {
+            some role in subject_realm_roles
+            role in tenant_admin_roles
+          }
+
+          # Check if an account is a tenant IdP manager (can ONLY manage IdP, NOT users):
+          default is_tenant_idp_manager = false
+          is_tenant_idp_manager if {
+            some role in subject_realm_roles
+            role in tenant_idp_manager_roles
+          }
+
+          # Check if an account is a regular client (no admin or tenant management roles):
           default is_client = false
           is_client if {
             not is_admin
+            not is_tenant_admin
+            not is_tenant_idp_manager
+          }
+
+          # Check if account has client-level permissions (clients, tenant admins, or IdP managers):
+          default has_client_permissions = false
+          has_client_permissions if {
+            is_client
+          }
+          has_client_permissions if {
+            is_tenant_admin
+          }
+          has_client_permissions if {
+            is_tenant_idp_manager
           }
 
           # Allow metadata, reflection and health to everyone:
@@ -139,9 +211,9 @@ spec:
             startswith(grpc_method, "/grpc.health.")
           }
 
-          # Allow specific methods to clients:
+          # Allow specific methods to clients (and tenant admins/IdP managers who inherit client permissions):
           allow if {
-            is_client
+            has_client_permissions
             grpc_method in {
               "/osac.public.v1.ClusterTemplates/Get",
               "/osac.public.v1.ClusterTemplates/List",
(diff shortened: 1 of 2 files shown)


charts/keycloak/templates/_start.tpl
```diff
@@ -104,5 +104,5 @@ https-certificate-key-file=/opt/keycloak/tls/tls.key
 chmod 600 "${conf_file}"
 
 # Start Keycloak:
-exec /opt/keycloak/bin/kc.sh start --import-realm
+exec /opt/keycloak/bin/kc.sh start --import-realm --features=organization
```
jhernand: Is this really needed? In my environment, without this, I can already use organizations.

internal/auth/auth_subject.go
```diff
@@ -28,23 +28,45 @@ type Subject struct {
 	// Tenants is the set of tenants that the subject belongs to. It may be the universal set if the subject has
 	// access to all tenants.
 	Tenants collections.Set[string]
+
+	// Organizations is the set of organizations that the subject belongs to. It may be the universal set if the
+	// subject has access to all organizations (e.g., platform admins).
+	Organizations collections.Set[string]
```
jhernand: What is the difference between tenant and organization? I'd say there is no difference, so we should use the existing `Tenants` field.

internal/auth/auth_subject.go
```diff
@@ -28,23 +28,45 @@ type Subject struct {
...
+
+	// Roles is the list of realm roles assigned to the subject.
+	Roles []string
+
+	// IsOrgAdmin indicates if the subject has organization admin permissions.
+	IsOrgAdmin bool
+
+	// IsOrgIdpManager indicates if the subject has organization IdP manager permissions.
+	IsOrgIdpManager bool
```
jhernand: Having all these fields here makes me think that we intend to make authorization decisions based on them inside the server. That deviates from the principle of having all authorization decisions (except visibility) outside of the server, in Authorino+Keycloak.

charts/keycloak/templates/_start.tpl
CrystalChun: Right, removed it! Thank you

internal/auth/auth_subject.go
CrystalChun: Yes that's right, thank you. I've removed it and completely redid the PR 😅

internal/auth/auth_subject.go
CrystalChun: Gotcha, thank you! Redid the PR, how does it look?

charts/service/templates/grpc-server/authconfig.yaml
```diff
@@ -90,9 +90,23 @@ spec:
...
+          # Organization IdP manager roles - users with these roles can manage IdP config and assign roles
+          org_idp_manager_roles := {
+            "organization-idp-manager",
+          }
+
           # Get the gRPC method:
           grpc_method := input.context.request.http.path
 
+          # Get the target organization from the request (extracted by the gRPC interceptor)
+          target_organization := input.context.context_extensions.organization
```
jhernand: I think that Keycloak can be configured to add the organization identifier to the token as a claim by adding the "organization" scope to the client. For example, we can add it to the `osac-cli` client like this:

```diff
diff --git a/charts/keycloak/files/realm.json b/charts/keycloak/files/realm.json
index 285aa35..220d0fe 100644
--- a/charts/keycloak/files/realm.json
+++ b/charts/keycloak/files/realm.json
@@ -555,7 +555,7 @@
     "authenticationFlowBindingOverrides" : { },
     "fullScopeAllowed" : true,
     "nodeReRegistrationTimeout" : -1,
-    "defaultClientScopes" : [ "groups", "basic", "username" ],
+    "defaultClientScopes" : [ "groups", "basic", "username", "organization" ],
     "optionalClientScopes" : [ ]
   }, {
     "id" : "e37d4620-4ae2-432e-90d8-2b7f6a662f43",
```

Then, if you create a user and assign it to an organization, "org-a" for example, you will get the following token:

```shell
$ ./osac get token --payload
{
  "auth_time": 1778168023,
  "azp": "osac-cli",
  "exp": 1778168326,
  "iat": 1778168026,
  "iss": "https://keycloak.keycloak.svc.cluster.local:8000/realms/osac",
  "jti": "onrtdg:a330f03c-77a3-cdeb-af0d-791523bab9cc",
  "organization": [
    "org-a"
  ],
  "scope": "openid organization",
  "sid": "40a06f68-83e3-4223-8d86-217c5bb0b9f3",
  "sub": "17a73e41-16e1-418a-b89f-a356b5346f01",
  "typ": "Bearer",
  "username": "bob@org-a.com"
}
```

Note the new `organization` claim. Can we use that instead of requiring the server to extract it from the request?

CrystalChun: Thank you for the feedback! I wasn't sure about this so I asked Claude 😅 

>   Scenario: Malicious Request                    
>                                                  
>   Alice is in org-a. She tries to create a user  
>   in org-b:                                    
>                                                  
>   With extraction (current):                     
>   1. JWT: organization = ["org-a"]
>   2. Extract from request: target_organization = 
>   "org-b"                                        
>   3. OPA checks: Is "org-b" in ["org-a"]? ❌     
>   Denied
>                                                  
>   Without extraction:                          
>   1. JWT: organization = ["org-a"]               
>   2. No extraction                             
>   3. OPA: Alice is org admin? ✓ Method allowed
>   4. Server creates user with organization =     
>   "org-b", tenants = ["org-a"] (from our         
>   override)                                      
>   5. ❌ Mismatch: User says org-b but visible to 
>   org-a!    


Is this actually a valid concern or should we remove the server extraction?

I just added organization to the default client scopes, thank you for the suggestion!

jhernand: I think it isn't a valid concern. When an user creates an object (another user, or any other thing) the server decides what tenant to assign based on the "tenants" list passed by Authorino and on the "metadata.tenants" field of the object itself. If the user includes in "metadata.tenants" something that isn't in the tenants passed by Authorino the request is rejected. That is checked by the `determineAssignedTenants` method of the generic server:

https://github.com/osac-project/fulfillment-service/blob/654c83a354960433e135a768306f8ec5a73bcbdd/internal/servers/generic_server.go#L1152-L1287

Exactly here:

https://github.com/osac-project/fulfillment-service/blob/654c83a354960433e135a768306f8ec5a73bcbdd/internal/servers/generic_server.go#L1214-L1242

I think that is sufficient. I will check that we have a test verifying it, and add it if we don't.

CrystalChun: Gotcha, thank you for the explanation Juan! I've removed this section where we're extracting it.

jhernand: Great! Here is a new test to verify that behavior: https://github.com/osac-project/fulfillment-service/pull/504

charts/service/templates/grpc-server/authconfig.yaml
```diff
@@ -90,6 +90,17 @@ spec:
             "admins",
           }
 
+          # Organization admin roles - users with these roles can manage users in their organization
+          org_admin_roles := {
+            "organization-admin",
```
jhernand: Given that organizations and tenants are the same, and that we already have the "tenant" name in many parts of the code, would it make sense to name this `tenant-admin` and in general `tenant-...` to be consistent with the theme?

charts/service/templates/grpc-server/authconfig.yaml
```diff
@@ -109,6 +120,31 @@ spec:
             input.auth.identity.authnMethod == "jwt"
           }
 
+          # Get the subject's organization(s) from JWT claims or service account namespace
+          # For JWT users, this comes from the "organization" scope which is in defaultClientScopes.
+          # The organization claim is required for org admins and IdP managers.
```
jhernand: While the above comment is technically correct, I'd say that is too much detail. I think it is enough to say that we take the organizations from the `organizations` claim. How that is technically achieved in the Keycloak realm configuration is not that interesting here, and may change in the future leaving this comment out of date.

charts/service/templates/grpc-server/authconfig.yaml
```diff
@@ -109,6 +120,31 @@ spec:
...
+          default subject_organizations = []
```
jhernand: Can this be `subject_tenants`? In general, can we use "tenants" instead of "organizations" here as much as possible?

charts/service/templates/grpc-server/authconfig.yaml
```diff
@@ -122,10 +158,26 @@ spec:
...
+            some role in subject_realm_roles
+            role in org_idp_manager_roles
+          }
+
           # Check if an account is a client account:
           default is_client = false
           is_client if {
             not is_admin
+            not is_org_admin
+            not is_org_idp_manager
```
jhernand: Does this mean that if we make a user an tenant admin it will stop being able to create clusters, VMs, etc?

charts/service/templates/grpc-server/authconfig.yaml
CrystalChun: Yes, updated! Thank you

charts/service/templates/grpc-server/authconfig.yaml
CrystalChun: Gotcha, that makes sense. I've removed these comments except for the first line, is that ok?

charts/service/templates/grpc-server/authconfig.yaml
CrystalChun: Yes, updated! Thank you

charts/service/templates/grpc-server/authconfig.yaml
CrystalChun: Good point, I guess they should inherit the ability to manage all other resources too. I've updated it so they can now. How does it look?

jhernand (approved):

PR #503, opened 2026-05-07
OSAC-251: Add network_attachments field for multi-NIC support

ori-amizur:
Add support for multiple network interfaces on ComputeInstances via the new network_attachments field. This enables VMs to attach to multiple subnets with subnet-specific security group policies.

Changes:
- Add NetworkAttachment message type to proto (subnet + security_groups)
- Add network_attachments field to ComputeInstanceSpec (repeated)
- Deprecate flat subnet and security_groups fields in favor of network_attachments for clearer multi-NIC semantics
- Add EffectiveNetworkAttachments() function to normalize both deprecated (subnet/security_groups) and new (network_attachments) into a unified representation for validation and reconciliation
- Update CLI to support --network-attachment flag
- Update reconciler to resolve subnet namespace annotation and validate NetworkAttachment references
- Update private server to populate network_attachments in CR
- Add comprehensive test coverage for effective network logic

Backward compatibility:
- Existing ComputeInstances using subnet/security_groups continue to work via synthetic NetworkAttachment conversion
- Validation prevents mixing deprecated and new fields on same object
- Proto fields marked deprecated with migration path documented


ori-amizur:
This PR depends on https://github.com/osac-project/osac-operator/pull/230 to be merged, and then dependencies will be updated to use the changes in the computeinstance CRD.

internal/cmd/cli/create/computeinstance/create_compute_instance_cmd_test.go
adriengentil, line 1: Use gingko to be consistent with other tests

proto/public/osac/public/v1/compute_instance_type.proto
```diff
@@ -331,18 +337,10 @@ enum ComputeInstanceConditionType {
 
   // Indicates that the infrastructure resources (compute, storage) have been allocated.
   //
-  // When `status` is `TRUE`, the compute instance has been created and
+  // When `status` is `TRUE`, the KubeVirt VirtualMachine has been created and
```
adriengentil: it's a compute instance from fulfillment-service point of view

adriengentil (commented):
## Review: OSAC-251 multi-NIC network_attachments

### Security

**`private_compute_instances_server.go`**
- `isBeingDeleted` guard skips ALL network validation (`validateNetworkReferences`) before tenant-ownership is confirmed. A crafted request with `deletionTimestamp` set bypasses VAL-01–04. Move the guard inside `validateNetworkReferences` where tenancy is already enforced, or gate behind an auth check first.
- `grpcstatus.Errorf(grpccodes.InvalidArgument, "%s", err.Error())` reflects the error string verbatim to the caller. Benign now, but injection risk if the error ever embeds user-supplied values. Wrap: `"invalid network configuration: %s"`.

**`effective_network.go`**
- `EffectiveNetworkAttachments` doesn't reject a `NetworkAttachment` with an empty `subnet`. The empty-subnet entry passes through; the reconciler then silently drops it with no condition set on the CR. Reject empty subnet here or in the server validator.

**`computeinstance_reconciler_function.go`**
- `getSubnetCR` failure → `continue` → NIC silently dropped, no CR condition set. User gets fewer NICs than requested with no observable error. Set `ConfigurationApplied=False` condition and return an error to trigger requeue.
- `getSecurityGroupCR` not-found → NIC added without the SG ref. VM is programmed with fewer security groups than requested, with no user-visible signal. Security posture gap.

---

### Integration

**Reconciler**
- Legacy path: `subnet` is resolved to a `SubnetRef`, but legacy `security_groups` are silently ignored in the reconciler (they are validated server-side but never resolved to refs). If intentional, add a warning log or condition; otherwise fix.
- `return spec, nil` inserted mid-function (legacy early-exit). Any post-processing logic that previously ran after this block is now dead for legacy instances. Verify nothing was skipped.

**Field mask**
- `hasMaskPrefix(mask, "spec.network_attachments")` — confirm this matches how callers send the mask. The proto JSON name is `networkAttachments` (camelCase). If the mask string isn't normalised, a camelCase mask silently skips validation.

**`go.mod`**
- `osac-operator/api` pinned to a pseudo-version (commit hash, not a tag). A force-push or rebase upstream breaks `go.sum`. Tag the operator API release before this merges.

**CLI**
- `--subnet` / `--security-group` are deprecated but no `flags.MarkDeprecated(...)` call is made. They appear identical to `--network-attachment` in help output and shell completions. Add `cobra MarkDeprecated`.

---

### Refactoring

**`parseNetworkAttachmentFlag` helpers**
- `extractSecurityGroupListSuffix` lowercases a copy, finds the index, then slices the original by that same index. Fragile if Unicode characters appear before the marker (byte offsets may differ). Use `strings.IndexFold` or keep the search entirely on the lowercase copy.
- Three private functions (`extractSecurityGroupListSuffix`, `parseMainSubnetOnly`, `parseNetworkAttachmentFlag`) for one flag format. Consider collapsing to a single-pass `strings.Cut`-based approach. Also: two separate lowercase calls on the same input string.

**Reconciler duplication**
- The `for _, att := range ciSpec.GetNetworkAttachments()` loop duplicates the `getSubnetCR` call and warn-and-continue pattern from the legacy block above. Extract a `resolveSubnetRef(ctx, subnetID) (string, error)` helper.

---

### Tests

**Missing reconciler tests**
- `getSecurityGroupCR`: no test for `hubClient.List` returning an error. The reconciler logs and continues; the combined behaviour is unspecified.
- `buildSpec` with `network_attachments`: no test for "subnet CR exists, SG CR not found" → NIC added without SG ref. Silent partial application is unspecified.

**Missing server validation tests**
- `network_attachments[i]`: subnet not found / not READY
- `network_attachments[i]`: SG not found / SG belongs to wrong VirtualNetwork
- `network_attachments[i]`: `security_groups` set without `subnet`
- `isBeingDeleted=true` skip-validation path

**CLI parser tests**
- `Test_buildSpec_WhenNetworkingFlagsAreSetItShouldPopulateSpec` passes `"tmpl"` directly, bypassing `templateID` resolution. Confirm the `buildSpec` signature change didn't introduce new required non-nil parameters.

---

**Must-fix before merge**
1. `isBeingDeleted` bypasses all network validation before auth check — security hole
2. Silent NIC/SG drop in reconciler with no CR condition and no requeue
3. Empty-subnet `NetworkAttachment` passes through `EffectiveNetworkAttachments` silently
4. `MarkDeprecated` missing on deprecated cobra flags
5. Tag `osac-operator/api` — pseudo-version is fragile

tzvatot:
/lgtm

Pending addressing @adriengentil's comments

tzvatot (approved):
/lgtm

PR #514, opened 2026-05-08
OSAC-478: Fix deleting an Organization

CrystalChun:
# Description 
Due to the move to Keycloak Organizations, users are not cascade deleted when an Organization is deleted, which includes the break-glass account.

When an Organization is deleted, we should also delete the break-glass account from the realm.

# Testing

1. Create an organization using OSAC CLI
2. Observe organization created successfully and break-glass account exists in Keycloak console
3. Delete organization via OSAC CLI
4. Observe organization is deleted in Keycloak console and break-glass account is deleted in realm

Assisted-by: Claude

/cc @jhernand


jhernand (approved):

CrystalChun:
Sorry @jhernand , had to rebase off `main`, may I get your review again?

jhernand (approved):

PR #518, opened 2026-05-11
OSAC-827: Introduce Project proto and API

CrystalChun:
# Description

Projects are created and managed by tenant admins and provide a separate boundary within an organization (tenant) for users to manage resources in.

Projects will map to OpenShift projects with support for nested projects.

# Testing

- [x] `buf lint` passes
- [x] `buf generate` succeeds

Assisted-by: Claude

/cc @jhernand 

## Summary by CodeRabbit

* **New Features**
  * Added a Projects resource with create, retrieve, update, and delete operations exposed via public APIs
  * Support for nested projects (parent references) to model hierarchies
  * Project lifecycle states: pending, active, failed, deleting, delete_failed, with status messages
  * Enhanced listing with pagination, filtering, and ordering
  * Optimistic locking for safer concurrent updates

[Review Change Stack](https://app.coderabbit.ai/change-stack/osac-project/fulfillment-service/pull/518)


proto/private/osac/private/v1/project_type.proto
```diff
@@ -0,0 +1,200 @@
...
+  // - The Organization exists
+  // - The Organization.status.state is SYNCED
+  //
+  // If the parent organization is not found or not in SYNCED state, the project will transition to FAILED state.
+  //
+  // Example: "org-abc123"
+  string organization = 1 [
+    (google.api.field_behavior) = REQUIRED,
+    (google.api.field_behavior) = IMMUTABLE
+  ];
```
jhernand: As an organization is the same that a tenant and a user can't see at all objects that belong to another tenant there is no need for this. There is also a `metadata.tenants` field containing the tenants assigned by the system, although that is mostly irrelevant for regular users because they can't change it.

proto/private/osac/private/v1/project_type.proto
```diff
@@ -0,0 +1,200 @@
...
+  // The system validates during reconciliation that:
+  // - The parent Project exists
+  // - The parent Project.status.state is ACTIVE
+  // - The parent Project.spec.organization matches this project's organization
+  // - No circular dependencies exist in the project hierarchy
+  //
+  // If validation fails, the project will transition to FAILED state with an appropriate error message.
+  //
+  // Example: "project-xyz789"
+  optional string parent_project = 2 [(google.api.field_behavior) = IMMUTABLE];
```
jhernand: If this is really the identifier of a project then `project-xyz789` is a bad example because object identifiers generated by the system are UUIDs. Names like `project-xyz789` can be asigned and changed by the user and they aren't unique.

Also the name should be just `parent`, unless a project can have multiple parents of different types.

proto/private/osac/private/v1/project_type.proto
```diff
@@ -0,0 +1,200 @@
...
+
+  // Optional description of the project.
+  //
+  // Human-readable description explaining the purpose or scope of this project. Maximum length: 1024 characters.
+  //
+  // Example: "Production environment for the customer-facing web application"
+  optional string description = 3;
```
jhernand, line 79: If we are going to have a description then we should also have a `title`, like in other types.

CrystalChun: I didn't realize this, thank you for letting me know! I've added `title`

proto/private/osac/private/v1/project_type.proto
CrystalChun: This makes sense, thank you Juan! I've updated the example and changed the name of the field to `parent`

proto/private/osac/private/v1/project_type.proto
CrystalChun: Gotcha, thank you for letting me know! I've removed the field.

jhernand (approved):

PR #521, opened 2026-05-12
OSAC-711: Handle decommissioned hubs to prevent stuck finalizers

ori-amizur:
When a hub is deleted from the database while resources still reference it, reconcilers now distinguish between permanent hub deletion (NotFound) and transient errors (network failures, timeouts). This prevents resources from being stuck with finalizers indefinitely.

Changes:
- Add ClassifyHubError() to identify decommissioned hubs via gRPC NotFound
- Update hub_cache.Get() to classify errors before returning
- Update delete() methods in all 7 reconcilers to remove finalizers when hub not found:
  * cluster_reconciler_function.go
  * computeinstance_reconciler_function.go
  * publicip_reconciler_function.go
  * publicippool_reconciler_function.go
  * securitygroup_reconciler_function.go
  * subnet_reconciler_function.go
  * virtualnetwork_reconciler_function.go
- Add comprehensive unit tests for error classification

Fixes: [OSAC-711](https://redhat.atlassian.net/browse/OSAC-711)


adriengentil (commented):
## Review: OSAC-711 — Decommissioned Hub Handling

### Bug: Cluster reconciler relies on implicit defer

`cluster/delete()` uses `defer func() { if err == nil { t.removeFinalizer() } }()`. The new block returns `nil` without calling `t.removeFinalizer()` explicitly, relying on the defer to fire — and it does fire correctly. But every other reconciler (`computeinstance`, `publicip`, `subnet`, etc.) calls `t.removeFinalizer()` explicitly before `return nil`. The cluster block is the silent exception, documented only by a comment ("Return nil to trigger the defer"). If someone adds another early-return path later and follows the explicit pattern, the cluster behavior silently diverges.

Recommend calling `t.removeFinalizer()` explicitly in the cluster block for consistency, and removing the comment.

---

### Bug: Cache eviction gap limits effectiveness

`hub_cache.Get()` returns a cached entry if one exists — `ClassifyHubError` is only reached on cache miss (first access via `create()`). If a hub was cached before decommission, subsequent calls return the stale entry. The reconciler then fails at the K8s API call level with a connection error, not with `ErrHubNotFound`. The new finalizer-removal path won't trigger until the service restarts (cache cleared → next `Get()` calls `create()` → NotFound → classified).

Worth documenting this limitation or fixing it (evict the cache entry when hub usage fails, then re-check).

---

### Refactoring: 7 identical 12-line blocks

Every reconciler has the same copy-paste block with one variable changing (the resource ID field name). 84 lines that could be 1 shared helper + 7 call sites. Also eliminates the implicit-vs-explicit `removeFinalizer` inconsistency:

```go
// in controllers package
func RemoveFinalizerOnDecommissionedHub(ctx context.Context, logger *slog.Logger, hubID, resourceKey, resourceVal string, removeFinalizer func()) {
    logger.WarnContext(ctx, "Hub not found (likely decommissioned), removing finalizer to allow archiving",
        slog.String("hub_id", hubID),
        slog.String(resourceKey, resourceVal),
    )
    removeFinalizer()
}
```

---

### Comments: overly verbose

The added inline comments explain things the surrounding code already says:
- `// Hub no longer exists - remove finalizer and allow archiving` — duplicates the log message
- `// For transient errors (network, timeout, etc.), continue retrying` — explains `return`
- `// Return nil to trigger the defer which removes the finalizer` — documents a fragile contract instead of making it explicit

Project's existing style is sparse (e.g., `// Remember to remove the finalizer if there was no error:`). The `ClassifyHubError` godoc is also over-documented — multi-paragraph with Parameters/Returns blocks for a 10-line function.

---

### Tests: unit coverage good, integration missing

`errors_test.go` covers `ClassifyHubError` cases well. Missing:

1. **`hub_cache.Get()` test** — no test verifying a gRPC NotFound from `create()` surfaces as `ErrHubNotFound` from `Get()`.
2. **Reconciler `delete()` tests** — no test verifying that when `hubCache.Get()` returns `ErrHubNotFound`, the finalizer is removed and `nil` is returned. This is the core behavior being added.
3. **`codes.PermissionDenied` / `codes.Unauthenticated`** — not transient (won't resolve with retries) but not `NotFound` either. Currently treated as retryable. Worth a test documenting this is intentional, or a third classification.

---

### Non-delete paths (minor, follow-up)

Only `delete()` methods get the new handling. Resources not marked for deletion will keep reconciling against a decommissioned hub indefinitely. Not a blocker for this PR but worth a follow-up ticket.

---

### Security: clean

Classification correctly scopes to gRPC `NotFound` only. `errors.Is` chain is correct. No privilege escalation or data leakage. Allowing archiving on hub-not-found is sound — the alternative is leaked finalizers blocking GC indefinitely.

---

| Area | Status |
|------|--------|
| Core logic | Correct |
| Cluster defer vs explicit inconsistency | Bug (subtle) |
| Cache eviction gap | Bug (limits effectiveness) |
| 7 copy-paste blocks | Refactor opportunity |
| Inline comments | Too verbose |
| Unit tests (`ClassifyHubError`) | Good |
| Integration tests (cache + reconcilers) | Missing |
| Security | Clean |

ori-amizur:
> ## Review: [OSAC-711](https://redhat.atlassian.net/browse/OSAC-711) — Decommissioned Hub Handling
> ### Bug: Cluster reconciler relies on implicit defer
> `cluster/delete()` uses `defer func() { if err == nil { t.removeFinalizer() } }()`. The new block returns `nil` without calling `t.removeFinalizer()` explicitly, relying on the defer to fire — and it does fire correctly. But every other reconciler (`computeinstance`, `publicip`, `subnet`, etc.) calls `t.removeFinalizer()` explicitly before `return nil`. The cluster block is the silent exception, documented only by a comment ("Return nil to trigger the defer"). If someone adds another early-return path later and follows the explicit pattern, the cluster behavior silently diverges.
> 
> Recommend calling `t.removeFinalizer()` explicitly in the cluster block for consistency, and removing the comment.

This is a strange comment: the title is **Bug** but it says that if a user will do something it will become a bug.  Anyway, changed accordingly

> 
> ### Bug: Cache eviction gap limits effectiveness
> `hub_cache.Get()` returns a cached entry if one exists — `ClassifyHubError` is only reached on cache miss (first access via `create()`). If a hub was cached before decommission, subsequent calls return the stale entry. The reconciler then fails at the K8s API call level with a connection error, not with `ErrHubNotFound`. The new finalizer-removal path won't trigger until the service restarts (cache cleared → next `Get()` calls `create()` → NotFound → classified).
> 
> Worth documenting this limitation or fixing it (evict the cache entry when hub usage fails, then re-check).
> 

Added eviction logic for a hub in the hub cache every 5 minutes, so the hub will not remain forever in the cache.

> ### Refactoring: 7 identical 12-line blocks
> Every reconciler has the same copy-paste block with one variable changing (the resource ID field name). 84 lines that could be 1 shared helper + 7 call sites. Also eliminates the implicit-vs-explicit `removeFinalizer` inconsistency:
> 
> ```go
> // in controllers package
> func RemoveFinalizerOnDecommissionedHub(ctx context.Context, logger *slog.Logger, hubID, resourceKey, resourceVal string, removeFinalizer func()) {
>     logger.WarnContext(ctx, "Hub not found (likely decommissioned), removing finalizer to allow archiving",
>         slog.String("hub_id", hubID),
>         slog.String(resourceKey, resourceVal),
>     )
>     removeFinalizer()
> }
> ```
> 

Refactored

> ### Comments: overly verbose
> The added inline comments explain things the surrounding code already says:
> 
> * `// Hub no longer exists - remove finalizer and allow archiving` — duplicates the log message
> * `// For transient errors (network, timeout, etc.), continue retrying` — explains `return`
> * `// Return nil to trigger the defer which removes the finalizer` — documents a fragile contract instead of making it explicit
> 
> Project's existing style is sparse (e.g., `// Remember to remove the finalizer if there was no error:`). The `ClassifyHubError` godoc is also over-documented — multi-paragraph with Parameters/Returns blocks for a 10-line function.
> 
> ### Tests: unit coverage good, integration missing
> `errors_test.go` covers `ClassifyHubError` cases well. Missing:
> 
> 1. **`hub_cache.Get()` test** — no test verifying a gRPC NotFound from `create()` surfaces as `ErrHubNotFound` from `Get()`.
> 2. **Reconciler `delete()` tests** — no test verifying that when `hubCache.Get()` returns `ErrHubNotFound`, the finalizer is removed and `nil` is returned. This is the core behavior being added.
> 3. **`codes.PermissionDenied` / `codes.Unauthenticated`** — not transient (won't resolve with retries) but not `NotFound` either. Currently treated as retryable. Worth a test documenting this is intentional, or a third classification.
> 

Added tests

> ### Non-delete paths (minor, follow-up)
> Only `delete()` methods get the new handling. Resources not marked for deletion will keep reconciling against a decommissioned hub indefinitely. Not a blocker for this PR but worth a follow-up ticket.
> 
> ### Security: clean
> Classification correctly scopes to gRPC `NotFound` only. `errors.Is` chain is correct. No privilege escalation or data leakage. Allowing archiving on hub-not-found is sound — the alternative is leaked finalizers blocking GC indefinitely.

internal/controllers/errors_test.go
```diff
@@ -0,0 +1,116 @@
...
+			Expect(errors.Is(result, ErrHubNotFound)).To(BeFalse())
+			Expect(result).ToNot(BeNil())
+			Expect(result).To(Equal(grpcErr))
+		})
+	})
+})
+
+var _ = Describe("ErrHubNotFound", func() {
+	It("should have correct error message", func() {
+		Expect(ErrHubNotFound.Error()).To(Equal("hub not found"))
```
tzvatot: This is the const of the error - seems redundant to assert it.

internal/controllers/hub_cache_test.go
```diff
@@ -0,0 +1,230 @@
...
+					Return(nil, status.Error(codes.NotFound, "hub not found")).
+					Times(2)
+
+				cache := &hubCache{
+					logger:      logger,
+					client:      mockClient,
+					scheme:      scheme,
+					entries:     map[string]*cachedHubEntry{},
+					entriesLock: &sync.Mutex{},
+					ttl:         100 * time.Millisecond, // Short TTL for testing
```
tzvatot: It's not used - making it identical to the prev test. I would expect the test to wait ttl+gitter and then test the `cache.Get`.

internal/controllers/hub_cache_test.go
```diff
@@ -0,0 +1,230 @@
...
+				Expect(errors.Is(err, ErrHubNotFound)).To(BeTrue())
+
+				// Second call (before TTL expires) - since errors aren't cached, this will call again
+				_, err = cache.Get(ctx, "test-hub-id")
+				Expect(errors.Is(err, ErrHubNotFound)).To(BeTrue())
+			})
+		})
+
+		Context("concurrent access", func() {
+			It("should be thread-safe", func() {
```
tzvatot: the mutex in `Get()` serializes all 10 calls. The test proves sequential lock acquisition works, not concurrent safety. `go test -race` would be more effective.

internal/controllers/cluster/cluster_reconciler_function_test.go
```diff
@@ -484,3 +485,60 @@ var _ = Describe("update tenant annotation", func() {
...
+				Hub: hubID,
+			}.Build(),
+		}.Build()
+
+		hubCache := controllers.NewMockHubCache(ctrl)
+		hubCache.EXPECT().
+			Get(gomock.Any(), hubID).
+			Return(nil, controllers.ErrHubNotFound)
+
+		t := &task{
```
tzvatot: Other reconciles use `newTaskForDelete` helper. Any specific reason to build a task for cluster? in case the cluster delete task changes, this code may break as it's deviating.

internal/controllers/subnet/subnet_reconciler_function_test.go
```diff
@@ -275,6 +275,24 @@ var _ = Describe("delete", func() {
 		Expect(hasFinalizer(t.subnet)).To(BeTrue())
 	})
 
+	It("should remove finalizer when hub cache returns ErrHubNotFound", func() {
```
tzvatot, line 278: `"should remove finalizer when hub cache returns ErrHubNotFound"` tests - 6 total,  follow the same structure: `mock → expect → create task → assert finalizer → delete → assert nil error → assert no finalizer`. Only the resource type differs. Do you think we can consolidate somehow to make a single test helper that will pass the resource?

tzvatot (changes requested):

internal/controllers/hub_cache_test.go
ori-amizur: It is used - the test tests eviction of already existing entry

internal/controllers/hub_cache_test.go
ori-amizur: Removing test

internal/controllers/cluster/cluster_reconciler_function_test.go
ori-amizur: Done

internal/controllers/hub_cache_test.go
tzvatot: Eviction happens after TTL expired. The test does not seems to wait that long. Am I missing something?

internal/controllers/subnet/subnet_reconciler_function_test.go
ori-amizur: This is relatively short test.  It still differs with the following:
The object is different (subnet/computeinstance/cluster etc).
The task is different (every such object has its own task).
The delete function is different.  Every such object has its own delete function.
The hasFinalizer function is different.
So I think it will complicate things without much gain.

tzvatot: Make sense. Thanks!

PR #526, opened 2026-05-13
NO-ISSUE: add service-account-fulfillment-controller to admin service accounts

omer-vishlitzky:
## Summary

The `fulfillment-controller` Keycloak client authenticates as `service-account-fulfillment-controller` via JWT. It needs admin access but was missing from `admin_service_accounts` in both the Helm chart and kustomize manifest. Every installer overlay had to add it independently, which is error-prone and was flagged during review of PR #99 (osac-installer).

Adding it to the base so the overlays don't need to override `admin_service_accounts` at all.

## Test plan

- [ ] Integration tests pass
- [ ] Prow E2E tests pass

### DIFF
--- a/charts/service/templates/grpc-server/authconfig.yaml
+++ b/charts/service/templates/grpc-server/authconfig.yaml
@@ -83,6 +83,7 @@ spec:
           admin_service_accounts := {
             "service-account-osac-admin",
             "service-account-osac-controller",
+            "service-account-fulfillment-controller",
           }
 
           # Admin groups are groups that are allowed to act as administrators.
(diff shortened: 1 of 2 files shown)


charts/service/templates/grpc-server/authconfig.yaml
```diff
@@ -83,6 +83,7 @@ spec:
           admin_service_accounts := {
             "service-account-osac-admin",
             "service-account-osac-controller",
+            "service-account-fulfillment-controller",
```
jhernand, line 86: If you are going to use "service-account-fulfillment-controller" then remove "service-account-osac-controller" as both are used for the same, and update the integration tests accordingly.

PR #531, opened 2026-05-13
OSAC-827: Add conditions to Project type

CrystalChun:
# Description
Projects will be synced to both the openshift cluster and to Keycloak so having these conditions in the Project status allows a user to easily see if the Project is synced or if it failed and why.

# Testing

- [x] `buf lint` passes
- [x] `buf generate` succeeds

Assisted-by: Claude

/cc @jhernand 

## Summary by CodeRabbit

* **New Features**
  * Projects now support enhanced status monitoring through Kubernetes-style condition reporting. Track hub synchronization and identity management (Keycloak) sync operations in detail, including condition status, specific reason codes, and descriptive messages for each tracked condition, enabling better visibility into project synchronization state and faster issue identification.

[Review Change Stack](https://app.coderabbit.ai/change-stack/osac-project/fulfillment-service/pull/531)


proto/public/osac/public/v1/project_type.proto
```diff
@@ -162,3 +214,57 @@ enum ProjectState {
...
+
+  // Indicates whether infrastructure resources have been successfully created for this project.
+  //
+  // Status meanings:
+  // - True: Infrastructure resources exist and are ready
+  // - False: Infrastructure resources failed to sync or are partially synced
+  //
+  // This condition tracks whether the underlying infrastructure (such as OpenShift Projects/namespaces
+  // across managed clusters) has been successfully created.
+  PROJECT_CONDITION_TYPE_HUB_SYNC = 1;
```
jhernand: This condition doesn't belong in the public API. The existence of "hubs" is an internal implementation detail that shouldn't be visible to regular users.

proto/public/osac/public/v1/project_type.proto
```diff
@@ -162,3 +214,57 @@ enum ProjectState {
...
+
+  // Indicates whether access control resources have been configured for project-level permissions.
+  //
+  // Status meanings:
+  // - True: Access control resources created successfully
+  // - False: Failed to create access control resources
+  //
+  // This condition tracks whether the authorization resources (such as Keycloak Authorization Resources)
+  // have been successfully configured for fine-grained project permissions.
+  PROJECT_CONDITION_TYPE_KEYCLOAK_SYNC = 2;
```
jhernand: Same, the fact that we use Keycloak as the IDP should not be visible to regular users of the API.

proto/public/osac/public/v1/project_type.proto
CrystalChun: Gotcha, thanks for letting me know! I've removed this from the public type

proto/public/osac/public/v1/project_type.proto
CrystalChun: Gotcha, thanks for letting me know! I've removed this from the public type

jhernand (approved):

PR #534, opened 2026-05-14
OSAC-881: Add PR checklist enforcement

omer-vishlitzky:
Adds a PR template with required checkboxes and a GitHub Action to enforce all are checked before merge.

## PR Checklist

- [x] I have added tests for this change, or verified they are not applicable
- [x] I have opened an E2E PR in osac-test-infra, or verified it is not applicable
- [x] I have opened a PR to osac-installer for any required deployment changes (kustomize overlays, image tags, env vars, service endpoints, API methods, Helm charts), or verified none are needed

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,5 @@
+## PR Checklist
+
+- [ ] I have added tests for this change, or verified they are not applicable
+- [ ] I have opened an E2E PR in osac-test-infra, or verified it is not applicable
+- [ ] I have opened a PR to osac-installer for any required deployment changes (kustomize overlays, image tags, env vars, service endpoints, API methods, Helm charts), or verified none are needed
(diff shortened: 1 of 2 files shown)


jhernand:
> [x] I have opened a PR to osac-installer for any required deployment changes (kustomize overlays, image tags, env vars, service endpoints, API methods, Helm charts), or verified none are needed

I think that contributors of this project shouldn't be required to understand and responsible for updating a different project that uses different technology. They should be responsible for backwards compatibility of the contracts (the values of the Helm chart and the gRPC API), and only between releases of the project, not for every pull request, and only when there aren't good reasons to break it.

jhernand (approved):

tzvatot:
In my experience, checklist enforcement works well initially when it's new, but over time developers learn to check all boxes to unblock CI without actually doing the work. Kind of like the "I agree to all terms and conditions" (sure, I read it all, all 1000~ pages of it).
What's not strictly enforced gets ignored under developer load. 
Worth considering which items can be enforced structurally (automation, CI gates) rather than relying on developer discipline, and removing checkboxes as automation replaces them.
For example, once the auto-bump installer workflow lands, the "I opened an installer PR" checkbox becomes redundant.

PR #535, opened 2026-05-14
Add auto-bump workflow for osac-installer submodule

omer-vishlitzky:
When a PR merges to main, this workflow automatically creates a PR on osac-installer
bumping the osac-fulfillment-service submodule pointer to the new merge commit.

The bump PR is assigned to the original PR author so the responsibility chain is clear:
merging to the component repo is not done until the submodule is bumped in osac-installer.

Requires `OSAC_BOT_PAT` org secret (osac-dev-bot fine-grained PAT with contents + pull-requests write access).

### DIFF
--- a/.github/workflows/bump-osac-installer.yaml
+++ b/.github/workflows/bump-osac-installer.yaml
@@ -1,0 +1,63 @@
+name: Bump submodule in osac-installer
+
+on:
+  pull_request:
+    types: [closed]
+    branches: [main]
+
+jobs:
+  bump:
+    if: github.event.pull_request.merged == true
+    runs-on: ubuntu-latest
+    env:
+      SUBMODULE_PATH: base/osac-fulfillment-service
+      SUBMODULE_NAME: osac-fulfillment-service
+    steps:
+      - name: Checkout osac-installer
+        uses: actions/checkout@v4
+        with:
+          repository: osac-project/osac-installer
+          token: ${{ secrets.OSAC_BOT_PAT }}
+
+      - name: Update submodule pointer
+        id: update
+        run: |
+          git submodule update --init "$SUBMODULE_PATH"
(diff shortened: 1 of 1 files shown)


jhernand (changes requested):
I think that we should not update any other project automatically when things are merged here, as there may be work in progress that temporarily breaks the contracts (the values of the Helm Chart and the gRPC API). Only when a new release of this project is made should other projects be updated. The commitment of the developers of this project should be to preserve backwards compatibility only between releases, and only when there aren't good reasons to break it.

jhernand (approved):

tzvatot (commented):
Great to see this automated.

A few findings:

**1. Reusable workflow candidate** — This exact workflow (64 lines) is duplicated across 3 repos (fulfillment-service#535, osac-operator#241, osac-aap#298), differing only in `SUBMODULE_PATH` and `SUBMODULE_NAME`. A shared reusable workflow with `workflow_call` inputs would eliminate maintaining three copies and prevent drift. Could live in `osac-installer` itself or a shared `.github` repo.

**2. Re-run fragility** — If `git push` succeeds but `gh pr create` fails (e.g., network blip), re-running the workflow will fail with "branch already exists." Consider checking if the branch exists before pushing, or using `--force` on push, to make re-runs idempotent.

**3. `cd ../..` is fragile** — This assumes the submodule path is always exactly two levels deep (`base/osac-fulfillment-service`). If the path structure ever changes, this silently breaks. Using `cd $GITHUB_WORKSPACE` or saving/restoring the original directory would be more robust.

**4. Stale branch accumulation** — After bump PRs merge, `bump/osac-fulfillment-service/pr-<N>` branches remain on osac-installer. Consider enabling "auto-delete head branches" on osac-installer if not already set.

PR #552, opened 2026-05-18
OSAC-851: Sync Projects to OpenShift Hubs

CrystalChun:
# Description
Add functionality to sync an OSAC Project to all OpenShift Hubs and update the status of the OSAC Project based off of the sync status.

# Testing

- [x] go build passes
- [x] All unit tests pass with `ginkgo run -r ./internal`

Assisted-by: Claude

/cc @jhernand


jhernand (commented):
This looks technically good, but I think it doesn't have any real effect: nothing is going to use the namespaces it creates. Users of the system will not have access to hubs, and therefore they will not be able to do anything with those namespaces. The controllers that we run in those hubs have their own idea of the namespaces that they need, and they create them as needed.

@avishayt what is the plan for these namespaces? How are they going to be used?

avishayt:
> This looks technically good, but I think it doesn't have any real effect: nothing is going to use the namespaces it creates. Users of the system will not have access to hubs, and therefore they will not be able to do anything with those namespaces. The controllers that we run in those hubs have their own idea of the namespaces that they need, and they create them as needed.
> 
> @avishayt what is the plan for these namespaces? How are they going to be used?

Do we want to continue having each controller creating namespaces?  I thought we should have it more organized so that on each hub all resources belonging to a project would be in the same namespace.

CrystalChun:
> > This looks technically good, but I think it doesn't have any real effect: nothing is going to use the namespaces it creates. Users of the system will not have access to hubs, and therefore they will not be able to do anything with those namespaces. The controllers that we run in those hubs have their own idea of the namespaces that they need, and they create them as needed.
> > @avishayt what is the plan for these namespaces? How are they going to be used?
> 
> Do we want to continue having each controller creating namespaces? I thought we should have it more organized so that on each hub all resources belonging to a project would be in the same namespace.

Yeah I think that should be the goal, it will require a separate change to move everything and ensure it functions as expected.

jhernand:
> > > This looks technically good, but I think it doesn't have any real effect: nothing is going to use the namespaces it creates. Users of the system will not have access to hubs, and therefore they will not be able to do anything with those namespaces. The controllers that we run in those hubs have their own idea of the namespaces that they need, and they create them as needed.
> > > @avishayt what is the plan for these namespaces? How are they going to be used?
> > 
> > 
> > Do we want to continue having each controller creating namespaces? I thought we should have it more organized so that on each hub all resources belonging to a project would be in the same namespace.
> 
> Yeah I think that should be the goal, it will require a separate change to move everything and ensure it functions as expected.

I think that is not realistic. Different pieces of software that we use, use namespaces with their own criteria, for example (I may be wrong in some of the details):

- HyperShift creates its own namespace per hosted cluster.
- Assisted service creates agents for an InfraEnv in the same namespace of the InfraEnv.
- KubeVirt uses namespaces for network isolation by tenant, not by project.
- Or very own controllers create a single namespace for all objects.

Forcing all those pieces of software (and the ones to come) to use the namespace structure that you want is probably unfeasible.

In addition generating unique namespace names like `osac-org-<org>-project-<project>` will be most probably impossible, given the limitation of length of namespaces (63 characters). Even more if we want to have nested projects.

I very much advice to abandon that idea.

If we need or want to be able to link Kubernetes namespaces to our tenants and projects I'd suggest to label or annotate them with the identifiers of those tenants and projects.

PR #556, opened 2026-05-19
OSAC-855: sync image tags and rego policies in bump workflow

omer-vishlitzky:
Run sync-image-tags.sh --fix and sync-authconfig-rego.py --fix after updating the submodule pointer, so bump PRs also update image tags and rego policies in osac-installer.


eranco74 (commented):
/lgtm

PR #543, opened 2026-05-15
NO-ISSUE: Replace multi-valued `tenants` and `creators` with singular fields

jhernand:
## Summary

- Replaces `repeated string tenants` with `string tenant` and `repeated string creators` with
  `string creator` in all protobuf `Metadata` messages (public, private, and test APIs).
- Adds two database migrations (39, 40) that convert every table from `text[]` columns with GIN
  indexes to `text` columns with B-tree indexes, preserving the first array element.
- Simplifies the `TenancyLogic` and `AttributionLogic` interfaces: `DetermineDefaultTenants`
  becomes `DetermineDefaultTenant` returning a single string, and `DetermineAssignedCreators`
  becomes `DetermineAssignedCreator` returning a single string. Visibility and assignability
  methods still return sets.
- Updates the generic server, DAO layer, filter translator, all controller reconcilers, CLI table
  layouts, integration tests and documentation accordingly.

## Test plan

- [x] All 54 unit test suites pass (`ginkgo run -r internal`)
- [x] Integration tests pass (`ginkgo run it`)

PR #557, opened 2026-05-19
NO-ISSUE: Add `SharedTenant` and `SystemTenant` named constants

jhernand:
## Summary

Introduces `SharedTenant` and `SystemTenant` variables in the `auth` package so that the
`"shared"` and `"system"` tenant strings are defined once and referenced everywhere, reducing the
risk of typos and making future renames easier. The `SharedTenants` and `SystemTenants` sets now
derive from these variables instead of inline strings, and all test files that previously hardcoded
`"shared"` in metadata tenant fields now use `auth.SharedTenant`.

Addresses the review feedback in
https://github.com/osac-project/fulfillment-service/pull/543#pullrequestreview-4317844885.

## Test plan

- All unit tests pass (`ginkgo run -r internal`).

### DIFF
--- a/internal/auth/tenancy_logic.go
+++ b/internal/auth/tenancy_logic.go
@@ -37,11 +37,17 @@ type TenancyLogic interface {
 	DetermineVisibleTenants(ctx context.Context) (collections.Set[string], error)
 }
 
+// SystemTenant is the tenant that is assigned to objects that are only visible to the system.
+const SystemTenant = "system"
+
 // SystemTenants is the set of tenants that are assigned to objects that are only visible to the system.
-var SystemTenants = collections.NewSet("system")
+var SystemTenants = collections.NewSet(SystemTenant)
+
+// SharedTenant is the tenant that is always visible to all users.
+const SharedTenant = "shared"
 
 // SharedTenants is the set of tenants that are always visible to all users.
-var SharedTenants = collections.NewSet("shared")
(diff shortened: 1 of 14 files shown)


internal/auth/tenancy_logic.go
```diff
@@ -37,11 +37,17 @@ type TenancyLogic interface {
...
+// Tenants that are assigned to objects that are only visible to the system.
+var (
+	SystemTenant  = "system"
+	SystemTenants = collections.NewSet(SystemTenant)
+)
 
-// SharedTenants is the set of tenants that are always visible to all users.
-var SharedTenants = collections.NewSet("shared")
+// Tenants that are always visible to all users.
+var (
```
tzvatot: Coderabbit also commented on this - any reason for using `var` and not `const`?

tzvatot (changes requested):

internal/auth/tenancy_logic.go
```diff
@@ -37,11 +37,17 @@ type TenancyLogic interface {
...
+	SharedTenant  = "shared"
+	SharedTenants = collections.NewSet(SharedTenant)
+)
```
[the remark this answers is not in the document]
jhernand: Done.

internal/auth/tenancy_logic.go
jhernand: No, no reason, fixed.

PR #558, opened 2026-05-19
OSAC-831: Public grpc server for PublicIPAttachment resources

DakCrowder:
[OSAC-831](https://redhat.atlassian.net/browse/OSAC-831)

Implements the public facing grpc server for basic CRUD and List operations for PublicIPAttachments.  Currently no real validation exists - will be a follow up PR

### DIFF
--- a/internal/servers/public_ip_attachments_server.go
+++ b/internal/servers/public_ip_attachments_server.go
@@ -150,7 +150,7 @@ func (s *PublicIPAttachmentsServer) List(ctx context.Context,
 	for i, privateItem := range privateItems {
 		publicItem, err := s.attachmentFromPrivate(ctx, privateItem)
 		if err != nil {
-			return nil, err
+			return nil, grpcstatus.Errorf(grpccodes.Internal, "failed to process public IP attachment")
 		}
 		publicItems[i] = publicItem
 	}
@@ -176,7 +176,7 @@ func (s *PublicIPAttachmentsServer) Get(ctx context.Context,
 	privateObject := privateResponse.GetObject()
 	publicObject, err := s.attachmentFromPrivate(ctx, privateObject)
 	if err != nil {
-		return nil, err
+		return nil, grpcstatus.Errorf(grpccodes.Internal, "failed to process public IP attachment")
 	}
 
 	response := &publicv1.PublicIPAttachmentsGetResponse{}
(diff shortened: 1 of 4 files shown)


SiddarthR56 (approved):

internal/servers/public_ip_attachments_server.go
```diff
@@ -0,0 +1,258 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	privateItems := privateResponse.GetItems()
+	publicItems := make([]*publicv1.PublicIPAttachment, len(privateItems))
+	for i, privateItem := range privateItems {
+		publicItem, err := s.attachmentFromPrivate(ctx, privateItem)
+		if err != nil {
+			return nil, err
```
akshaynadkarni: In `List` and `Get`, mapper errors from `attachmentFromPrivate` are returned as raw errors rather than being wrapped in gRPC status codes. The [`PublicIPsServer` follows a different pattern](https://github.com/osac-project/fulfillment-service/blob/330bf3e22df3e05f1f8fe9b3640c9344fd450fa6/internal/servers/public_ips_server.go#L157-L166) where mapper failures are wrapped in `grpcstatus.Errorf(grpccodes.Internal, ...)` so the client gets a clean `INTERNAL` status instead of `UNKNOWN` with the raw error string.

Is there a reason for not following that pattern here?

internal/servers/public_ip_attachments_server.go
```diff
@@ -0,0 +1,258 @@
...
+func (s *PublicIPAttachmentsServer) Create(ctx context.Context,
+	request *publicv1.PublicIPAttachmentsCreateRequest) (*publicv1.PublicIPAttachmentsCreateResponse, error) {
+	// Map the public IP attachment to private format:
+	publicObject := request.GetObject()
+	if publicObject == nil {
+		return nil, grpcstatus.Errorf(grpccodes.InvalidArgument, "object is mandatory")
+	}
+	privateObject, err := s.attachmentToPrivate(ctx, publicObject)
+	if err != nil {
+		return nil, err
```
akshaynadkarni: The inbound mapping error here is returned raw, but the [outbound mapping error at line 209](https://github.com/osac-project/fulfillment-service/blob/330bf3e22df3e05f1f8fe9b3640c9344fd450fa6/internal/servers/public_ip_attachments_server.go#L209) already wraps it in `grpcstatus.Errorf(grpccodes.Internal, ...)`. This should probably be wrapped the same way for consistency.

internal/servers/public_ip_attachments_server.go
DakCrowder: Ah good catch, intended to add them back (were originally in the helpers) but forgot when refactoring.  Will wrap those errs w/ the status code info

akshaynadkarni (approved):
LGTM

PR #553, opened 2026-05-19
NO-ISSUE: Convert `EffectiveNetworkAttachments` tests to Ginkgo

jhernand:
## Summary

- Convert the `EffectiveNetworkAttachments` tests from standard `testing.T`
  table-driven style to Ginkgo `Describe` / `DescribeTable` with `Entry` blocks.
- Add a Ginkgo suite file for the `computeinstancespec` package.

Using `DescribeTable` makes it possible to focus on a particular test case
with `FEntry`, which helps when debugging, and makes the output of
`ginkgo run` consistent and less noisy across all packages.

## Test plan

- [x] `ginkgo run internal/computeinstancespec` passes (10 specs, 0 failures).

PR #565, opened 2026-05-20
OSAC-251: Remove deprecated subnet and security_groups fields

ori-amizur:
[OSAC-768](https://redhat.atlassian.net/browse/OSAC-768): Make subnet a mandatory field for ComputeInstance creation

Replace deprecated single-subnet/security-groups fields with network_attachments array for multi-NIC VM support.

Changes:
- Reserve proto fields 12 and 13 (subnet, security_groups)
- Add database migration 43 to convert existing data
- Require network_attachments to be non-empty (at least one entry)
- Remove all legacy field handling code and validation
- Refactor computeinstancespec package (rename to validate_network)
- Convert computeinstancespec tests to Ginkgo/Gomega
- Update CLI to reject legacy flags
- Update controller to only use network_attachments

The migration transforms {subnet: "x", security_groups: ["y"]} into {network_attachments: [{subnet: "x", security_groups: ["y"]}]} and erases the old fields from the database.

## Release Notes

* **Refactor**
  * Consolidated compute instance networking configuration to use network attachments exclusively; legacy subnet and security group fields are no longer supported.

* **Documentation**
  * Updated command help text and schema documentation to reflect the network attachment model.

* **Database**
  * Added migration to convert existing instances from legacy networking format to network attachments.


proto/public/osac/public/v1/compute_instance_type.proto
```diff
@@ -152,14 +152,11 @@ message ComputeInstanceSpec {
   // User data for the compute instance (e.g. cloud-init, ignition).
   optional string user_data = 11;
 
-  // Deprecated: use network_attachments. Single subnet for the VM's primary NIC.
-  optional string subnet = 12 [deprecated = true];
+  // Reserved field numbers for removed fields (do not reuse).
+  reserved 12, 13;
+  reserved "subnet", "security_groups";
```
jhernand, line 157: We haven't released any version yet, so I think we don't need to keep this backwards compatible. Remove these fields completely. And renumber the other fields.

internal/computeinstancespec/validate_network_test.go
```diff
@@ -0,0 +1,77 @@
...
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	privatev1 "github.com/osac-project/fulfillment-service/internal/api/osac/private/v1"
+)
+
+func TestComputeInstanceSpec(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "ComputeInstanceSpec Suite")
+}
```
jhernand: This should go in a separate `..._suite_test.go` file, like in other packages. Consider rebasing on top of https://github.com/osac-project/fulfillment-service/pull/553 to get that.

internal/computeinstancespec/validate_network.go
```diff
@@ -0,0 +1,42 @@
...
+
+	privatev1 "github.com/osac-project/fulfillment-service/internal/api/osac/private/v1"
+)
+
+// ValidateNetworkAttachments validates the network_attachments field.
+// It checks that:
+// - The field must be provided (not nil)
+// - The array must not be empty (at least one attachment required)
+// - All attachments have non-empty subnet
+func ValidateNetworkAttachments(networkAttachments []*privatev1.NetworkAttachment) error {
```
jhernand, line 29: Can this be a private method of the server instead of a separate function? Eventually this kind of validation may need to use the logger, or other things, and we will start to add parameters for that. That is avoided if you make it a method.

Consider adding integration tests (see the `it` folder) to test this in a more real environment.

internal/database/migrations/43_migrate_legacy_network_fields_test.go
```diff
@@ -0,0 +1,124 @@
...
+			err = row.Scan(&actualData)
+			Expect(err).ToNot(HaveOccurred())
+
+			// Compare as JSON to ignore formatting differences:
+			var actual, expected map[string]interface{}
+			err = json.Unmarshal([]byte(actualData), &actual)
+			Expect(err).ToNot(HaveOccurred())
+			err = json.Unmarshal([]byte(expectedData), &expected)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(actual).To(Equal(expected))
```
jhernand: Consider using the `MatchJSON` matcher, it is easier and produces better error messages.

internal/database/migrations/43_migrate_legacy_network_fields_test.go
```diff
@@ -0,0 +1,124 @@
...
+		},
+		Entry(
+			"Instance with only subnet",
+			`{"spec": {"subnet": "subnet-123", "template": "template-1"}}`,
+			`{"spec": {"template": "template-1", "network_attachments": [{"subnet": "subnet-123", "security_groups": []}]}}`,
```
jhernand: These one-line JSON documents are difficult to read and change. Consider pretty-printing them.

proto/public/osac/public/v1/compute_instance_type.proto
ori-amizur: This was a clear requirement to make them reserved

internal/computeinstancespec/validate_network_test.go
ori-amizur: Please don't merge https://github.com/osac-project/fulfillment-service/pull/553 since the test file you mentioned will be deleted.  Will add the suite file

internal/computeinstancespec/validate_network.go
ori-amizur: Currently logger is not needed.  Anyway, it is called from 2 separate packages so it is better to leave it as is.

ori-amizur:
Putting on hold until vmaas e2e test is fixed

eranco74:
## Code Review Findings

### 1. Migration number collision (critical)

`49_migrate_legacy_network_fields.up.sql` collides with the existing `49_add_tenant_foreign_key.up.sql` on main. golang-migrate will fail with `duplicate migration version 49` and the service won't start. The migration file and test need to be renumbered (to 50 or whatever the next available number is).

### 2. Migration produces `{"subnet": null}` for security-groups-only rows (low risk, defensive fix recommended)

For rows that have `security_groups` but no `subnet` key, `data->'spec'->>'subnet'` returns SQL NULL, so the migrated network_attachment gets `{"subnet": null}`. After deserialization, `ValidateNetworkAttachments` rejects empty subnet, making those compute instances permanently un-reconcilable.

In practice this is low risk — the existing API has always enforced "subnet is required when security_groups are set" via `validateLegacyFields`, so these rows shouldn't exist. But for defensive correctness, consider adding `AND data->'spec'->>'subnet' IS NOT NULL` to the WHERE clause so any orphaned security-groups-only rows get their legacy fields stripped without creating an invalid network_attachment.

The test case "Instance with only security groups (no subnet)" currently expects `"subnet": null` — this would need updating to match whatever the fix is (skip or strip).

jhernand (approved):

PR #571, opened 2026-05-20
OSAC-204: Add IDP proto and Get/List APIs

CrystalChun:
# Description

Defines the IDP proto to allow admins to get/list IDPs associated with their organizations or at the realm-level through OSAC.

# Testing

- [x] `buf lint` passes
- [x] `buf generate` succeeds

Assisted-by: Claude

/cc @jhernand


proto/private/osac/private/v1/identity_provider_type.proto
```diff
@@ -0,0 +1,105 @@
...
+  IdentityProviderSpec spec = 3;
+
+  // Status is the observed state of the identity provider.
+  IdentityProviderStatus status = 4;
+}
+
+// IdentityProviderSpec represents the desired state of an identity provider.
+message IdentityProviderSpec {
+  // Human-readable display name for this IdP.
+  string display_name = 1;
```
jhernand: To keep this consistent with other types that have similar things this should be `title` and `description`. The first for a short description and the second for a longer description.

proto/private/osac/private/v1/identity_provider_type.proto
```diff
@@ -0,0 +1,105 @@
...
+
+  // Type of identity provider as a free-form string.
+  // Common values: "ldap", "kerberos", "oidc", "saml", "google", "github", "facebook", "microsoft"
+  // The exact set of supported values depends on the underlying IdP system (e.g., Keycloak).
+  string type = 2;
```
jhernand: If we are going to attach a meaning or a validation to this field then it should be an enum, with descriptive names and documented meaning.

proto/private/osac/private/v1/identity_provider_type.proto
```diff
@@ -0,0 +1,105 @@
...
+  bool enabled = 3;
+
+  // Configuration contains provider-specific settings.
+  // Secrets (passwords, client secrets) are automatically filtered and never returned in responses.
+  //
+  // Example configuration keys:
+  // - LDAP: connectionUrl, bindDn, usersDn, authType, vendor
+  // - OIDC: authorizationUrl, tokenUrl, clientId, issuer, defaultScope
+  // - SAML: singleSignOnServiceUrl, singleLogoutServiceUrl, signingCertificate
+  map<string, string> config = 4;
```
jhernand: This should be documented, preferably using specific types for each kind of identity provider. I'd suggest you consider using a `oneof`, for example:

```protobuf
message IdentityProviderSpec {
   ...

  // Provider specific settings.
  oenof config {
    LdapConfig ldap = 4;
    OidcConfig oidc = 5;
    ...
  }
}

// Documentation of the LDAP configuration.
message LdapConfig {
  // Documentation of the connection URL.
  string connection_url = 1;
  
  // Documentation of the bind DN.
  string bind_dn = 2;

  ...
}

// Documentation of the OIDC config.
message OidcConfig {
  ...
}

...
```

I'd also suggest you start by including only one of those `*Config` types, maybe the OIDC one. The others can be added later.

proto/public/osac/public/v1/identity_provider_type.proto
```diff
@@ -0,0 +1,105 @@
...
+// IdentityProviderStatus represents the operational status of an identity provider.
+message IdentityProviderStatus {
+  // Phase represents the current phase of the identity provider.
+  IdentityProviderPhase phase = 1;
+
+  // Message provides additional context about the current status.
+  string message = 2;
+
+  // Conditions represent specific aspects of the identity provider's status.
+  repeated IdentityProviderCondition conditions = 3;
```
jhernand: You will probably want to introduce details about the sync with Keycloak here, and those are details that we don't want to expose to users, so this shouldn't be part of the public API.

proto/private/osac/private/v1/identity_providers_service.proto
```diff
@@ -0,0 +1,132 @@
...
+  int32 total = 2;
+
+  // List of identity providers.
+  repeated IdentityProvider items = 3;
+}
+
+// Request to get an identity provider by alias.
+message IdentityProvidersGetRequest {
+  // Alias of the identity provider.
+  string alias = 1;
```
jhernand: What is the alias? If we need to have an alias, and we need to search by alias, then it should be a field of the type so that the user can use the `filter` parameter to search: `filter: this.spec.alias == "my_alias"`.

proto/private/osac/private/v1/identity_providers_service.proto
```diff
@@ -0,0 +1,132 @@
...
+}
+
+// Response containing the identity provider.
+message IdentityProvidersGetResponse {
+  // The identity provider.
+  IdentityProvider object = 1;
+}
+
+// Request to list organization identity providers.
+message IdentityProvidersListOrganizationRequest {
```
jhernand: When regular tenant user list identity providers they are already given only the ones of their organization/tenant. When a cloud admin lists identity providers they can use a filter: `this.metadata.tenant == "my-org"`. So I think this operation isn't needed.

proto/private/osac/private/v1/identity_providers_service.proto
```diff
@@ -0,0 +1,132 @@
...
+
+// Response containing the organization identity provider.
+message IdentityProvidersGetOrganizationResponse {
+  // The identity provider.
+  IdentityProvider object = 1;
+}
+
+// IdentityProviders service provides internal operations for managing identity providers.
+// Note: IdP assignment to organizations is a link/relationship, not a property on the IdP.
+// Therefore, separate endpoints are needed for realm-level vs organization-level operations.
```
jhernand: This is an important detail of how Keycloak organizes things, but it should be hidden for users.

proto/private/osac/private/v1/identity_providers_service.proto
```diff
@@ -0,0 +1,132 @@
...
+    };
+  }
+
+  // Retrieves an identity provider by alias at the realm level.
+  rpc Get(IdentityProvidersGetRequest) returns (IdentityProvidersGetResponse) {
+    option (google.api.http) = {
+      get: "/api/private/v1/identity-providers/{alias}"
+      response_body: "object"
+    };
+  }
```
jhernand, line 141: You need to add here all the other operations: Create, Update and Delete. Currently the CLI expects all those operations, otherwise none of the generic commands, including `get`, will work. That is a limitation/bug of the CLI that we may address in the future.

proto/private/osac/private/v1/identity_providers_service.proto
CrystalChun: I'm not sure why, but Keycloak uses `alias` as the identity provider's name/id and all of their APIs use `alias` to refer to the identity provider (e.g. https://www.keycloak.org/docs-api/latest/rest-api/index.html#_get_adminrealmsrealmidentity_providerinstancesalias) 

In reality, `alias` is the name of the identity provider. Should this field be called `name` to make it more clear?

proto/public/osac/public/v1/identity_provider_type.proto
CrystalChun: Gotcha, I do have a question about this. Can the tenant admins use the private API?
If not, how can they see the health status of the identity provider? Could we perhaps only allow certain users to reach certain API endpoints?

jhernand: Tenant admins won't have access to the private API at all, and we can't make an exception for that.

If tenant admins need information about the health of their identity provider, then that is a good reason to make the conditions visible in the public API. But make sure that they don't include details that are specific to the implementation.

proto/private/osac/private/v1/identity_providers_service.proto
jhernand: If we need to store the alias, for whatever the reason, it should be `status.alias` (or maybe even `status.keycloak_alias`), and should only be in the private API.

proto/private/osac/private/v1/identity_provider_type.proto
CrystalChun: Will do! Should we only have the types we're defining in the config https://github.com/osac-project/fulfillment-service/pull/571#discussion_r3279591060

E.g. if we're only defining OIDC, then the enum should only contain OIDC because we don't really support the others? Or should we already define all of the possible values listed here in case support for them is added in the future?

jhernand: If you use the "oneof" that I suggested then I think this isn't needed: you will know what time has been selected because only one of the "oneof" fields will be populated. So if you go that direction I think you can just remove this "type" field.

CrystalChun: Great idea! Thanks Juan! I've updated this to use "oneof" how does it look?

proto/private/osac/private/v1/identity_provider_type.proto
CrystalChun: Thank you Juan! I've modified the protos to have this. I've also added LDAP along with OIDC, how does it look?

proto/public/osac/public/v1/identity_provider_type.proto
CrystalChun: Gotcha, ok I've added a minimal status. Thank you for the discussion and recommendations!!

proto/private/osac/private/v1/identity_providers_service.proto
CrystalChun: Gotcha, I've completely removed alias and used name instead seeing as `alias` is really a keycloak construct.

jhernand (approved):

proto/private/osac/private/v1/identity_providers_service.proto
CrystalChun: Removed it, thank you!

proto/private/osac/private/v1/identity_providers_service.proto
CrystalChun: Ah ok thank you for letting me know! Added these as well.

proto/private/osac/private/v1/identity_provider_type.proto
CrystalChun: Added them. Thank you!

PR #595, opened 2026-05-27
OSAC-58: filter unpublished catalog items from public API

tzvatot:
## Summary
- Public `List` endpoints for cluster and compute instance catalog items now inject a `this.published == true` CEL filter, excluding unpublished items from results
- Public `Get` endpoints return `NotFound` for unpublished catalog items
- Private API endpoints are unaffected - admins can still see all items

## Changes
- `catalog_item_validation.go`: add `addPublishedFilter` helper that appends the published CEL clause to user-provided filters
- `cluster_catalog_items_server.go`: apply filter in `List`, add published check in `Get`
- `compute_instance_catalog_items_server.go`: same changes
- New `catalog_item_validation_test.go`: table-driven tests for `addPublishedFilter`
- Updated existing tests to set `Published: true` explicitly
- Added tests: list excludes unpublished, list with user filter excludes unpublished, get returns not found for unpublished

## Test plan
- [x] All 800 server unit tests pass (60 catalog-related)
- [x] `addPublishedFilter` table-driven tests cover empty, simple, and compound filter inputs
- [x] Unpublished items excluded from List (with and without user-provided filter)
- [x] Unpublished items return NotFound from Get
- [x] Published items unaffected in List, Get, Update, Delete
- [ ] E2E: `test_unpublished_catalog_item_not_visible_in_public_api` (blocked on deployment)

## Summary by CodeRabbit

* **New Features**
  * Public List/Get now enforce published-only visibility; non-published items are hidden from other users.
  * Incoming filter expressions are syntax-validated; malformed filters are rejected with an error.

* **Behavior Changes**
  * Retrieving an unpublished item owned by another user returns "not found"; creators can still access their unpublished items.

* **Tests**
  * Added and updated tests covering publication visibility, filter validation, and NotFound behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/osac-project/fulfillment-service/pull/595?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/internal/servers/catalog_item_validation.go
+++ b/internal/servers/catalog_item_validation.go
@@ -15,8 +15,10 @@ package servers
 
 import (
 	"encoding/json"
+	"fmt"
 	"strings"
 
+	"github.com/google/cel-go/cel"
 	"github.com/santhosh-tekuri/jsonschema/v6"
 	grpccodes "google.golang.org/grpc/codes"
 	grpcstatus "google.golang.org/grpc/status"
@@ -27,6 +29,21 @@ import (
 	privatev1 "github.com/osac-project/fulfillment-service/internal/api/osac/private/v1"
 )
 
+// validateCELSyntax checks that a filter string is a syntactically valid, complete CEL expression.
+// This prevents filter bypass attacks where a malicious filter like "true) || (true" could
+// break out of parenthesized composition and change operator precedence.
+func validateCELSyntax(filter string) error {
+	env, err := cel.NewEnv()
+	if err != nil {
+		return fmt.Errorf("failed to create CEL environment: %w", err)
+	}
+	_, issues := env.Parse(filter)
+	if issues != nil && issues.Err() != nil {
+		return fmt.Errorf("syntax error: %w", issues.Err())
+	}
+	return nil
+}
+
 // catalogItem is implemented by both ClusterCatalogItem and ComputeInstanceCatalogItem.
 type catalogItem interface {
 	proto.Message
@@ -34,17 +51,6 @@ type catalogItem interface {
 	GetTemplate() string
 	GetFieldDefinitions() []*privatev1.FieldDefinition
 	GetMetadata() *privatev1.Metadata
-}
-
-const publishedFilter = "this.published == true"
-
-// addPublishedFilter appends a CEL clause that restricts results to published catalog items.
-// Used by public servers to hide unpublished items from end users.
-func addPublishedFilter(filter string) string {
-	if filter == "" {
-		return publishedFilter
-	}
-	return "(" + filter + ") && " + publishedFilter
 }
 
 // applyFieldDefinitions processes field definitions from a catalog item against a resource spec.
--- a/internal/servers/catalog_item_validation.go
+++ b/internal/servers/catalog_item_validation.go
@@ -36,6 +36,17 @@ type catalogItem interface {
 	GetMetadata() *privatev1.Metadata
 }
 
+const publishedFilter = "this.published == true"
+
+// addPublishedFilter appends a CEL clause that restricts results to published catalog items.
+// Used by public servers to hide unpublished items from end users.
+func addPublishedFilter(filter string) string {
+	if filter == "" {
+		return publishedFilter
+	}
+	return "(" + filter + ") && " + publishedFilter
+}
+
 // applyFieldDefinitions processes field definitions from a catalog item against a resource spec.
 // For non-editable fields: overrides user-provided values with the catalog item default.
 // For editable fields with user values: validates against the JSON Schema.
--- a/internal/servers/catalog_item_validation.go
+++ b/internal/servers/catalog_item_validation.go
@@ -14,6 +14,7 @@ language governing permissions and limitations under the License.
 package servers
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"strings"
@@ -28,6 +29,7 @@ import (
 	"google.golang.org/protobuf/types/known/structpb"
 
 	privatev1 "github.com/osac-project/fulfillment-service/internal/api/osac/private/v1"
+	"github.com/osac-project/fulfillment-service/internal/auth"
 )
 
 var (
@@ -51,6 +53,18 @@ func validateCELSyntax(filter string) error {
 		return fmt.Errorf("syntax error: %w", issues.Err())
 	}
 	return nil
+}
+
+// buildPublishedClause returns a CEL clause that shows published items to everyone
+// and unpublished items only to the user who created them.
+func buildPublishedClause(ctx context.Context) string {
+	user := auth.SubjectFromContext(ctx).User
+	return fmt.Sprintf("(this.published || this.metadata.creator == '%s')", user)
+}
+
+// isCreator checks whether the current user created the given item.
+func isCreator(ctx context.Context, itemCreator string) bool {
+	return itemCreator != "" && auth.SubjectFromContext(ctx).User == itemCreator
 }
 
 // catalogItem is implemented by both ClusterCatalogItem and ComputeInstanceCatalogItem.
--- a/internal/servers/catalog_item_validation.go
+++ b/internal/servers/catalog_item_validation.go
@@ -14,7 +14,6 @@ language governing permissions and limitations under the License.
 package servers
 
 import (
-	"context"
 	"encoding/json"
 	"fmt"
 	"strings"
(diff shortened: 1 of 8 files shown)


internal/servers/catalog_item_validation.go
```diff
@@ -36,6 +36,17 @@ type catalogItem interface {
...
+const publishedFilter = "this.published == true"
+
+// addPublishedFilter appends a CEL clause that restricts results to published catalog items.
+// Used by public servers to hide unpublished items from end users.
+func addPublishedFilter(filter string) string {
+	if filter == "" {
+		return publishedFilter
+	}
+	return "(" + filter + ") && " + publishedFilter
+}
```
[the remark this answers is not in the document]
tzvatot: False positive. The user filter goes through cel.Env.Compile() in FilterTranslator.Translate() (dao/filter_translator.go:203), which parses the entire concatenated string into a typed AST - not string-interpolated into SQL. The wrapping parens ensure correct grouping.

A user cannot bypass the published check because: (1) CEL is a compiled language, unbalanced parens fail compilation, (2) the CEL environment is strongly typed against the protobuf message so only valid field references compile, (3) the resulting AST is translated node-by-node into SQL by FilterTranslator.translate() which only supports a known set of operators and expression kinds. No path from CEL to SQL injection either.

internal/servers/compute_instance_catalog_items_server_test.go
```diff
@@ -182,12 +187,93 @@ var _ = Describe("Compute instance catalog items server", func() {
...
+			Expect(response.GetItems()[0].GetId()).To(Equal(targetID))
+		})
+
+		It("Get returns not found for unpublished object", func() {
+			createResponse, err := server.Create(ctx, publicv1.ComputeInstanceCatalogItemsCreateRequest_builder{
+				Object: publicv1.ComputeInstanceCatalogItem_builder{
+					Title:    "Unpublished item",
+					Template: "my-ci-template-id",
+				}.Build(),
+			}.Build())
```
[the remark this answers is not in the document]
tzvatot: Fixed in 9b537af - added explicit `Published: false` to match the cluster catalog test.

jhernand:
This means that tenant admins (which don't have access to the private API) will not be able to see the unpublished catalog items that they created: they will have to take note in a piece of paper of the identifier, because they will no longer be able to see them after changing the `published` flag to `false`. Is that the intent?

internal/servers/catalog_item_validation.go
```diff
@@ -36,6 +36,17 @@ type catalogItem interface {
 	GetMetadata() *privatev1.Metadata
 }
 
+const publishedFilter = "this.published == true"
```
jhernand: Published is a boolean, so `this.published` should be enough.

internal/servers/catalog_item_validation.go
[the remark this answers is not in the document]
jhernand: @tzvatot you are right in that this can't result in SQL injection, but it can result in bypassing the filter. For example the user could send `true) || (true`. That would result in `(true) || (true) && this.published` and the result of that is `true` regardless of the value of `this.published`.

internal/servers/catalog_item_validation.go
```diff
@@ -36,6 +36,17 @@ type catalogItem interface {
...
+
+// addPublishedFilter appends a CEL clause that restricts results to published catalog items.
+// Used by public servers to hide unpublished items from end users.
+func addPublishedFilter(filter string) string {
```
jhernand: These function's don't validate anything, why are they in a `_validation.go` file? In addition, such generic names in a package with dozens of servers will probably clash in the future. Consider making these methods of the corresponding server.

internal/servers/catalog_item_validation.go
tzvatot: Fixed - simplified to `this.published`.

internal/servers/catalog_item_validation.go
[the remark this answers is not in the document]
tzvatot: Good catch. Fixed by validating the user's filter is a complete, syntactically valid CEL expression (via `cel.Parse()`) before composing. A malicious input like `true) || (true` fails CEL parsing due to unbalanced parens and is rejected with `InvalidArgument`. Added test cases for the bypass attempt and for valid filters with `||` to confirm safe composition.

internal/servers/catalog_item_validation.go
tzvatot: Moved `addPublishedFilter` to methods on `ClusterCatalogItemsServer` and `ComputeInstanceCatalogItemsServer`. The `validateCELSyntax` helper stays in `catalog_item_validation.go` since it is validation and is shared by both servers.

tzvatot:
> This means that tenant admins (which don't have access to the private API) will not be able to see the unpublished catalog items that they created: they will have to take note in a piece of paper of the identifier, because they will no longer be able to see them after changing the `published` flag to `false`. Is that the intent?

Not the intent. Unpublished items should be visible to the user who created them.

Fixed by using creator-based visibility: the published filter is now this.published || this.metadata.creator == "<current_user>", so creators can see and manage their own unpublished items through the public API. The Get endpoint applies the same logic.

Global catalog items (tenant="shared") are managed via the private API, so this does not affect them.

internal/servers/catalog_item_validation.go
```diff
@@ -25,8 +29,44 @@ import (
...
+		return fmt.Errorf("syntax error: %w", issues.Err())
+	}
+	return nil
+}
+
+// buildPublishedClause returns a CEL clause that shows published items to everyone
+// and unpublished items only to the user who created them.
+func buildPublishedClause(ctx context.Context) string {
+	user := auth.SubjectFromContext(ctx).User
+	return fmt.Sprintf("(this.published || this.metadata.creator == '%s')", user)
```
jhernand: This is assuming that the "creator" will be the name of the user, which is  true today, but may change in the future. For example, we may want to use a group name, or something else.

And we are getting into the same realm of problems with this string concatenation: what if the "creator" can be used to trick the server into bypassing the check? At the very list consider using `%q` instead of `%s`.

My advice is to avoid this logic completely, and instead let the (in progress) authorization system handle this. For example, if a user wants to create a catalog item that isn't published yet, they can create it in a project (also in progress) that only they can see. Later, when they are happy, they can create it in the project where users need it.

tzvatot: Agreed. Reverted the creator-based visibility in 784ae80. The published filter is back to the simple this.published for all users. Unpublished item visibility will be handled by the authorization/projects system.

adriengentil:
does this change includes [this exception](https://github.com/osac-project/enhancement-proposals/tree/main/enhancements/catalog-items#public-vs-private-api-split):
```
Exception: a Tenant User can always Get a catalog item referenced by one of their existing CNAs, even if that item is unpublished.
```
?

internal/servers/catalog_item_validation.go
jhernand: My suggestion was to remove completely the "published" flag and all its associated logic.

tzvatot:
> does this change includes [this exception](https://github.com/osac-project/enhancement-proposals/tree/main/enhancements/catalog-items#public-vs-private-api-split):
> 
> ```
> Exception: a Tenant User can always Get a catalog item referenced by one of their existing CNAs, even if that item is unpublished.
> ```
> 
> ?

Yes, this is now implemented. A user can Get an unpublished catalog item if they have an existing cluster or compute instance that references it. The check uses a tenant-scoped query so it only considers resources visible to the caller. The List endpoint still filters unpublished items as before - only direct Get by ID has this exception, matching the enhancement proposal.

internal/servers/catalog_item_validation.go
```diff
@@ -25,8 +29,44 @@ import (
...
+}
```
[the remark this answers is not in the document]
tzvatot: No longer applicable - the creator-based visibility was reverted in 784ae80. The published filter is now the simple `this.published` with no user identity embedded in the CEL expression.

internal/servers/cluster_catalog_items_server.go
```diff
@@ -259,6 +267,16 @@ func (s *ClusterCatalogItemsServer) Update(ctx context.Context,
...
 
+func (s *ClusterCatalogItemsServer) addPublishedFilter(filter string) (string, error) {
+	if filter == "" {
+		return "this.published", nil
+	}
+	if err := validateCELSyntax(filter); err != nil {
+		return "", grpcstatus.Errorf(grpccodes.InvalidArgument, "invalid filter: %v", err)
+	}
+	return "(" + filter + ") && this.published", nil
+}
```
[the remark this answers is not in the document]
tzvatot: Intentional duplication per reviewer feedback - Juan asked to make these methods on the corresponding servers rather than shared package-level functions, to avoid generic name clashes in a package with dozens of servers. The logic is 7 lines and unlikely to drift independently.

jhernand (approved):

PR #600, opened 2026-05-27
OSAC-851: Implement delete projects in controller

CrystalChun:
# Description
Adds the deletion function to the projects controller. Deletion of a parent project is blocked if there are existing child projects.

# Testing

- [x] go build passes
- [x] All unit test pass

/cc @jhernand 

## Summary by CodeRabbit

* **New Features**
  * Projects cannot be deleted if child projects exist; deletion is blocked with a `DELETE_FAILED` status message.
  * Keycloak authorization resources are now automatically cleaned up during project deletion with proper error handling and status tracking.

* **Tests**
  * Added comprehensive test coverage for project deletion and cleanup scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/osac-project/fulfillment-service/pull/600?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


internal/controllers/project/project_reconciler_function.go
```diff
@@ -33,6 +33,12 @@ import (
 	"github.com/osac-project/fulfillment-service/internal/masks"
 )
 
+const (
+	// AnnotationKeycloakResourceID is the annotation key used to store the Keycloak authorization resource ID
+	// for a project. This ID is needed to clean up the Keycloak resource when the project is deleted.
+	AnnotationKeycloakResourceID = "keycloak.osac.io/resource-id"
```
jhernand: If we need this I think it is better to acknoledge it and add a field to the status of the project, something like "status.keylcoak_id".

internal/controllers/project/project_reconciler_function.go
```diff
@@ -274,6 +278,58 @@ func (t *task) checkCircularDependency(ctx context.Context, parent *privatev1.Pr
 
 // delete performs the deletion cleanup for a project.
 func (t *task) delete(ctx context.Context) error {
+	// Check for child projects before deletion
+	listResp, err := t.r.projectsClient.List(ctx, &privatev1.ProjectsListRequest{
+		Filter: proto.String(fmt.Sprintf("this.spec.parent == '%s'", t.project.GetId())),
```
jhernand, line 278: I didn't realize we have "spec.parent". I think we shouldn't. At some point we should have a "metadata.project" in all objects, and once we have that we should use it also for this purpose. But it is OK for now.

CrystalChun: Ah do we intend the `metadata.project` to point to the parent project?

jhernand: Yes, that was my intent: "metadata.project" is, for any object, the project that contains it, so that naturally extends to being the "parent" of a project.

CrystalChun: That makes sense! I'll add a comment to change it once the metadata.project is available. Thank you!

internal/controllers/project/project_reconciler_function.go
CrystalChun: Good idea, added the status field and removed the annotation. Thank you!

internal/controllers/project/project_reconciler_function.go
```diff
@@ -274,6 +272,58 @@ func (t *task) checkCircularDependency(ctx context.Context, parent *privatev1.Pr
 
 // delete performs the deletion cleanup for a project.
 func (t *task) delete(ctx context.Context) error {
+	// Check for child projects before deletion
+	listResp, err := t.r.projectsClient.List(ctx, &privatev1.ProjectsListRequest{
+		//TODO: Update to use metadata.project once OSAC-1064 is implemented
+		Filter: proto.String(fmt.Sprintf("this.spec.parent == '%s'", t.project.GetId())),
+		Limit:  proto.Int32(1), // We only need to know if any exist
```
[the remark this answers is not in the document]
CrystalChun: The project_id is system generated and shouldn't contain quote/control characters

jhernand (approved):

PR #614, opened 2026-05-29
OSAC-1060: Restrict resource creation in shared tenant

SiddarthR56:
## Summary
Goal: Prevent accidental or unauthorized creation of resources in the reserved "system" and "shared" tenants. These tenants are special (system-internal and globally-shared infrastructure) and should only allow object creation for specific resource types that legitimately belong there.

## Changes
New tenant guard in generic_server.go:
- Introduces an allowedTenants field on GenericServer that defaults to DefaultAllowedTenants (all tenants except system and shared).
- On Create: returns PermissionDenied if the assigned tenant is not in the allowed set.
- On Update: returns PermissionDenied if an object is being moved into a disallowed tenant.
- New builder methods AddAllowedTenants(...) and RemoveAllowedTenants(...) let individual servers opt in or out.

New constant in tenancy_logic.go:
- DefaultAllowedTenants = AllTenants - {SystemTenant, SharedTenant} — a reusable set representing where objects can normally be created.

Servers that explicitly opt-in to shared-tenant creation (via AddAllowedTenants(auth.SharedTenant)):
- Hubs
- Host Types
- Roles
- Cluster Catalog Items
- Compute Instance Catalog Items
- Compute Instance Templates
- Cluster Templates
- Network Classes
- Public IP Pools

## Test updates:

- A new dedicated test file servers_shared_tenant_test.go validates the restriction across all server types.
- Existing tenancy tests were refactored/cleaned up — removed redundant shared-tenant creation tests from individual server test files now that there's a centralized test suite for it.
- Integration tests updated to account for the new restriction.

## Summary by CodeRabbit

* **New Features**
  * Tenant allowlist enforcement for create/update operations; requests using reserved tenants (system/shared) are denied.
  * Servers for several private/public resources now support configurable tenant-scoping so some endpoints can allow the shared tenant.

* **Bug Fixes**
  * Prevented unauthorized tenant assignment across networks, subnets, compute resources, IPs, projects and related resources.

### DIFF
--- a/internal/servers/compute_instance_templates_server_test.go
+++ b/internal/servers/compute_instance_templates_server_test.go
@@ -14,7 +14,6 @@ language governing permissions and limitations under the License.
 package servers
 
 import (
-	"context"
 	"fmt"
 
 	. "github.com/onsi/ginkgo/v2"
@@ -22,46 +21,9 @@ import (
 	"google.golang.org/protobuf/types/known/fieldmaskpb"
 
 	publicv1 "github.com/osac-project/fulfillment-service/internal/api/osac/public/v1"
-	"github.com/osac-project/fulfillment-service/internal/database"
 )
 
 var _ = Describe("Compute instance templates server", func() {
-	var (
-		ctx context.Context
-		tx  database.Tx
-	)
-
-	BeforeEach(func() {
-		var err error
-
-		// Create a context:
-		ctx = context.Background()
-
-		// Prepare the database pool:
-		db, err := server.NewInstance().Build()
-		Expect(err).ToNot(HaveOccurred())
-		DeferCleanup(db.Close)
-		pool, err := db.Pool(ctx)
-		Expect(err).ToNot(HaveOccurred())
-		DeferCleanup(pool.Close)
-
-		// Create the transaction manager:
-		tm, err := database.NewTxManager().
-			SetLogger(logger).
-			SetPool(pool).
-			Build()
-		Expect(err).ToNot(HaveOccurred())
-
-		// Start a transaction and add it to the context:
-		tx, err = tm.Begin(ctx)
-		Expect(err).ToNot(HaveOccurred())
-		DeferCleanup(func() {
-			err := tm.End(ctx, tx)
-			Expect(err).ToNot(HaveOccurred())
-		})
-		ctx = database.TxIntoContext(ctx, tx)
-	})
-
 	Describe("Builder", func() {
 		It("Creates server with logger and tenancy logic", func() {
 			// Create the public server:
--- a/internal/servers/servers_shared_tenant_test.go
+++ b/internal/servers/servers_shared_tenant_test.go
@@ -14,8 +14,6 @@ language governing permissions and limitations under the License.
 package servers
 
 import (
-	"context"
-
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	"go.uber.org/mock/gomock"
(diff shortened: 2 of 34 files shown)


jhernand:
The "system" tenant should probably be as disallowed as "shared", and both should probably be disallowed by default. I'd suggest you add to the server an "allowedTenants" field. This should be a set of strings, and the default value should be the universal set minus "system" and "shared". Then in the builder you can add methods like "AddAllowedTenants" and "RemoveAllowedTenants", so that only in the very special cases where we do allow creation of these objects we explicitly add them. The only such case, I think is currently for cluster templates and compute instance templates.

I also noticed that you were using "shared" in the unit tests, and that you replaced it with "system". That makes those tests even more wrong, as objects may exist in the "shared" tenant, but certainly not in the "system" tenant. Can you change those tests to use a fake tenant like "my-tenant" instead?

Finally try to refer to objects as "objects" in the error messages and documentation, not "resources".

SiddarthR56:
> The "system" tenant should probably be as disallowed as "shared", and both should probably be disallowed by default. I'd suggest you add to the server an "allowedTenants" field. This should be a set of strings, and the default value should be the universal set minus "system" and "shared". Then in the builder you can add methods like "AddAllowedTenants" and "RemoveAllowedTenants", so that only in the very special cases where we do allow creation of these objects we explicitly add them. The only such case, I think is currently for cluster templates and compute instance templates.
> 
> I also noticed that you were using "shared" in the unit tests, and that you replaced it with "system". That makes those tests even more wrong, as objects may exist in the "shared" tenant, but certainly not in the "system" tenant. Can you change those tests to use a fake tenant like "my-tenant" instead?
> 
> Finally try to refer to objects as "objects" in the error messages and documentation, not "resources".

Wasn't entirely sure what the system tenant was intended for—I assumed it was mainly for internal object creation.

I updated the PR to block creation of any objects in both system and shared tenants by default, switched the tests to use a fake tenant

jhernand (approved):
Looks good, thanks @SiddarthR56 .

internal/servers/servers_shared_tenant_test.go
```diff
@@ -0,0 +1,701 @@
...
+				}.Build(),
+			}.Build())
+			Expect(updateResponse).To(BeNil())
+			Expect(err).To(HaveOccurred())
+			status, ok := grpcstatus.FromError(err)
+			Expect(ok).To(BeTrue())
+			Expect(status.Code()).To(Equal(grpccodes.PermissionDenied))
+			Expect(status.Message()).To(ContainSubstring("shared"))
+			Expect(status.Message()).To(ContainSubstring("not allowed"))
+		})
```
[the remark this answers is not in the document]
SiddarthR56: Not adding this — the behavior is already covered by the "Preserves private data during update" test in clusters_server_test.go (around line 888). That test creates a cluster via DAO with SharedTenant, then updates it through the public server, which implicitly proves the carve-out works (the update would fail otherwise).

internal/servers/compute_instance_templates_server_test.go
```diff
@@ -22,44 +22,13 @@ import (
...
-
-		// Start a transaction and add it to the context:
-		tx, err = tm.Begin(ctx)
-		Expect(err).ToNot(HaveOccurred())
-		DeferCleanup(func() {
-			err := tm.End(ctx, tx)
-			Expect(err).ToNot(HaveOccurred())
-		})
-		ctx = database.TxIntoContext(ctx, tx)
+		ctx = setupTestDB()
```
jhernand: This is a good idea, but it adds a lot of noise to the patch. And it is also here: https://github.com/osac-project/fulfillment-service/pull/635 . I'd suggest you remove these changes from the patch, if possible.

internal/servers/servers_shared_tenant_test.go
```diff
@@ -0,0 +1,719 @@
...
+
+	privatev1 "github.com/osac-project/fulfillment-service/internal/api/osac/private/v1"
+	publicv1 "github.com/osac-project/fulfillment-service/internal/api/osac/public/v1"
+	"github.com/osac-project/fulfillment-service/internal/auth"
+	"github.com/osac-project/fulfillment-service/internal/database/dao"
+)
+
+var _ = Describe("Reserved tenant creation restriction", func() {
+	Describe("Cluster", func() {
+		var templateID string
```
akshaynadkarni, line 261: The new centralized test file validates that non-opted-in servers _reject_ creation in the shared tenant, but it doesn't test that opted-in servers (Hubs, HostTypes, Roles, ClusterCatalogItems, etc.) _accept_ creation in the shared tenant, does it? If someone accidentally removes an `AddAllowedTenants(auth.SharedTenant)` call from one of those servers, no test would catch the regression.

Consider adding a companion `Describe` block (or a table-driven test) that creates objects via each opted-in server with `Tenant: auth.SharedTenant` and asserts success. This would provide bidirectional coverage: rejection for non-opted-in and acceptance for opted-in.

jhernand (approved):

akshaynadkarni (approved):
@SiddarthR56 Changes LGTM. 

However, you will need to: 
- resolve branch conflicts
- ensure the vmaas test passes: Did you put up a PR to fix the vmaas tests?

PR #622, opened 2026-06-01
OSAC-1079: Bump chart versions from 0.0.0 to 0.1.0

eliorerz:
## Summary
- Bump all chart versions from `0.0.0` to `0.1.0`: service, ca, keycloak, postgres, prometheus
- Also bumps `appVersion` in `charts/service/Chart.yaml` from `0.0.0` to `0.1.0`

Part of Helm chart production readiness ([OSAC-1072](https://redhat.atlassian.net/browse/OSAC-1072)). Replaces placeholder versions with real semver as the first step toward OCI publication.

## Test plan
- [ ] `helm lint charts/service`
- [ ] `helm lint charts/ca`
- [ ] `helm lint charts/keycloak`
- [ ] `helm lint charts/postgres`
- [ ] `helm lint charts/prometheus`

## Release Notes

* **Chores**
  * Updated deployment configuration chart versions to 0.1.0.

### DIFF
--- a/charts/service/Chart.yaml
+++ b/charts/service/Chart.yaml
@@ -15,5 +15,5 @@ apiVersion: v2
 name: fulfillment-service
 description: Fulfillment service
 type: application
-version: 0.0.0
-appVersion: 0.0.0
+version: 0.1.0
+appVersion: 0.1.0
--- a/it/charts/keycloak/Chart.yaml
+++ b/it/charts/keycloak/Chart.yaml
@@ -15,5 +15,5 @@ apiVersion: v2
 name: keycloak
 description: A Helm chart for Keycloak for use in OSAC development and testing environments (not for production)
 type: application
-version: 0.0.0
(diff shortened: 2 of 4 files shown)


jhernand:
The "ca", "keycloak", "postgres" and "prometheus" are all intended only for the integration tests only and they shouldn't be published or used for anything else. Therefore changing the version isn't needed. In addition they have all been moved to "it/charts".

The version of the application is automatically set when the chart is built and published, here: https://github.com/osac-project/fulfillment-service/blob/main/.github/workflows/publish-charts.yaml . So I think this isn't needed either.

eliorerz:
Closing — the publish-charts.yaml workflow already handles versioning at publish time by overriding --version and --app-version from the git tag. The in-source Chart.yaml version (0.0.0) is a static placeholder that is never used in published charts, so bumping it is unnecessary.

PR #634, opened 2026-06-03
OSAC-1205: Add InstanceType resource — proto, codegen, migration, and server integration

ygalblum:
## Summary

Adds the InstanceType resource to fulfillment-service, covering the full vertical slice from proto definitions through database migration:

- **OSAC-1206**: Define InstanceType proto schema for public and private APIs (type + service protos with full CRUD RPCs, HTTP annotations, OpenAPI docs)
- **OSAC-1207**: Generate Go code from proto definitions (buf generate output for both public/private APIs)
- **OSAC-1208**: Create database migration 51 with `instance_types` and `archived_instance_types` tables, wire into GenericServer `setPayload()`, and add migration test coverage

## Jira

Epic: [OSAC-1205](https://redhat.atlassian.net/browse/OSAC-1205)

## Test plan

- [x] `buf lint` passes
- [x] `go build ./...` passes
- [x] Migration test follows existing pattern (51_create_instance_types_tables_test.go)
- [ ] `ginkgo run -r internal` — unit tests pass
- [ ] `ginkgo run it` — integration tests pass

## Summary by CodeRabbit

* **New Features**
  * Added instance type management with full CRUD, lifecycle states (active/deprecated/obsolete), deprecation metadata and replacement suggestions
  * Instance types available via public and private APIs and emitted in event notifications for real-time updates
* **Tests**
  * Added migration and integration tests to validate instance type persistence and retrieval
* **Chores**
  * Introduced database migration to persist instance types and archived records


proto/public/osac/public/v1/instance_type_type.proto
```diff
@@ -0,0 +1,90 @@
...
+  //
+  // DNS-label style: lowercase alphanumeric and hyphens, max 63 chars, must start with a letter
+  // (e.g., "standard-4-16", "highmem-8-64").
+  string name = 1;
+
+  // Metadata of the instance type.
+  Metadata metadata = 2;
+
+  // Number of CPU cores. Immutable after creation.
+  int32 cores = 3;
```
jhernand: Consider adding "spec" and "status" messages/fields, otherwise if we eventually need to do any reconciliation of these types it will be difficult to distinguish between the desired and actual state. I'd say that all those fields below belong into the "spec" message.

proto/public/osac/public/v1/instance_type_type.proto
```diff
@@ -0,0 +1,90 @@
...
+  // Human friendly description of the instance type, using Markdown format.
+  string description = 5;
+
+  // Current lifecycle state of the instance type.
+  InstanceTypeState state = 6;
+
+  // Deprecation details. Only meaningful when state is DEPRECATED or OBSOLETE.
+  InstanceTypeDeprecation deprecation = 7;
+
+  // Unique identifier of the instance type. Holds the same value as the name field for GenericDAO compatibility.
```
jhernand: That may change, and the user doesn't need to know that, don't make that explicit in the documentation. Also, make it the first field, like in all other types.

proto/private/osac/private/v1/instance_type_type.proto
```diff
@@ -0,0 +1,44 @@
...
+//
+
+syntax = "proto3";
+
+package osac.private.v1;
+
+import "google/protobuf/timestamp.proto";
+import "osac/private/v1/metadata_type.proto";
+
+enum InstanceTypeState {
```
jhernand, line 22: Historically, and unfortunately, we haven't been writing documentation for the private types, but we should. Please document this like you documented the public API.

internal/database/migrations/51_create_instance_types_tables_test.go
```diff
@@ -0,0 +1,80 @@
...
+			`select exists (
+				select from information_schema.tables
+				where table_name = 'instance_types'
+			)`,
+		)
+		err = row.Scan(&exists)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(exists).To(BeTrue())
+
+		// Verify the archived_instance_types table exists:
```
jhernand: This test is redundant: the "database_schema_test.go" file already contains logic to check that the archived table exists for all the object tables.

internal/database/migrations/51_create_instance_types_tables.up.sql
```diff
@@ -0,0 +1,47 @@
...
+-- Create the instance_types tables following the generic schema pattern.
+
+create table instance_types (
+  id text not null primary key,
+  name text not null default '',
+  creation_timestamp timestamp with time zone not null default now(),
+  deletion_timestamp timestamp with time zone not null default 'epoch',
+  finalizers text[] not null default '{}',
+  creator text not null default '',
+  tenant text not null default '',
```
jhernand, line 23: There should be a foreign key constraint on this column, referencing the "organizations" table, something like this:

```sql
alter table add constraint instance_types_tenant_fk foreign key (tenant) references organizations (id)',
```

internal/database/migrations/51_create_instance_types_tables.up.sql
```diff
@@ -0,0 +1,47 @@
...
+  labels jsonb not null default '{}'::jsonb,
+  annotations jsonb not null default '{}'::jsonb,
+  data jsonb not null,
+  version integer not null default 0
+);
```
jhernand, line 28: As we are moving towards making tenants and names immutable, it would be nice if you added this:

```sql
-- Attach the trigger to the table to enforce immutability of the id, name and tenant columns:
create trigger check_immutable_columns
  before update on instance_types
  for each row
  execute function check_immutable_columns('id', 'name', 'tenant');
```

We will do that later anyhow, but it is nice to have it now.

proto/public/osac/public/v1/instance_type_type.proto
ygalblum: I see your point. This was implemented based on the EP. But, it obviously this can be done. I guess it will be something like:
```
enum InstanceTypeState {
  INSTANCE_TYPE_STATE_UNSPECIFIED = 0;
  ACTIVE = 1;       // Fully available for new VM creation
  DEPRECATED = 2;   // Available with warnings, migration recommended
  OBSOLETE = 3;     // Not available for new VMs, visible for lookups only
}

message InstanceTypeDeprecation {
  InstanceTypeState state = 1;                   // DEPRECATED or OBSOLETE (matches parent InstanceType.state)
  string replacement = 2;                        // Optional: suggested replacement instance type name
  google.protobuf.Timestamp deprecated = 3;      // Optional: when deprecation was announced (auto-set on DEPRECATED transition)
  google.protobuf.Timestamp obsolete = 4;        // Optional: when it becomes/became obsolete (admin-specified or auto-set)
}

message InstanceTypeSpec {
  string name = 1;                               // Primary identifier (globally unique, immutable, e.g., "standard-4-16")
  Metadata metadata = 2;                         // Standard metadata (name field matches top-level name)
  int32 cores = 3;                               // Immutable compute specification
  int32 memory_gib = 4;                          // Immutable compute specification
  string description = 5;                        // Mutable descriptive text
  InstanceTypeState state = 6;                   // Mutable lifecycle state
  InstanceTypeDeprecation deprecation = 7;       // Optional: only set when state is DEPRECATED or OBSOLETE
}

message InstanceTypeStatus {
  InstanceTypeState state = 1;
  InstanceTypeDeprecation deprecation = 2;
} 

message  InstanceType {
  string id = 1;
  Metadata metadata = 2;
  InstanceTypeSpec spec = 3;
  InstanceTypeStatus status = 4;
}
```
@avishayt since you wrote the EP, what do you think?

proto/public/osac/public/v1/instance_type_type.proto
ygalblum: I'm not sure what you mean by the user doesn't need to know. What doesn't he need to know the value or the fact that it's unique.

Truthfully, I'm not sure if I want to keep this field or not. Mainly because it is redundant as the name is already the unique identifier. I added it (on top of what the EP states) following a discussion we had about human vs machine identifiers.

Having said that, in the previous comment the option I added has `id` in the base message and then `name` in the `spec`. Maybe we need to consolidate on the differet fields (`id`, `spec.name` and `metadata.name`)

internal/database/migrations/51_create_instance_types_tables.up.sql
ygalblum: Added here: 1bd2692

internal/database/migrations/51_create_instance_types_tables.up.sql
ygalblum: Added here: 7198c47

internal/database/migrations/51_create_instance_types_tables_test.go
ygalblum: Removed in 1dddd11

proto/private/osac/private/v1/instance_type_type.proto
ygalblum: I'll address this once we decide whether we should change the structure

proto/public/osac/public/v1/instance_type_type.proto
jhernand: We shoudn't and can't deviate from the existent layout: objects need to have a top level `id` field which needs to be the primary key. It is used by operations like `Get`, `Update` and `Delete` to identify the object, and is used in the URL of REST requests. If we deviate from that the CLI (which uses gRPC reflection and assumes that layout) will break and won't be able to work with these objects.

I didn't notice the top level `name` that you added. It needs to be `metadata.name`. It already exists, and you shouldn't add a new one.

The use of `spec` and `status` is debatable, but I very much prefer if we use them. We had cases already where we started without and then had to add them later, and move fields around. In this particular case I'd say that all the fields belong into `spec`, as all of then indicate the desired state of the object. So `status` should exist, but empty , for now.

So, it should look like this, I think (plus documentation):

```protobuf
message InstanceType {
  string id = 1;
  Metadata metadata = 2;
  InstanceTypeSpec spec = 3;
  InstanceTypeStatus status = 4;
}
```

proto/public/osac/public/v1/instance_type_type.proto
jhernand: I mean that making `id` and `metadata.name` identical for these objects is an implementation detail that the user shouldn't rely on, so it shouldn't be stated in the documentation.

proto/public/osac/public/v1/instance_type_type.proto
```diff
@@ -0,0 +1,90 @@
...
+// serialized independently.
+message InstanceTypeDeprecation {
+  // Current deprecation state (DEPRECATED or OBSOLETE). Mirrors parent InstanceType.state.
+  InstanceTypeState state = 1;
+
+  // Suggested replacement instance type name (e.g., "standard-8-32").
+  string replacement = 2;
+
+  // When deprecation was announced. Auto-set on transition to DEPRECATED.
+  optional google.protobuf.Timestamp deprecated = 3;
```
jhernand: Message types, like `google.protobuf.Timestamp`, always have explicit presence, so there is no need for `optional`.

proto/public/osac/public/v1/instance_type_type.proto
ygalblum: I agree about the structure and don't mind emptying the status. I agree it is merely a reflection of some fields in the spec. I've seen this pattern in some CRDs , e.g. `StorageProfile`, but it's really not necessary.

As for the `id`, we have decided that the name is the identifier. So, where should that be set and/or enforced? Do we expect users to not pass an ID and the server will set the ID according to our expectation?

proto/public/osac/public/v1/instance_type_type.proto
ygalblum: But we want them to. We want them to know that if there's an `InstanceType` named `small-1-2`, they can reference this name in the parameter passed in the `ComputeInstance`

proto/public/osac/public/v1/instance_type_type.proto
jhernand: Also, I'd suggest you name these `deprecation_timestamp` and `retirement_timestamp`, or `..._timestamp`. That would be more consistent with the existing `metadata.creation_timestamp`, `metadata.deletion_timestamp`, etc.

proto/public/osac/public/v1/instance_type_type.proto
```diff
@@ -0,0 +1,90 @@
...
+// lifecycle-managed instance types.
+//
+// This is similar to the instance type concepts used by cloud providers (AWS EC2 instance types, Azure VM sizes,
+// GCP machine types).
+message InstanceType {
+  // Primary user-facing identifier of the instance type (globally unique, immutable).
+  //
+  // DNS-label style: lowercase alphanumeric and hyphens, max 63 chars, must start with a letter
+  // (e.g., "standard-4-16", "highmem-8-64").
+  string name = 1;
```
jhernand: Remember to remove this: it is already in `metadata.name`. Currently it isn't mandatory, or unique, but will be soon.

proto/public/osac/public/v1/instance_type_type.proto
jhernand: Currently users can't set the id, because it needs to be globally unique, even for different tenants. The system can't copy the name provided by the user and use it as the id either. The id needs to be generated by the system. Currently we generate a UUIDV7, here:

https://github.com/osac-project/fulfillment-service/blob/5b919b8947a6713758f34a3b44070072ddadc736/internal/database/dao/generic_dao_create.go#L65-L69

Only in very special cases, where the object is created by the system, we allow it to specify the id explicitly. That happens, for example, for tenants and cluster templates because they are created by the system and are global (tenants) or shared by all tenants (cluster templates).

When we have completed the work to add projects, and make names mandatory immutable and unique you will be able to do what you want. Meanwhile you have to stick to the current rules.

proto/public/osac/public/v1/instance_type_type.proto
jhernand: That is reasonable and doable, except that you can't expect the name to be unique. Your implementation of the server will have to accept either an id or a name. When it is an id it can store it directly. When it is a name it will have to find the id corresponding to that name and store it. That is, for example, what we do when the user specifies a host type to create a cluster:

https://github.com/osac-project/fulfillment-service/blob/5b919b8947a6713758f34a3b44070072ddadc736/internal/servers/private_clusters_server.go#L178-L187

proto/public/osac/public/v1/instance_type_type.proto
avishayt: Thanks Juan, great catch regarding spec and status - I definitely agree.  However, I don't think ID makes sense here.  For now, InstanceTypes are created only by the Cloud Provider Admin with global scope.  It only makes sense to refer to them by name (e.g., "medium") rather than UUID.

proto/public/osac/public/v1/instance_type_type.proto
avishayt: As I wrote about InstanceTypes are global in scope and created by the Cloud Provider Admin.  They should be referenced only by name (unique and immutable).

proto/public/osac/public/v1/instance_type_type.proto
ygalblum: Addressed here: 8fbf74f

proto/public/osac/public/v1/instance_type_type.proto
ygalblum: Addressed here: 8fbf74f

proto/private/osac/private/v1/instance_type_type.proto
ygalblum: Addressed here: 8fbf74f

proto/public/osac/public/v1/instance_type_type.proto
ygalblum: See: 8fbf74f

I've aligned the structure to the rest of OSAC. At this stage, I also left the `id` field separate from `metadata.name`. I think we should keep this separation for now as we want to have a solution level change to use names. So, no point in delaying the progress here.

proto/public/osac/public/v1/instance_type_type.proto
ygalblum: See: 8fbf74f

jhernand (approved):

proto/private/osac/private/v1/instance_type_type.proto
```diff
@@ -0,0 +1,102 @@
...
+
+  // Human friendly description of the instance type, using Markdown format.
+  string description = 3;
+
+  // Current lifecycle state of the instance type.
+  InstanceTypeState state = 4;
+
+  // Deprecation details. Only meaningful when state is DEPRECATED or OBSOLETE.
+  InstanceTypeDeprecation deprecation = 5;
+}
```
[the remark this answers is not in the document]
ygalblum: This enforcement will handled in the next phase of implementation, when the server is implemented

PR #650, opened 2026-06-05
Expose available IP count in public pools CLI & add describe for publicipattachment

SiddarthR56:
## Summary
Add a status field with available count to the public PublicIPPool proto, allowing tenants to see how many IPs are free for allocation
Display an "AVAILABLE" column in osac get publicippool table output and detail view
Add osac describe publicipattachment command for inspecting attachment details by ID or name

## Changes
Proto (proto/public/osac/public/v1/public_ip_pool_type.proto):
- Added PublicIPPoolStatus message with available field (field number 6, matching private proto)
- Added status field to the public PublicIPPool message

CLI — get publicippool (internal/cmd/cli/get/publicippool/get_publicippool_cmd.go):
- Added "AVAILABLE" column to the list table view
- Added "Available:" row to the single-pool detail view

CLI — describe publicipattachment (new):
- Added internal/cmd/cli/describe/publicipattachment/describe_publicipattachment_cmd.go
- Registered in internal/cmd/cli/describe/describe_cmd.go
- Resolves by ID or name, renders: ID, Name, Public IP, Compute Instance, Public IP Address, State, Message

Naming rationale
Public CLI uses "AVAILABLE" (standalone column, consistent with AWS AvailableIpAddressCount)
Private rendering table keeps "FREE" (alongside TOTAL/ALLOCATED, consistent with kubectl-view-allocations)

## Testing
```
$ ./osac get publicippools
ID                                    NAME     CIDRS            IP-FAMILY  AVAILABLE
019dfd63-044e-7400-b353-541a809fa401  pool1  192.168.99.0/24  IPV4       253

$ ./osac describe publicipattachment attachment1
ID:                 019e5fee-0742-78b7-8c4a-e2501f44783a
Name:               attachment1
Public IP:          019728a4-3f5c-7def-8abc-1234567890ab
Compute Instance:   01972f1b-a4e9-7c82-9def-abcdef123456
Public IP Address:  192.168.1.42
State:              READY
Message:            -
```

## Summary by CodeRabbit

* **New Features**
  * Pool availability metrics are now visible in CLI and API outputs for better resource management. List views include an additional availability count column alongside other pool information for quick reference. Detailed pool views display comprehensive capacity and status information. This enhancement provides operators with improved visibility into resource allocation, pool readiness, and available capacity.


jhernand (approved):

internal/cmd/cli/describe/describe_cmd.go
```diff
@@ -33,6 +34,7 @@ func Cmd() *cobra.Command {
 	result.AddCommand(cluster.Cmd())
 	result.AddCommand(computeinstance.Cmd())
 	result.AddCommand(publicip.Cmd())
+	result.AddCommand(publicipattachment.Cmd())
```
akshaynadkarni, line 37: Do you think we should add `publicipattachment` to the alias table and subcommands assertion in `describe_cmd_test.go`? Something like:

- `Entry("publicipattachment", publicipattachment.Cmd, "publicipattachments")` in the alias table
- `"publicipattachment"` to the `ContainElements` list

internal/cmd/cli/get/publicippool/get_publicippool_cmd.go
```diff
@@ -117,7 +117,8 @@ func renderPoolTable(w *terminal.Console, pools []*publicv1.PublicIPPool) {
 			cidrs = "-"
 		}
 		ipFamily := strings.TrimPrefix(p.GetSpec().GetIpFamily().String(), "IP_FAMILY_")
-		fmt.Fprintf(writer, "%s\t%s\t%s\t%s\n", p.GetId(), name, cidrs, ipFamily)
+		available := fmt.Sprintf("%d", p.GetStatus().GetAvailable())
```
akshaynadkarni, line 120: Will this return `0` instead of `-` when a pool has no status? That could be misleading since `0` suggests no IPs are available rather than status not being populated yet. Same thing for line 146.

Maybe we should do something like this for consistency with the other columns?

```go
available := "-"
if p.GetStatus() \!= nil {
    available = fmt.Sprintf("%d", p.GetStatus().GetAvailable())
}
```

WDYT?

proto/public/osac/public/v1/public_ip_pool_type.proto
```diff
@@ -45,6 +48,12 @@ message PublicIPPoolSpec {
   IPFamily ip_family = 3 [(google.api.field_behavior) = IMMUTABLE];
 }
 
+// Capacity information for a PublicIPPool exposed to tenants.
+message PublicIPPoolStatus {
+  // Number of IP addresses available for new allocations from this pool.
+  int64 available = 6 [(google.api.field_behavior) = OUTPUT_ONLY];
```
akshaynadkarni, line 54: Curious: you're using field number 6 here, is this intentional to match the private proto's field numbering?

SiddarthR56: Yes, intentional — matches the private proto's field number for readability.

internal/cmd/cli/get/publicippool/get_publicippool_cmd.go
SiddarthR56: The status will never be nil for pools exposed via the public API (the server only returns pools that are READY with available > 0). That said, I can add a defensive guard if you'd prefer.

PR #658, opened 2026-06-08
OSAC-1156: Add implementation_strategy to SecurityGroup proto spec

ori-amizur:
SecurityGroup now carries its own implementation_strategy field (OUTPUT_ONLY, IMMUTABLE), set to "network_policy" at creation time. This replaces the annotation-based approach where the strategy was copied from the parent VirtualNetwork's K8s CR.

- Proto: add implementation_strategy field to SecurityGroupSpec
- Server: set "network_policy" on Create, preserve on Update
- Controller: pass strategy via spec field, remove annotation and parent VN lookup for strategy
- Migration 52: backfill existing records with "network_policy"

## Summary by CodeRabbit

* **New Features**
  * Security groups now include a system-managed implementation strategy field that specifies the backend used to enforce security rules. This field is automatically set at creation and is immutable thereafter.

* **Chores**
  * Updated Go toolchain to version 1.26.3.
  * Updated related dependency versions.


internal/servers/private_security_groups_server.go
```diff
@@ -184,6 +189,11 @@ func (s *PrivateSecurityGroupsServer) Update(ctx context.Context,
 		return
 	}
 
+	// Preserve immutable implementation_strategy from existing object
+	if request.GetObject().GetSpec() != nil && existingSecurityGroup.GetSpec() != nil {
+		request.GetObject().GetSpec().SetImplementationStrategy(existingSecurityGroup.GetSpec().GetImplementationStrategy())
+	}
+
```
danmanor, line 197: Should we return an error if the caller sends a different value instead of silently overwriting it? OUTPUT_ONLY should prevent well-behaved clients from setting it, but if something does slip through, this would quietly eat it and that could be confusing to debug.

proto/private/osac/private/v1/security_group_type.proto
```diff
@@ -87,6 +87,14 @@ message SecurityGroupSpec {
...
   // Example: Allow all outbound traffic, or restrict to specific destinations
   repeated SecurityRule egress = 3;
+
+  // Implementation strategy that determines the backend used to enforce security rules.
+  // Set by the system at creation time; currently always "network_policy" (standard Kubernetes
+  // NetworkPolicy). Immutable after creation.
+  string implementation_strategy = 4 [
+    (google.api.field_behavior) = OUTPUT_ONLY,
+    (google.api.field_behavior) = IMMUTABLE
+  ];
```
jhernand, line 99: Can this be an enum with documented values?

internal/servers/private_security_groups_server.go
ori-amizur: Done — added validation in `validateImmutableFieldsSecurityGroup` that rejects the update with `InvalidArgument` if the caller explicitly sends a non-empty `implementation_strategy` that differs from the stored value. The silent preservation is still needed afterward to handle the normal case where well-behaved clients omit the OUTPUT_ONLY field (it arrives as empty string).

proto/private/osac/private/v1/security_group_type.proto
ori-amizur: Keeping as `string` for consistency with VirtualNetwork and PublicIPPool which use the same `string implementation_strategy` pattern. Converting to an enum would change the `protojson` serialization format (stored `"network_policy"` vs enum name `"SECURITY_GROUP_IMPLEMENTATION_STRATEGY_NETWORK_POLICY"`) and break existing database records without a data migration. Documented the known values in the comment instead.

danmanor:
looks good to me, @jhernand WDYT ?

danmanor:
@ori-amizur rebase

jhernand:
> looks good to me, @jhernand WDYT ?

I still think this needs to be an enum. I understand that we made the mistake of not using an enum in the other types, and that is a good excuse to make the mistake here as well. I will approve if you commit to change to an enum here and in the other places, including migrations of existing data, etc.

jhernand (changes requested):

PR #664, opened 2026-06-09
Add UI deployment and report internal/external OIDC url

rawagner:
Separate internal and external Keycloak URLs in capabilities; add UI to helm chart

The `AuthnCapabilities` proto response now advertises `TokenIssuer` objects (each with `internal_url` and `external_url`) instead of a flat list of issuer strings. This fixes CLI login from outside the cluster: the CLI previously received the internal Keycloak service URL and could not reach it. Now the external URL is used for OAuth flows while the internal URL is used for in-cluster JWT validation and OIDC discovery.

The `--grpc-authn-trusted-token-issuers` flag now accepts `internalUrl=externalUrl` pairs. `auth.externalIssuerUrl` is added as a required helm value.

The fulfillment-service helm chart now includes a UI deployment (Deployment, ConfigMap, Service, Route/HTTPRoute). A [GitHub Actions workflow](https://github.com/osac-project/osac-ui/pull/18) is added to the osac-ui repo to publish the container image to ghcr.io on every push to main.


jhernand:
I think this doesn't belong in this repository. The UI should have its own Helm chart, in its own repository.

Regarding the trusted token issuers reported by the server, it should always be a URL that is reachable and usable by external clients. If the user configures a "issuerUrl" that is not reachable by the CLI, or by the UI, or by other systems, then that is a user configuration error.

rawagner:
UI is a component of OSAC as any other. Why separate helm chart only for UI ?

Regarding the internal/external URL - when you deploy UI, the UI's proxy runs in-cluster and needs to talk to keycloak service. The frontend part runs in user's browser and needs the external keycloak URL. So a client (UI) needs both of these URLs. Should capabilities API be a single source of truth ? Or we want to split this config into Capabilities API (that would report external URLs only) and some other mechanism that reports internal URLs ? IMO not having a single source of truth would make the configuration of OSAC harder/error-prone

jhernand:
> UI is a component of OSAC as any other. Why separate helm chart only for UI ?

We have separate charts for separate components:

- For the fulfillment service: https://github.com/osac-project/fulfillment-service/tree/main/charts/service
- For the operator: https://github.com/osac-project/osac-operator/tree/main/charts

Those are used and coordinated by the umbrella Helm charts in osac-installer:

https://github.com/osac-project/osac-installer/tree/main/charts/osac

As the UI has its separate UI repository, and evolves separately, it should also have its own Helm chart, in its own repository.

> Regarding the internal/external URL - when you deploy UI, the UI's proxy runs in-cluster and needs to talk to keycloak service. The frontend part runs in user's browser and needs the external keycloak URL. So a client (UI) needs both of these URLs. Should capabilities API be a single source of truth ? Or we want to split this config into Capabilities API (that would report external URLs only) and some other mechanism that reports internal URLs ? IMO not having a single source of truth would make the configuration of OSAC harder/error-prone

There should be only a Keycloak URL, not internal/external. That unique URL should be visible for internal and for external users. Keycloak can't be configured to have different "iss" claims depending on where you are coming from. Any client, including the UI, should use that unique URL, matching the "iss" included in the tokens. If you find an environment where the URL needs to be different for internal/external users that is a user configuration error that should be corrected: the user is probably passing the wrong value to the "issuerUrl" field of the chart.

rawagner:
> > UI is a component of OSAC as any other. Why separate helm chart only for UI ?
> 
> We have separate charts for separate components:
> 
> * For the fulfillment service: https://github.com/osac-project/fulfillment-service/tree/main/charts/service
> * For the operator: https://github.com/osac-project/osac-operator/tree/main/charts
> 
> Those are used and coordinated by the umbrella Helm charts in osac-installer:
> 
> https://github.com/osac-project/osac-installer/tree/main/charts/osac
> 
> As the UI has its separate UI repository, and evolves separately, it should also have its own Helm chart, in its own repository.
> 


Ok, that makes sense. Thank you, i will move the UI related components to its own chart.

> > Regarding the internal/external URL - when you deploy UI, the UI's proxy runs in-cluster and needs to talk to keycloak service. The frontend part runs in user's browser and needs the external keycloak URL. So a client (UI) needs both of these URLs. Should capabilities API be a single source of truth ? Or we want to split this config into Capabilities API (that would report external URLs only) and some other mechanism that reports internal URLs ? IMO not having a single source of truth would make the configuration of OSAC harder/error-prone
> 
> There should be only a Keycloak URL, not internal/external. That unique URL should be visible for internal and for external users. Keycloak can't be configured to have different "iss" claims depending on where you are coming from. Any client, including the UI, should use that unique URL, matching the "iss" included in the tokens. If you find an environment where the URL needs to be different for internal/external users that is a user configuration error that should be corrected: the user is probably passing the wrong value to the "issuerUrl" field of the chart.

Thanks for the pointers - the issue is indeed with the my keycloak instance misconfiguration.

PR #667, opened 2026-06-09
OSAC-953: Pass tenant and project name to get/delete requests

CrystalChun:
# Description

The tenant and project name are used to determine authorization.

# Testing

- [x] Ran integration test
    - Logged in as admin on CLI and created organization test-org, added user ben, assigned ben tenant-admin role
    - Logged in as ben on CLI and created project, verify able to see project
    - Logged in as charles on CLI and verify not able to see project
    - Deleted project successfully
   
/cc @jhernand

### DIFF
--- a/internal/auth/grpc_external_auth_interceptor.go
+++ b/internal/auth/grpc_external_auth_interceptor.go
@@ -32,15 +32,20 @@ import (
 	"google.golang.org/protobuf/proto"
 )
 
+// MetadataFetcher fetches object metadata (tenant and project name) for authorization.
+// Returns empty strings if object not found or on error.
+type MetadataFetcher func(ctx context.Context, id string) (tenant, project string)
+
 // GrpcExternalAuthType is the name of the external authentication type.
 const GrpcExternalAuthType = "external"
 
 // GrpcExternalAuthInterceptorBuilder contains the data and logic needed to build an interceptor that performs
 // authentication and authorization by calling an external service using the Envoy ext_authz gRPC protocol.
 type GrpcExternalAuthInterceptorBuilder struct {
-	logger        *slog.Logger
-	grpcClient    *grpc.ClientConn
-	publicMethods []string
+	logger          *slog.Logger
+	grpcClient      *grpc.ClientConn
+	publicMethods   []string
+	metadataFetcher MetadataFetcher
 }
 
 // GrpcExternalAuthInterceptor is an interceptor that performs authentication and authorization by calling an external
@@ -50,9 +55,10 @@ type GrpcExternalAuthInterceptorBuilder struct {
 // downstream handlers. The interceptor also has access to the request message, allowing it to include object-specific
 // details (like identifiers and tenants) in the authorization request.
 type GrpcExternalAuthInterceptor struct {
-	logger        *slog.Logger
-	authClient    envoyauthv3.AuthorizationClient
-	publicMethods []*regexp.Regexp
+	logger          *slog.Logger
+	authClient      envoyauthv3.AuthorizationClient
+	publicMethods   []*regexp.Regexp
+	metadataFetcher MetadataFetcher
 }
 
 // NewGrpcExternalAuthInterceptor creates a builder that can then be used to configure and create an authentication
@@ -86,6 +92,13 @@ func (b *GrpcExternalAuthInterceptorBuilder) AddPublicMethodRegex(value string) 
 	return b
 }
 
+// SetMetadataFetcher sets the function used to retrieve object metadata (tenant and project name) for authorization.
+// This is optional - if not set, metadata will not be fetched for authorization.
+func (b *GrpcExternalAuthInterceptorBuilder) SetMetadataFetcher(value MetadataFetcher) *GrpcExternalAuthInterceptorBuilder {
+	b.metadataFetcher = value
+	return b
+}
+
 // Build uses the data stored in the builder to create and configure a new interceptor.
 func (b *GrpcExternalAuthInterceptorBuilder) Build() (result *GrpcExternalAuthInterceptor, err error) {
 	// Check parameters:
@@ -112,9 +125,10 @@ func (b *GrpcExternalAuthInterceptorBuilder) Build() (result *GrpcExternalAuthIn
 
 	// Create and populate the object:
 	result = &GrpcExternalAuthInterceptor{
-		logger:        b.logger,
-		authClient:    authClient,
-		publicMethods: publicMethods,
+		logger:          b.logger,
+		authClient:      authClient,
+		publicMethods:   publicMethods,
+		metadataFetcher: b.metadataFetcher,
 	}
 	return
 }
@@ -269,6 +283,18 @@ func (i *GrpcExternalAuthInterceptor) buildCheckRequest(ctx context.Context, met
 		id := i.extractId(ctx, request)
 		if id != "" {
 			extensions["id"] = id
+		}
+
+		// For project Get/Delete/Update operations, fetch the authoritative tenant and name from the database
+		// to prevent clients from spoofing these values for authorization bypass.
+		if i.shouldFetchProjectMetadata(method, id) && i.metadataFetcher != nil {
+			tenant, name := i.metadataFetcher(ctx, id)
+			if tenant != "" {
+				extensions["tenant"] = tenant
+			}
+			if name != "" {
+				extensions["name"] = name
+			}
 		}
 	}
 
@@ -436,6 +462,19 @@ func (i *GrpcExternalAuthInterceptor) isPublicMethod(method string) bool {
 		}
 	}
 	return false
+}
+
+// shouldFetchProjectMetadata determines if we should fetch project metadata from the database for authorization.
+// This is needed for Get/Delete/Update operations to prevent clients from spoofing tenant/name values for
+// authorization bypass.
+func (i *GrpcExternalAuthInterceptor) shouldFetchProjectMetadata(method string, id string) bool {
+	if id == "" {
+		return false
+	}
+	// Fetch metadata for Projects Get, Delete, and Update operations
+	return method == "/osac.public.v1.Projects/Get" ||
+		method == "/osac.public.v1.Projects/Delete" ||
+		method == "/osac.public.v1.Projects/Update"
 }
 
 // grpcExternalAuthInternalError is the error returned an internal error is detected that pevents completing the
--- a/internal/auth/grpc_external_auth_interceptor_metadata_test.go
+++ b/internal/auth/grpc_external_auth_interceptor_metadata_test.go
@@ -1,0 +1,349 @@
+/*
+Copyright (c) 2025 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with the
+License. You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific
+language governing permissions and limitations under the License.
+*/
+
+package auth
+
+import (
+	"context"
+	"crypto/tls"
+	"crypto/x509"
+	"encoding/json"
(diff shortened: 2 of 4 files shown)


jhernand (changes requested):
I think we can't trust the caller to provide the correct tenant or name. What if they mliciously, or by accident, provide the wrong values?

Given that we are already extracting the `id` and sending it, I see two alternatives:

1. Once we now the `id` and the type of object we can fetch it from the database (with the `Get` method of the DAO) and then extract the tenant and name from there.

2. As we are sending the `id` already, the external authentication service could use the gRPC API to fetch the object, and extract any information it needs.

Option 1 seems more reasonable to me, as putting the logic to fetch the object via gRPC in Authorino would be more complicated and difficult to debug and change.

internal/auth/metadata/project_metadata_fetcher.go
```diff
@@ -0,0 +1,76 @@
...
+// GetProjectMetadata fetches tenant and name for a project by ID.
+// Returns empty strings if project not found or on error.
+// This bypasses authorization checks since it's used FOR authorization.
+func (f *DAOProjectMetadataFetcher) GetProjectMetadata(ctx context.Context, id string) (tenant, name string) {
+	// Query the database directly to avoid any authorization middleware
+	// that might expect a subject in context
+	ctx = database.TxManagerIntoContext(ctx, f.txMgr)
+
+	var projectTenant, projectName string
+	_, err := database.WithNewTx(ctx, func(txCtx context.Context) (struct{}, error) {
```
jhernand: Don't use this `WithNewTx` method. Both the transaction manager and the transaction have `Run` methods that you can use instead. We will probably remove this `WithNewTx` method.

internal/auth/metadata/project_metadata_fetcher.go
```diff
@@ -0,0 +1,76 @@
...
+		tx, err := database.TxFromContext(txCtx)
+		if err != nil {
+			return struct{}{}, err
+		}
+
+		// Direct SQL query to fetch tenant and name columns
+		query := `SELECT tenant, name FROM projects WHERE id = $1`
```
jhernand: I know that is common practice to use scream case for SQL, but in this project we (at least I) try to avoid it.

internal/auth/metadata/project_metadata_fetcher.go
```diff
@@ -0,0 +1,76 @@
...
+		logger: logger,
+		dao:    projectDAO,
+		txMgr:  txMgr,
+	}
+}
+
+// GetProjectMetadata fetches tenant and name for a project by ID.
+// Returns empty strings if project not found or on error.
+// This bypasses authorization checks since it's used FOR authorization.
+func (f *DAOProjectMetadataFetcher) GetProjectMetadata(ctx context.Context, id string) (tenant, name string) {
```
jhernand: Why project metadata? Isn't the *object* metadata what you need to check authorization? I mean that you need to know the name of the tenant and the name of the project of the object. The name of the tenant is in the `metadata.tenant` field of the object, and the name of the project is (well, will be) in the `metadata.project` field of the object. So shouldn't this be `MetadataFetcher` and the method just `Fetch`. Actually, as the contract is very simple, can you maybe use just a function type?

```go
type MetadataFetcher func (ctx context.Context, id string) (tenant, project string)
```

internal/auth/metadata/project_metadata_fetcher.go
```diff
@@ -0,0 +1,76 @@
...
+	return &DAOProjectMetadataFetcher{
+		logger: logger,
+		dao:    projectDAO,
+		txMgr:  txMgr,
+	}
+}
+
+// GetProjectMetadata fetches tenant and name for a project by ID.
+// Returns empty strings if project not found or on error.
+// This bypasses authorization checks since it's used FOR authorization.
```
jhernand: We should not bypass visibility checks for this, even if it is for authorization. Did you find a real problem if you use the same context/transaction?

Note that if we implement row level security in PostgreSQL this will most probably break, or make it impossible to implement.

internal/auth/metadata/project_metadata_fetcher.go
```diff
@@ -0,0 +1,76 @@
...
+	logger *slog.Logger
+	dao    *dao.GenericDAO[*privatev1.Project]
+	txMgr  database.TxManager
+}
+
+// Ensure DAOProjectMetadataFetcher implements auth.ProjectMetadataFetcher
+var _ auth.ProjectMetadataFetcher = (*DAOProjectMetadataFetcher)(nil)
+
+// NewDAOProjectMetadataFetcher creates a new DAOProjectMetadataFetcher.
+func NewDAOProjectMetadataFetcher(logger *slog.Logger, projectDAO *dao.GenericDAO[*privatev1.Project], txMgr database.TxManager) *DAOProjectMetadataFetcher {
```
jhernand: If you are going to have multiple parameters for the constructor, use a builder.

You shouldn't need to pass the transaction manager, you can get it from the context with `database.TxManagerFromContext`.

Can't this be in the `database` or `dao` packages?

internal/auth/grpc_external_auth_interceptor.go
```diff
@@ -270,6 +288,28 @@ func (i *GrpcExternalAuthInterceptor) buildCheckRequest(ctx context.Context, met
 		if id != "" {
 			extensions["id"] = id
 		}
+
+		// For project Get/Delete operations, fetch the authoritative tenant and name from the database
+		// to prevent clients from spoofing these values for authorization bypass.
+		// For Update, extract from the request object which contains full metadata.
```
jhernand: You can't trust the request object in the `Update` request, like you can't trust anything sent by the caller.

internal/auth/grpc_external_auth_interceptor_metadata_test.go
```diff
@@ -0,0 +1,197 @@
...
+	var ctx context.Context
+
+	BeforeEach(func() {
+		interceptor = &GrpcExternalAuthInterceptor{}
+		ctx = context.Background()
+	})
+
+	Describe("shouldFetchProjectMetadata", func() {
+		It("should return true for Projects Get with non-empty ID", func() {
+			result := interceptor.shouldFetchProjectMetadata("/osac.public.v1.Projects/Get", "some-id")
```
jhernand: This may be just a personal quirk, but I always try to avoid using in tests things that aren't part of the contract of the tested component. Like this `shouldFetchProjectMetadata` method, for example. Is there a way to test the behavior using only the contract of the component?

internal/cmd/service/start/grpcserver/start_grpc_server_cmd.go
```diff
@@ -232,6 +234,56 @@ func (c *runnerContext) run(cmd *cobra.Command, argv []string) error {
 	// Calculate the user agent:
 	userAgent := fmt.Sprintf("%s/%s", grpcServerUserAgent, version.Get())
 
+	// Create the tenancy logic (needed by auth interceptor for project metadata fetching):
```
jhernand: I now your IA did this comment, but it is confusing: the tenancy logic is not only needed by the auth interceptor for metadata fetching. I'd suggest to remove that part of the comment.

jhernand (changes requested):

internal/cmd/service/start/grpcserver/start_grpc_server_cmd.go
CrystalChun: Yeah 😅 thanks Juan, I've removed it.

internal/auth/grpc_external_auth_interceptor_metadata_test.go
CrystalChun: Gotcha, thank you for letting me know! I've attempted to update it it to test the behavior using the contract, how does it look now?

internal/auth/grpc_external_auth_interceptor.go
CrystalChun: Gotcha, thanks for pointing that out! I'm checking it for all 3 functions now, how does it look?

internal/auth/metadata/project_metadata_fetcher.go
CrystalChun: Good to know, thank you! I've updated this.

internal/auth/metadata/project_metadata_fetcher.go
CrystalChun: Sounds good, thanks for letting me know. I've updated the query to not scream

internal/auth/metadata/fetcher.go
```diff
@@ -0,0 +1,47 @@
...
+	"context"
+	"log/slog"
+
+	"github.com/osac-project/fulfillment-service/internal/auth"
+	"github.com/osac-project/fulfillment-service/internal/database"
+)
+
+// NewProjectFetcher creates a metadata fetcher for projects that queries the database directly.
+// This bypasses authorization checks since it's used FOR authorization.
+func NewProjectFetcher(logger *slog.Logger, txMgr database.TxManager) auth.MetadataFetcher {
```
jhernand: I still think that this should have a builder, and should be in the `dao` package.

And I don't quite get why you need to fetch the metadata of the project. Isn't the metadata of the object what you need?

internal/cmd/service/start/grpcserver/start_grpc_server_cmd.go
```diff
@@ -232,6 +233,44 @@ func (c *runnerContext) run(cmd *cobra.Command, argv []string) error { //nolint:
 	// Calculate the user agent:
 	userAgent := fmt.Sprintf("%s/%s", grpcServerUserAgent, version.Get())
 
+	c.logger.InfoContext(
```
jhernand, line 237: I meant remove the last part of the comment, not the complete comment. Granted it wasn't very useful either, but was visually pleasant, at least for me.

internal/cmd/service/start/grpcserver/start_grpc_server_cmd.go
```diff
@@ -232,6 +233,44 @@ func (c *runnerContext) run(cmd *cobra.Command, argv []string) error { //nolint:
...
+			return fmt.Errorf("failed to create guest tenancy logic: %w", err)
+		}
+	default:
+		return fmt.Errorf(
+			"unknown tenancy logic '%s', valid values are 'default' and 'guest'",
+			c.args.tenancyLogic,
+		)
+	}
+
+	// Prepare the transactions manager (needed by auth interceptor):
```
jhernand: I mean just `// Prepare the transaction manager:`.

internal/auth/metadata/fetcher.go
```diff
@@ -0,0 +1,47 @@
...
+	return func(ctx context.Context, id string) (tenant, project string) {
+		var projectTenant, projectName string
+
+		err := txMgr.Run(ctx, func(ctx context.Context, tx database.Tx) error {
```
jhernand: No need to explicitly pass the transaction manager, you can use the transaction from the context:

```go
tx, err := database.TxFromContext(ctx)
if err != nil {
        ...
}
```

And then you can use it to run the query.

Unless you need to run that query in separate transaction, you don't need to use the `txMgr.Run(...)` method. And if you needed, you can also get the transaction manager from the context:

```go
tm, err := database.TxManagerFromContext(ctx)
if err != nil {
        ...
}
```

internal/cmd/service/start/grpcserver/start_grpc_server_cmd.go
CrystalChun: Oh gotcha, I've put it back without the latter portion

internal/cmd/service/start/grpcserver/start_grpc_server_cmd.go
CrystalChun: Updated it, thank you!

internal/auth/metadata/fetcher.go
CrystalChun: Gotcha, thanks for letting me know this! I've updated it to use the `TxFromContext` and used it to query the row. Does that look ok?

internal/auth/metadata/fetcher.go
CrystalChun: Moved it to the `dao` package and yes right it's the object's metadata. Updated it! Thank you!

jhernand (approved):

PR #668, opened 2026-06-10
OSAC-1343:  Proto definitions and codegen for BareMetalInstance API

adriengentil:
## Summary

- Add 12 proto files (public + private) for \`BareMetalInstance\`, \`BareMetalInstanceTemplate\`, and \`BareMetalInstanceCatalogItem\`
- Public API: full CRUD services for all three resources; public \`BareMetalInstanceCatalogItem\` omits \`tenant\` field (server-managed)
- Private API: full CRUD + \`Signal\` RPC for all three resources; private \`BareMetalInstanceStatus\` includes \`hub\` field
- Run \`buf generate\` and update reflection helper test to include new types

## Jira

https://redhat.atlassian.net/browse/OSAC-1343

## Test plan

- [x] \`buf lint\` passes (pre-existing warnings in \`identity_provider_type.proto\` excluded)
- [x] \`buf generate\` idempotent
- [x] \`go build ./...\` passes
- [x] Unit tests pass (\`ginkgo run -r internal\`)


proto/public/osac/public/v1/baremetal_instance_catalog_item_type.proto
```diff
@@ -0,0 +1,47 @@
...
+  // Human friendly long description of the catalog item, using Markdown format.
+  string description = 4;
+
+  // Identifier of the underlying bare metal instance template that this catalog item references.
+  string template = 5;
+
+  // Whether this catalog item is published.
+  bool published = 6;
+
+  // Field 7 is omitted: `tenant` is an internal field managed by the server, not exposed through the public API.
```
adriengentil: to be removed

proto/public/osac/public/v1/baremetal_instances_service.proto
jhernand, line 1: Historically, and unfortunately, we haven't been documenting the private API. We should start to do that. Please document it like you documented the public API.

proto/public/osac/public/v1/baremetal_instance_type.proto
```diff
@@ -0,0 +1,124 @@
...
+  // Run strategy controls the power state of the bare metal instance.
+  // An enum is used rather than a boolean to leave room for future states (e.g. suspended backends).
+  optional BareMetalInstanceRunStrategy run_strategy = 4;
+
+  // RestartRequestedAt is a timestamp signal to request a power cycle.
+  //
+  // Set this field to the current time (usually NOW) to request a restart.
+  // The controller will execute the restart if this timestamp is greater than
+  // `status.last_restarted_at`.
+  optional google.protobuf.Timestamp restart_requested_at = 5;
```
jhernand: Message types like `google.protobuf.Timestamp` are always optional, so there is no need for the `optional` keyword.

proto/public/osac/public/v1/baremetal_instance_type.proto
```diff
@@ -0,0 +1,124 @@
...
+
+import "google/protobuf/timestamp.proto";
+import "osac/public/v1/condition_status_type.proto";
+import "osac/public/v1/metadata_type.proto";
+
+// Contains the details of a bare metal instance.
+//
+// The `spec` contains the desired details, and may be modified by the user. The `status` contains the current status
+// of the bare metal instance, is provided by the system and can't be modified by the user.
+message BareMetalInstance {
```
jhernand, line 26: I have to say that it is a bit confusing to have `ComputeInstance` and `BareMetalInstance`. Isn't a bare metal instance a compute instance? It is a rhetorical question: I know one is a virtual machine and the other is physical machine. I am just taking the opportunity to say that the naming choice isn't ideal. Consider renaming. `VirtualMachine` and `PhysicalMachine` would be better names, in my opinion. Or maybe use `ComputeInstance` for both, and differentiate using fields.

proto/public/osac/public/v1/baremetal_instance_type.proto
jhernand: I think I proposed this approach some time ago, but since then I though that it may be better to not require a date for this. Maybe it is better to use a string, or an integer:

```proto
// This is used to trigger a restart of the physical machine. Change it to any value
// to trigger the restart. The value isn't important, only the change is important.
string restart_trigger = 5;
```

The controller can use the `status.restart_trigger` to compare with this, and trigger the restart when they are different.

proto/public/osac/public/v1/baremetal_instance_template_type.proto
```diff
@@ -0,0 +1,41 @@
...
+
+syntax = "proto3";
+
+package osac.public.v1;
+
+import "osac/public/v1/metadata_type.proto";
+
+// A bare metal instance template defines a hardware profile (host type, OS image, network configuration)
+// that can be used to create bare metal instances via a catalog item.
+message BareMetalInstanceTemplate {
```
jhernand, line 22: What is the difference between this and the `InstanceType` recently introduced? Can't we use the same concept for both virtual and physical machines?

proto/public/osac/public/v1/baremetal_instance_templates_service.proto
```diff
@@ -0,0 +1,52 @@
...
+
+service BareMetalInstanceTemplates {
+  rpc List(BareMetalInstanceTemplatesListRequest) returns (BareMetalInstanceTemplatesListResponse) {
+    option (google.api.http) = {get: "/api/fulfillment/v1/baremetal_instance_templates"};
+  }
+  rpc Get(BareMetalInstanceTemplatesGetRequest) returns (BareMetalInstanceTemplatesGetResponse) {
+    option (google.api.http) = {
+      get: "/api/fulfillment/v1/baremetal_instance_templates/{id}"
+      response_body: "object"
+    };
```
jhernand, line 79: You need to define all the methods here: `Create`, `List`, `Get`, `Update` and `Delete`. Even if they aren't implemented. If you don't then the CLI will not recognize this as a service and will not be able to use it.

jhernand (changes requested):

proto/public/osac/public/v1/baremetal_instance_type.proto
adriengentil: We choose a path where objects are duplicated by components  (template for clusters, vm, and bm are the same, same issue for catalogitems, even the new `InstanceType` look to be dedicated to VMs). I'm fine to change it, but we need first to align on the architecture guidelines. 

We have a tight deadline for 0.1, so I would prefer to move forward, and have a solid plan for 0.2. We can bring this issue during our next group meeting.

proto/public/osac/public/v1/baremetal_instance_template_type.proto
adriengentil: The InstanceType defines the what kind of hardware we want to provision. The template defines what to provision on the hardware. 
We don't have yet the notion of InstanceType for 0.1, I'll come with 0.2.

proto/public/osac/public/v1/baremetal_instance_type.proto
adriengentil: Just used a int, and check if they are different

jhernand (approved):

proto/public/osac/public/v1/baremetal_instance_template_type.proto
ygalblum: I'm at the process of introducing `InstanceTypes` and it is supposed to be part of 0.1. Having said that, the `InstanceTypes` currently cover only `Cores` and `Memory`. But, I don't see any reference to such parameters anywhere in the BareMetal realm. Where does it come from? The Template Parameters?

proto/private/osac/private/v1/baremetal_instance_type.proto
```diff
@@ -0,0 +1,126 @@
...
+  optional string user_data = 3;
+
+  // Run strategy controls the power state of the bare metal instance.
+  // An enum is used rather than a boolean to leave room for future states (e.g. suspended backends).
+  optional BareMetalInstanceRunStrategy run_strategy = 4;
+
+  // Used to trigger a restart of the physical machine. Change it to any value to trigger the restart.
+  // The value itself is not important; only the change matters.
+  // The controller compares this with `status.restart_trigger` and initiates a power cycle when they differ.
+  int64 restart_trigger = 5;
```
carbonin, line 56: This seems like a very odd API. Even if this is how we implement it on the backend can we make it so that the user doesn't have to choose a random number and instead gets a dedicated "restart" endpoint?

proto/public/osac/public/v1/baremetal_instance_type.proto
carbonin: I'd prefer to keep the API dedicated to individual implementations rather than trying to share. Sharing a definition in the API is going to lead towards just a bucket of optional fields to make the generic object act like a vm vs bare metal and that will in turn require a whole matrix of what fields can be used together and which do different things depending on what the backend object is .... No, I'd rather keep them separate.

proto/private/osac/private/v1/baremetal_instance_type.proto
jhernand: I think it isn't that odd; is just a natural side effect of the declarative, intent-based approach of the API. It can be simplified by the CLI/UI, if we wish. I'd very much prefer to avoid dedicated endpoints.

PR #672, opened 2026-06-10
OSAC-204: Add health status to IDP

CrystalChun:
# Description

Allows an admin to query the health of an IDP.

# Testing

- [x] `go run github.com/bufbuild/buf/cmd/buf@latest lint` passes
- [x] Build passes `go build ./...`
- [x] All unit tests successfully passed 

/cc @jhernand 

## Summary by CodeRabbit

* **New Features**
  * Triggerable health checks for identity providers, health status states (healthy/unhealthy/unknown), test result tracking, and last-checked timestamps.
* **Chores**
  * Enhanced telemetry/logging for identity-provider health and test events with sensitive message content redacted and nil-safe handling.
* **Tests**
  * Added tests validating logging output, redaction behavior, and nil-safe logging paths.


proto/public/osac/public/v1/identity_providers_service.proto
```diff
@@ -135,6 +136,79 @@ message IdentityProvidersUnassignRequest {
...
+
+  // Filter events by user ID.
+  optional string user_id = 5;
+
+  // Filter events from this timestamp onwards.
+  optional google.protobuf.Timestamp from_time = 6;
+
+  // Filter events until this timestamp.
+  optional google.protobuf.Timestamp to_time = 7;
+}
```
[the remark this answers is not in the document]
CrystalChun: This will be addressed in a later patch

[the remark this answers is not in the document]
CrystalChun: No it's already tracked elsewhere

proto/private/osac/private/v1/identity_provider_type.proto
```diff
@@ -217,3 +217,87 @@ message IdentityProviderCondition {
...
+
+  // HEALTH_STATUS_HEALTHY indicates the component is healthy.
+  HEALTH_STATUS_HEALTHY = 1;
+
+  // HEALTH_STATUS_UNHEALTHY indicates the component is unhealthy.
+  HEALTH_STATUS_UNHEALTHY = 2;
+
+  // HEALTH_STATUS_UNKNOWN indicates the health status is unknown.
+  HEALTH_STATUS_UNKNOWN = 3;
+}
```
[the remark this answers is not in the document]
CrystalChun: done

proto/public/osac/public/v1/identity_providers_service.proto
```diff
@@ -203,4 +277,44 @@ service IdentityProviders {
...
+
+  // Organization-scoped IdP operations
+  // These operations are scoped to a specific organization (tenant).
+  // Access is restricted to Tenant Admins, idp-manager role, and Cloud Provider Admins (for System org).
+
+  // Retrieves the identity provider assigned to an organization.
+  // Returns the IdP configuration from Keycloak for the specified organization.
+  rpc GetOrganizationIdentityProvider(GetOrganizationIdentityProviderRequest) returns (GetOrganizationIdentityProviderResponse) {
+    option (google.api.http) = {
+      get: "/api/fulfillment/v1/organizations/{organization_name}/identity_provider"
```
jhernand: We shouldn't have nested collections in the REST API. Indentity providers are already associated to a tenant, like any object, via the `metadata.tenant` field. A regular tenant user will only see the identity providers of their organization. A cloud administrator can also filter by tenant, using something like `filter=this.metadata.tenant = "some-tenant"`. So I think we don't need these additional methods for listing or getting identity providers.

proto/public/osac/public/v1/identity_providers_service.proto
```diff
@@ -203,4 +277,44 @@ service IdentityProviders {
...
+      response_body: "object"
+    };
+  }
+
+  // Tests connectivity to the organization's identity provider.
+  // Proxies the test request through Keycloak to validate:
+  // - LDAP: Bind credentials and user search
+  // - OIDC: Client credentials and token endpoint
+  rpc TestOrganizationIdentityProvider(TestOrganizationIdentityProviderRequest) returns (TestOrganizationIdentityProviderResponse) {
```
jhernand: I think we should avoid this kind of imperative test operation. If we have them, then the `fulfillment-grpc-server` will need to connect to Keycloak, with the corresponding credentials. Currently we only do that from the `fulfillment-controller`. To preserve that separation I'd suggest to have in the `spec` of the identity provider a field that triggers the check:

```proto
// This field is used to trigger a heath check of the identity provider. The user
// can write anything here, the value isn't important, the only important thing
// is that when it changes the system will eventually initiate the check, and
// eventually report the result in the `status`.
string check_trigger = 6;
```

The `spec` of the identity provider should also have a similar field, for example:

```proto
// This is used by the system to know when the `spec.status` has changed, and
// then start the check process. It can also be used by the user to see if their
// request to perform the health check has been acknowledged. The system
// will update it to the value of `spec.status` when the check is started.
string check_trigger = 3;
```

The `status` should also contain other fields or conditions to report the result of the check.

For the user it will be very simple: the look at the `spec.check_trigger` field and set it to a different value. Then they wait for the changes in the `status` containing the results.

We can also use a more restrictive type, like an integer, and restrict it to be always incremented, but a random string is also OK.

proto/public/osac/public/v1/identity_providers_service.proto
```diff
@@ -203,4 +277,44 @@ service IdentityProviders {
...
+  // Checks connectivity, configuration validity, and operational status via Keycloak.
+  rpc GetOrganizationIdentityProviderHealth(GetOrganizationIdentityProviderHealthRequest) returns (GetOrganizationIdentityProviderHealthResponse) {
+    option (google.api.http) = {
+      get: "/api/fulfillment/v1/organizations/{organization_name}/identity_provider/health"
+    };
+  }
+
+  // Lists authentication and authorization events from the organization's identity provider.
+  // Retrieves events from Keycloak's event store for the specified organization.
+  rpc ListOrganizationIdentityProviderEvents(ListOrganizationIdentityProviderEventsRequest) returns (ListOrganizationIdentityProviderEventsResponse) {
```
jhernand: Do we really need this? I'd say it is not that important, and we can either do it later or not do it at all.

jhernand (changes requested):

internal/api/osac/private/v1/identity_provider_logging.go
```diff
@@ -0,0 +1,99 @@
...
+		attrs = append(attrs, slog.Bool("has_ip_address", true))
+	}
+
+	// Redact error details - may contain sensitive info
+	if e.Error != nil {
+		attrs = append(attrs, slog.Bool("has_error", true))
+	}
+
+	return slog.GroupValue(attrs...)
+}
```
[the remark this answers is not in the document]
CrystalChun: done

proto/private/osac/private/v1/identity_provider_type.proto
```diff
@@ -217,3 +235,39 @@ message IdentityProviderCondition {
   // LastTransitionTime is the last time the condition transitioned from one status to another.
   google.protobuf.Timestamp last_transition_time = 5;
 }
+
+// IdentityProviderTestResult contains the results of testing IdP connectivity.
+message IdentityProviderTestResult {
```
tzvatot: 🟡 **`IdentityProviderTestResult` is defined but unused**

This message is defined in both public and private protos but no RPC, no field in `IdentityProviderStatus` or `IdentityProviderHealth`, and no server code references it. It looks like dead code left over from the imperative `TestIdentityProvider` RPC that was removed per jhernand's feedback.

It also has `LogValue` implementations and tests written for it.

**Recommendation:** Remove from both protos and all hand-written code referencing it. Add it back in the PR that actually uses it.

proto/private/osac/private/v1/identity_provider_type.proto
```diff
@@ -217,3 +235,39 @@ message IdentityProviderCondition {
...
+// IdentityProviderHealth represents the health status of an identity provider.
+message IdentityProviderHealth {
+  // Status indicates the overall health status.
+  IdentityProviderHealthStatus status = 1;
+
+  // Message provides details about the health status.
+  optional string message = 2;
+
+  // LastChecked is when the health was last checked.
+  google.protobuf.Timestamp last_checked = 3;
```
tzvatot, line 244: 💡 **`last_checked` should be `optional`**

When a health check has never been performed, `last_checked` will default to the zero-value Timestamp (1970-01-01T00:00:00Z), which is semantically different from "never checked." The `message` field in the same message correctly uses `optional`, and `health` on the parent status also uses `optional`.

Per the project's proto conventions: "Use `optional` for fields that may not be set."

**Recommendation:** Change to `optional google.protobuf.Timestamp last_checked = 3;` in both public and private protos.

internal/api/osac/private/v1/identity_provider_logging.go
```diff
@@ -0,0 +1,60 @@
...
+	"log/slog"
+)
+
+// LogValue implements slog.LogValuer for IdentityProviderTestResult to redact error details.
+// Redacts: detailed error messages that may contain credentials or connection strings.
+func (r *IdentityProviderTestResult) LogValue() slog.Value {
+	if r == nil {
+		return slog.Value{}
+	}
+
```
tzvatot: 💡 **Missing logging test for private API types**

The public API has both `identity_provider_logging.go` and `identity_provider_logging_test.go`, but the private API has no corresponding test file.

**Recommendation:** Add `identity_provider_logging_test.go` for the private package mirroring the public test file.

tzvatot (changes requested):

proto/private/osac/private/v1/identity_provider_type.proto
CrystalChun: Thank you for letting me know! Modified it to be optional.

proto/private/osac/private/v1/identity_provider_type.proto
CrystalChun: Thank you for catching this! Removed the unused code.

internal/api/osac/private/v1/identity_provider_logging.go
CrystalChun: Thanks for catching this! Added an associated test file for the private API.

proto/private/osac/private/v1/identity_provider_type.proto
jhernand: Note that composite types (like `google.protobuf.Timestamp`) are always optional: putting the `optional` keyword in front of them doesn't change anything, just adds extra token for the compiler to process. Not a big deal. It does make a difference for primitive types, like `string`, `bool`, etc: the generated code will use a pointer, or a similar construct, to be able to check if it is present or not.

CrystalChun: I didn't realize that, thank you for letting me know! Let me remove this.

jhernand: No need to remove it, it makes no difference. Just wanted to make sure you know that, for other cases.

internal/api/osac/private/v1/identity_provider_logging.go
jhernand, line 1: These files aren't generated, but they are in the `internal/api` directory, which is wiped out (or should be) whenever the Go code is regenerated. Do you really need these files? If so, then should be in a different place.

jhernand (changes requested):

internal/api/osac/private/v1/identity_provider_logging.go
CrystalChun: Gotcha, removed these files!

jhernand (approved):

PR #687, opened 2026-06-12
OSAC-365: canonicalize CIDR values across networking resources

tchughesiv:
## Canonicalize CIDR values across OSAC networking resources

**Jira:** https://redhat.atlassian.net/browse/OSAC-365

Normalize CIDR strings to masked canonical form (`netip.ParsePrefix` + `Masked()`), e.g. `10.0.1.5/24` → `10.0.1.0/24`.

### Scope

| Resource | Fields | Canonicalized | Notes |
|----------|--------|---------------|-------|
| VirtualNetwork | `ipv4_cidr`, `ipv6_cidr` | Create; Update | Prefix immutable ([#402](https://github.com/osac-project/fulfillment-service/pull/402)); equivalent notation OK; legacy heal on omit-path Update |
| Subnet | `ipv4_cidr`, `ipv6_cidr` | Create; Update | Same as VN ([#402](https://github.com/osac-project/fulfillment-service/pull/402)) |
| SecurityGroup rules | rule `ipv4_cidr` / `ipv6_cidr` | Create; Update | Fully mutable |
| PublicIPPool | `spec.cidrs[]` | Create only | Immutable after create (string-exact); no legacy heal on Update |
| Cluster | `pod_cidr`, `service_cidr` | Create; Update | No network CIDR immutability |

Shared helpers: `internal/servers/cidr_validation.go`, `internal/utils/cidr.go`. Cluster validation consolidated in `ValidateClusterSpecFields`.

### Operator reprovision (read before merge)

`DesiredConfigVersion` is a **string-exact** JSON hash of spec. Rewriting `10.0.1.5/24` → `10.0.1.0/24` (even with the same prefix) can reprovision READY VN/Subnet/Cluster.

- **VN/Subnet/SG (Netris):** AAP playbooks are name-idempotent — reprovision should be a no-op there; expect Progressing churn, not Netris misconfig.
- **Cluster:** higher risk (`HostedCluster` k8s apply); cluster CIDR immutability follow-up: [OSAC-1619](https://redhat.atlassian.net/browse/OSAC-1619).
- **PublicIPPool:** Create-only canonicalization; existing pools unchanged until recreated.

**Before prod:** staging smoke for legacy VN/Subnet omit-path heal; audit non-canonical CIDRs in staging/prod. Optional follow-up: semantic CIDR hashing in operator.

### Testing

- Unit + integration CI **pass** on this PR (`internal/servers` + `internal/utils`, Helm/Kustomize `it/`)
- New coverage: VN/Subnet/SG Create+Update, PublicIPPool Create, cluster Update field-mask round-trip

### Rollout

- [ ] Staging smoke: legacy VN/Subnet canonical heal → operator returns READY
- [ ] Pre-flight audit for non-canonical CIDRs in staging/prod

## Summary by CodeRabbit

* **New Features**
  * Network CIDR addresses are now automatically normalized across subnets, virtual networks, security groups, clusters, and IP pools.

* **Bug Fixes**
  * Fixed CIDR immutability enforcement with proper value preservation during updates.

* **Tests**
  * Added comprehensive CIDR canonicalization test coverage.

### DIFF
--- a/internal/servers/cidr_validation.go
+++ b/internal/servers/cidr_validation.go
@@ -1,0 +1,132 @@
+/*
+Copyright (c) 2026 Red Hat Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with the
+License. You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific
+language governing permissions and limitations under the License.
+*/
+
+package servers
+
+import (
+	"net/netip"
+
+	grpccodes "google.golang.org/grpc/codes"
+	grpcstatus "google.golang.org/grpc/status"
+)
+
+const (
+	cidrIPv4 = "IPv4"
+	cidrIPv6 = "IPv6"
+)
+
+// parseAndValidateCIDR parses cidrStr, validates it matches the expected IP version,
+// and returns the canonical CIDR notation.
+func parseAndValidateCIDR(cidrStr string, ipVersion string) (string, error) {
+	prefix, err := netip.ParsePrefix(cidrStr)
+	if err != nil {
+		return "", grpcstatus.Errorf(grpccodes.InvalidArgument,
+			"invalid %s CIDR format '%s': %v", ipVersion, cidrStr, err)
+	}
+
+	isIPv4 := prefix.Addr().Is4()
+	if ipVersion == cidrIPv4 && !isIPv4 {
+		return "", grpcstatus.Errorf(grpccodes.InvalidArgument,
+			"field 'ipv4_cidr' contains IPv6 address: %s", cidrStr)
+	}
+	if ipVersion == cidrIPv6 && isIPv4 {
+		return "", grpcstatus.Errorf(grpccodes.InvalidArgument,
+			"field 'ipv6_cidr' contains IPv4 address: %s", cidrStr)
+	}
+
+	return prefix.Masked().String(), nil
+}
+
+// cidrPrefixesEqual reports whether two CIDR strings denote the same network prefix.
+func cidrPrefixesEqual(a, b string) (bool, error) {
+	if a == b {
+		return true, nil
+	}
+	prefixA, err := netip.ParsePrefix(a)
+	if err != nil {
+		return false, err
+	}
+	prefixB, err := netip.ParsePrefix(b)
+	if err != nil {
+		return false, err
+	}
+	return prefixA.Masked() == prefixB.Masked(), nil
+}
+
+// validateImmutableCIDR rejects updates that change the network prefix; canonically equivalent
+// notation (e.g. 10.0.1.5/24 vs 10.0.1.0/24) is allowed.
+func validateImmutableCIDR(fieldName, existingCIDR, newCIDR, ipVersion string) error {
+	equal, err := cidrPrefixesEqual(existingCIDR, newCIDR)
+	if err == nil && equal {
+		return nil
+	}
+	from, to := existingCIDR, newCIDR
+	if existingCIDR != "" {
+		if canonical, err := parseAndValidateCIDR(existingCIDR, ipVersion); err == nil {
+			from = canonical
+		}
+	}
+	if canonical, err := parseAndValidateCIDR(newCIDR, ipVersion); err == nil {
+		to = canonical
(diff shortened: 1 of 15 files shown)


internal/servers/cidr_validation.go
```diff
@@ -0,0 +1,49 @@
...
+// and returns the canonical CIDR notation.
+func parseAndValidateCIDR(cidrStr string, ipVersion string) (string, error) {
+	_, network, err := net.ParseCIDR(cidrStr)
+	if err != nil {
+		return "", grpcstatus.Errorf(grpccodes.InvalidArgument,
+			"invalid %s CIDR format '%s': %v", ipVersion, cidrStr, err)
+	}
+
+	isIPv4 := network.IP.To4() != nil
+	if ipVersion == "IPv4" && !isIPv4 {
```
ygalblum: Should `IPv4` and `IPv6` defined as constants somewhere? Shouldn't break on a typo

internal/servers/cidr_validation.go
```diff
@@ -0,0 +1,49 @@
...
+	if ipVersion == "IPv6" && isIPv4 {
+		return "", grpcstatus.Errorf(grpccodes.InvalidArgument,
+			"field 'ipv6_cidr' contains IPv4 address: %s", cidrStr)
+	}
+
+	return network.String(), nil
+}
+
+// validateCIDR validates a CIDR string and checks if it matches the expected IP version.
+func validateCIDR(cidrStr string, ipVersion string) error {
```
ygalblum: Is this method used anywhere? I couldn't find it in this PR, but maybe somewhere else

internal/servers/private_subnets_server.go
```diff
@@ -228,16 +228,24 @@ func (s *PrivateSubnetsServer) validateSubnet(ctx context.Context,
 
 	// SUB-VAL-01: Validate IPv4 CIDR format
 	if spec.GetIpv4Cidr() != "" {
-		if err := validateCIDR(spec.GetIpv4Cidr(), "IPv4"); err != nil {
+		canonical, err := parseAndValidateCIDR(spec.GetIpv4Cidr(), "IPv4")
+		if err != nil {
 			return err
 		}
+		if existingSubnet == nil {
```
ygalblum: I guess this is done to facilitate further validation. But, it's not clear why and why it depends on existingSubnet being `nil` can you add a comment about it?

internal/servers/cidr_validation.go
tchughesiv: Good catch... I’ll add constants in `cidr_validation.go` and use them at call sites in this PR.

internal/servers/cidr_validation.go
tchughesiv: Yes... still used by SG rule validation:

https://github.com/osac-project/fulfillment-service/blob/8a26dc2567468b87e9718ffd01b0357b1670eaa9/internal/servers/private_security_groups_server.go#L344-L354

VN/Subnet use `parseAndValidateCIDR` to canonicalize on Create; SG keeps `validateCIDR` (validate only) per OSAC-365 scope:

https://github.com/osac-project/fulfillment-service/blob/8a26dc2567468b87e9718ffd01b0357b1670eaa9/internal/servers/cidr_validation.go#L45-L49

If you want SG (or `PublicIPPool` `spec.cidrs[]`) canonicalized here too, say the word... otherwise I’d track those as follow-ups.

internal/servers/cidr_validation.go
tchughesiv: added `cidrIPv4/cidrIPv6` constants in `cidr_validation.go`.

internal/servers/private_subnets_server.go
tchughesiv: added comments on both paths.

`existingVN == nil` / `existingSubnet == nil` distinguishes `Create` from `Update`. We canonicalize only on Create today because:
- OSAC-365 scope is new Create behavior, not a data migration.
- Legacy rows may already store non-canonical strings; normalizing on Update would rewrite stored values on unrelated updates.
- Immutability is string-based - a client sending the canonical form against a legacy stored value would fail even though the network is the same.

We can do Update canonicalization, but it’s a product decision about migrating existing data... we’d want agreed Update/migration rules, tests for legacy rows, and a quick check that anything downstream wouldn’t break if the stored CIDR text changed but the network didn’t.

internal/servers/cidr_validation.go
ygalblum: All good. I assumed it's used somewhere else. But, wanted to double check

ygalblum (approved):
/lgtm
/approve

tchughesiv:
@coderabbitai full review

danmanor (approved):

PR #691, opened 2026-06-14
OSAC-1400: validate field definitions at catalog item creation time

tzvatot:
## [OSAC-1400](https://redhat.atlassian.net/browse/OSAC-1400): Validate catalog item field definitions at creation time

**Jira:** https://redhat.atlassian.net/browse/OSAC-1400

### Summary

Non-editable field definitions without a default value are now rejected at catalog item creation time with `InvalidArgument`, instead of silently accepting them and failing with an `Internal` error when a user tries to provision from the catalog item.

### Changes

- Added `validateFieldDefinitions()` in `catalog_item_validation.go` that checks all field definitions for the invalid combination (`editable=false`, no default)
- Wired validation into `Create()` for both `PrivateClusterCatalogItemsServer` and `PrivateComputeInstanceCatalogItemsServer`
- Fixed existing tests that created non-editable field definitions without defaults

### Testing

- 6 new tests (3 per catalog item type): reject without default, accept with default, accept editable without default
- 2 existing tests updated to include default values
- Full suite: 916 server tests pass, 0 regressions

### Acceptance Criteria

- [x] Reject catalog item creation if a non-editable field definition has no default value
- [x] Return `InvalidArgument` with a clear error message
- [x] Apply to both ClusterCatalogItems and ComputeInstanceCatalogItems
- [x] Unit tests covering the validation

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened `field_definitions` validation for catalog items during both create and update.
  * Non-editable fields must include a default value; missing defaults are rejected, including when provided via update masks.
  * Invalid `validation_schema` JSON is now rejected with clear `InvalidArgument` errors; valid JSON is accepted.
  * Applies consistently to both cluster and compute instance catalog items.
* **Tests**
  * Expanded coverage to verify numeric default round-trips and to validate create/update rejection/acceptance cases for defaults, update masks, and `validation_schema`.

### DIFF
--- a/internal/servers/private_cluster_catalog_items_server_test.go
+++ b/internal/servers/private_cluster_catalog_items_server_test.go
@@ -522,7 +522,14 @@ var _ = Describe("Private cluster catalog items server", func() {
 				}.Build(),
 			}.Build())
 			Expect(err).ToNot(HaveOccurred())
-			Expect(response.GetObject()).ToNot(BeNil())
+			object := response.GetObject()
+			Expect(object).ToNot(BeNil())
+			DeferCleanup(func() {
+				_, err := server.Delete(ctx, privatev1.ClusterCatalogItemsDeleteRequest_builder{
+					Id: object.GetId(),
+				}.Build())
+				Expect(err).ToNot(HaveOccurred())
+			})
 		})
 
 		It("Accepts editable field definition without default value", func() {
@@ -539,7 +546,14 @@ var _ = Describe("Private cluster catalog items server", func() {
 				}.Build(),
 			}.Build())
 			Expect(err).ToNot(HaveOccurred())
-			Expect(response.GetObject()).ToNot(BeNil())
+			object := response.GetObject()
+			Expect(object).ToNot(BeNil())
+			DeferCleanup(func() {
+				_, err := server.Delete(ctx, privatev1.ClusterCatalogItemsDeleteRequest_builder{
+					Id: object.GetId(),
+				}.Build())
+				Expect(err).ToNot(HaveOccurred())
+			})
 		})
 
 		It("Rejects non-editable field definition without default when not first in list", func() {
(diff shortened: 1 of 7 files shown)


internal/servers/private_cluster_catalog_items_server_test.go
```diff
@@ -291,6 +292,7 @@ var _ = Describe("Private cluster catalog items server", func() {
 							Path:        "spec.node_sets.workers.size",
 							DisplayName: "Worker count",
 							Editable:    false,
+							Default:     structpb.NewNumberValue(3),
```
mennyaboush, line 295: If you changed the fixture, its better to assert the change.

mennyaboush (commented):
Nice work, LGTM. Two non-blocking suggestions:

The round-trip tests were updated to add Default values to the fixtures — worth asserting those defaults survive persistence after Get() (e.g. Expect(fd1.GetDefault().GetNumberValue()).To(Equal(3.0))). If you changed the fixture, it's better to assert the change.

Could add a test for the three-way combo (editable=true + default + validationSchema) to guard against accidental condition coupling in validateFieldDefinitions().

Not blockers - the PR is safe to merge as-is.

internal/servers/private_cluster_catalog_items_server_test.go
tzvatot: Good point. Added assertions for the Default values in the round-trip tests - both cluster (`3.0`) and compute instance (`16.0`). Also added `DeferCleanup` to the validation tests that were missing cleanup for successfully created objects.

adriengentil:
@tzvatot This task is part of a "global" change for catalog items? Or it's focused on CaaS/VMaaS only? because I think we would need the same change for BMaaS to stay consistent

tzvatot:
> @tzvatot This task is part of a "global" change for catalog items? Or it's focused on CaaS/VMaaS only? because I think we would need the same change for BMaaS to stay consistent

@adriengentil The Jira ticket scoped this to CaaS and VMaaS (ClusterCatalogItems and ComputeInstanceCatalogItems). BMaaS wasn't included since it was added more recently, but you're right - the validation should apply there too for consistency. The logic is in a shared `validateFieldDefinitions()` function in `catalog_item_validation.go`, so wiring it into `PrivateBareMetalInstanceCatalogItemsServer.Create/Update` is straightforward. I'll add it to this PR.

PR #699, opened 2026-06-15
OSAC-1449: Add fabric_manager and k8s_manager fields to NetworkClass proto

danmanor:
## `OSAC-1449`: Add fabric_manager and k8s_manager fields to NetworkClass proto

**Jira:** https://redhat.atlassian.net/browse/OSAC-1449
**Epic:** `OSAC-1438` — NetworkClass Two-Manager Model

### Summary

Adds `fabric_manager` (string) and `k8s_manager` (optional string) fields to the NetworkClass
message in both public and private API proto definitions. These fields identify which managers
handle physical fabric operations and K8s-to-fabric bridging for each network class, as described
in the unified-networking enhancement proposal.

### Changes

- **`proto/private/osac/private/v1/network_class_type.proto`** — added `fabric_manager` (field 10) and `k8s_manager` (field 11) to private `NetworkClass` message
- **`proto/public/osac/public/v1/network_class_type.proto`** — added matching fields to public `NetworkClass` message
- **`internal/api/`** — regenerated Go code via `buf generate`

### Testing

- **Unit tests:** no new tests — this is an additive proto schema change with no business logic. All 910 existing tests in `internal/servers` pass.
- **Integration tests:** N/A
- **Validation:** `buf lint`, `uv run dev.py lint`, `ginkgo run -r internal` all pass

### Acceptance Criteria

- [x] Add `fabric_manager` (string) field to NetworkClass proto in both public and private APIs
- [x] Add `k8s_manager` (optional string) field to NetworkClass proto in both public and private APIs
- [x] Keep all existing fields unchanged
- [x] Generated Go code compiles successfully after `buf generate`
- [x] `buf lint` passes


proto/private/osac/private/v1/network_class_type.proto
```diff
@@ -65,6 +65,16 @@ message NetworkClass {
...
+
+  // Identifier of the fabric manager that handles physical network operations for this class. The fabric manager
+  // controls all fabric-level operations: virtual network segments, security groups, external IPs, and NAT gateways.
+  // For example: "netris", "neutron".
+  string fabric_manager = 10;
+
+  // Identifier of the K8s manager that bridges the Kubernetes overlay to the physical fabric. Only needed for regions
+  // that host VMs — the K8s manager creates the overlay network on hosting clusters and bridges it to the fabric
+  // segment. For example: "cudn_localnet". If not set, the region does not support VM workloads.
+  optional string k8s_manager = 11;
```
jhernand, line 77: What is the user of the API going to do with these fields? I think they don't belong in the public API. Consider having them only in the private API.

danmanor: You are right. Thank you @jhernand  !

jhernand (approved):

proto/private/osac/private/v1/network_class_type.proto
danmanor: Fixed

PR #700, opened 2026-06-15
NO-ISSUE: Add Helm build artifacts to .gitignore

eliorerz:
## Summary
- Add `charts/*/charts/` and `charts/*/Chart.lock` to `.gitignore`
- These are generated by `helm dependency build` and should not be tracked

## Summary by CodeRabbit

* **Chores**
  * Updated build configuration to ignore Helm dependency artifacts.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,7 @@
 /osac-dev
 __debug_bin*
 __pycache__
+
+# Helm dependency build artifacts
+charts/*/charts/
+charts/*/Chart.lock


.gitignore
```diff
@@ -15,3 +15,7 @@
 /osac-dev
 __debug_bin*
 __pycache__
+
+# Helm dependency build artifacts
+charts/*/charts/
+charts/*/Chart.lock
```
jhernand, line 21: I think this isn't needed because the only chart in this repository does not embed other dependent charts, and we don't have plan to change that.

Anyhow, it doesn't hurt. I'd appreciate if you can remove the comment and sort the lines ad keep the lines alphabetically sorted. It helps to avoid duplicates.

PR #701, opened 2026-06-15
OSAC-240: Add describe networkclass CLI command and update 

SiddarthR56:
## Summary
- Add describe networkclass CLI subcommand with tabwriter output showing all NetworkClass fields (ID, name, title, description, capabilities, default, state, message)
- Update get networkclass table YAML to include description and per-capability (IPv4/IPv6/Dual-Stack) columns alongside existing ID, name, title, and default columns
- Register networkclass in the describe command tree with alias networkclasses and add corresponding test coverage (8 rendering tests + parent registration tests)

## Summary by CodeRabbit

* **New Features**
  * Added a new CLI command to describe network classes by ID or name.
  * Enhanced network class output to include additional details such as description and IP capability indicators (IPv4, IPv6, dual stack).

* **Tests**
  * Expanded CLI test coverage for the new network class subcommand and its rendering behavior.

* **Documentation**
  * Added command help text with examples for describing network classes.

### DIFF
--- a/internal/cmd/cli/describe/networkclass/describe_networkclass_cmd.go
+++ b/internal/cmd/cli/describe/networkclass/describe_networkclass_cmd.go
@@ -1,0 +1,166 @@
+/*
+Copyright (c) 2025 Red Hat Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with the
+License. You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific
+language governing permissions and limitations under the License.
+*/
+
+package networkclass
+
+import (
+	"fmt"
+	"io"
+	"strings"
+	"text/tabwriter"
+
+	"github.com/spf13/cobra"
+	"google.golang.org/protobuf/proto"
+
+	publicv1 "github.com/osac-project/fulfillment-service/internal/api/osac/public/v1"
+	"github.com/osac-project/fulfillment-service/internal/cmd/cli/lookup"
+	"github.com/osac-project/fulfillment-service/internal/config"
+	"github.com/osac-project/fulfillment-service/internal/terminal"
+)
+
+func Cmd() *cobra.Command {
+	runner := &runnerContext{}
+	result := &cobra.Command{
+		Use:                   "networkclass [FLAG...] ID|NAME",
+		Aliases:               []string{"networkclasses"},
+		Short:                 shortHelp,
+		Long:                  longHelp,
+		DisableFlagsInUseLine: true,
+		Args:                  cobra.ExactArgs(1),
+		RunE:                  runner.run,
+	}
+	return result
+}
+
+type runnerContext struct {
+	console *terminal.Console
+}
+
+func (c *runnerContext) run(cmd *cobra.Command, args []string) error {
+	ref := args[0]
+
+	ctx := cmd.Context()
+
+	c.console = terminal.ConsoleFromContext(ctx)
+
+	cfg := config.SettingsFromContext(ctx)
+	if !cfg.Armed() {
+		return fmt.Errorf("there is no configuration, run the 'login' command")
+	}
+
+	conn, err := cfg.Connect(ctx, cmd.Flags())
+	if err != nil {
+		return fmt.Errorf("failed to create gRPC connection: %w", err)
+	}
+	defer conn.Close()
+
+	client := publicv1.NewNetworkClassesClient(conn)
+
+	matched, err := lookup.Find(ref, "network class", func(filter string, limit int32) ([]*publicv1.NetworkClass, error) {
+		resp, err := client.List(ctx, publicv1.NetworkClassesListRequest_builder{
+			Filter: proto.String(filter),
+			Limit:  proto.Int32(limit),
+		}.Build())
+		if err != nil {
+			return nil, fmt.Errorf("failed to describe network class: %w", err)
+		}
+		return resp.GetItems(), nil
+	})
+	if err != nil {
+		return err
+	}
+
+	renderNetworkClass(c.console, matched)
+
+	return nil
+}
+
+func boolYesNo(v bool) string {
+	if v {
+		return "yes"
+	}
+	return "no"
+}
+
+// renderNetworkClass writes a formatted description of nc to w.
+func renderNetworkClass(w io.Writer, nc *publicv1.NetworkClass) {
+	writer := tabwriter.NewWriter(w, 0, 0, 2, ' ', 0)
+
+	name := "-"
+	if v := nc.GetMetadata().GetName(); v != "" {
+		name = v
+	}
+
+	title := "-"
+	if v := nc.GetTitle(); v != "" {
+		title = v
+	}
+
+	description := "-"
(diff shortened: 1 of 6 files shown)


internal/cmd/cli/describe/networkclass/describe_networkclass_cmd.go
```diff
@@ -0,0 +1,177 @@
...
+
+	conn, err := cfg.Connect(ctx, cmd.Flags())
+	if err != nil {
+		return fmt.Errorf("failed to create gRPC connection: %w", err)
+	}
+	defer conn.Close()
+
+	client := publicv1.NewNetworkClassesClient(conn)
+
+	matched, err := lookup.Find(ref, "network class", lookup.FindOptions{IncludeDeleted: c.includeDeleted}, func(filter string, limit int32) ([]*publicv1.NetworkClass, error) {
```
jhernand: I know we mostly don't care about readability of code, but this is awfully long and difficult to read. Consider making it a bit better.

internal/cmd/cli/describe/networkclass/describe_networkclass_cmd.go
```diff
@@ -0,0 +1,170 @@
...
+		}
+		return resp.GetItems(), nil
+	})
+	if err != nil {
+		return err
+	}
+
+	RenderNetworkClass(c.console, matched)
+
+	return nil
```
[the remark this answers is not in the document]
SiddarthR56: All Render* functions in this codebase follow the same convention of not checking fmt.Fprintf/Flush errors. This is intentional — these write to a terminal console where failures are non-actionable. Ignoring write errors in CLI output is idiomatic Go. No refactor is needed.

internal/cmd/cli/describe/networkclass/describe_networkclass_cmd.go
```diff
@@ -0,0 +1,170 @@
...
+	client := publicv1.NewNetworkClassesClient(conn)
+
+	matched, err := lookup.Find(ref, "network class", func(filter string, limit int32) ([]*publicv1.NetworkClass, error) {
+		resp, err := client.List(ctx, publicv1.NetworkClassesListRequest_builder{
+			Filter: proto.String(filter),
+			Limit:  proto.Int32(limit),
+		}.Build())
+		if err != nil {
+			return nil, fmt.Errorf("failed to describe network class: %w", err)
+		}
```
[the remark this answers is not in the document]
SiddarthR56: No other `describe` command adds a per-call timeout; they all pass `ctx` directly to `client.List`. The client configures keepalives (which can detect dead connections), but the CLI doesn't enforce per-RPC deadlines anywhere today. If we want per-RPC timeouts, that's a cross-cutting concern that should be handled consistently (e.g., via a CLI-wide `--timeout` flag), not introduced in a single `describe` command. Worth tracking as a future enhancement, but out of scope for this PR.

internal/cmd/cli/describe/networkclass/describe_networkclass_cmd.go
```diff
@@ -0,0 +1,170 @@
...
+	cfg := config.SettingsFromContext(ctx)
+	if !cfg.Armed() {
+		return fmt.Errorf("there is no configuration, run the 'login' command")
+	}
+
+	conn, err := cfg.Connect(ctx, cmd.Flags())
+	if err != nil {
+		return fmt.Errorf("failed to create gRPC connection: %w", err)
+	}
+	defer conn.Close()
```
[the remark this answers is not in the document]
SiddarthR56: The connection is just being marked closed and the process is about to exit; there's nothing meaningful to do with the error (no retry, no exit code change, no user action). This is consistent with how the Go ecosystem treats close on read-only/client resources (e.g. defer f.Close() on read-only file handles).

internal/cmd/cli/describe/networkclass/describe_networkclass_cmd.go
```diff
@@ -0,0 +1,170 @@
...
+	matched, err := lookup.Find(ref, "network class", func(filter string, limit int32) ([]*publicv1.NetworkClass, error) {
+		resp, err := client.List(ctx, publicv1.NetworkClassesListRequest_builder{
+			Filter: proto.String(filter),
+			Limit:  proto.Int32(limit),
+		}.Build())
+		if err != nil {
+			return nil, fmt.Errorf("failed to describe network class: %w", err)
+		}
+		return resp.GetItems(), nil
+	})
```
danmanor, line 78: I might be wrong here, but looking at `lookup.Find` on main — its signature is `Find[T any](ref, kind string, opts FindOptions, list ListFunc[T])`, so it expects `FindOptions` as the third argument. Every other describe command passes `lookup.FindOptions{IncludeDeleted: c.includeDeleted}` there.

This call is passing the closure directly as the third arg, which I would expect to fail to compile. Did you test this against the current main? Maybe the signature changed on your branch — just want to make sure this builds.

ori-amizur (commented):
Minor issues — none are correctness bugs, all consistency nits.

internal/cmd/cli/describe/networkclass/describe_networkclass_cmd.go
```diff
@@ -0,0 +1,170 @@
...
+	return result
+}
+
+type runnerContext struct {
+	logger  *slog.Logger
+	console *terminal.Console
+}
+
+func (c *runnerContext) run(cmd *cobra.Command, args []string) error {
+	ref := args[0]
```
ori-amizur, line 50: Unused field: `logger` is assigned on line 62 but never read. This also pulls in unnecessary imports (`log/slog` and `internal/logging`). The existing `describe cluster` command (`describe_cluster.go`) only has `console` in its `runnerContext`. Suggest removing the field and both imports.

internal/cmd/cli/describe/networkclass/describe_networkclass_cmd.go
```diff
@@ -0,0 +1,170 @@
...
+
+	return nil
+}
+
+func boolYesNo(v bool) string {
+	if v {
+		return "yes"
+	}
+	return "no"
+}
```
ori-amizur, line 93: Nit: this is exported (`RenderNetworkClass`) but the equivalent in the cluster package is unexported (`renderCluster`). Since the tests are in the same package and can access unexported symbols, consider renaming to `renderNetworkClass` for consistency.

internal/cmd/cli/describe/networkclass/describe_networkclass_cmd.go
```diff
@@ -0,0 +1,170 @@
+/*
```
ori-amizur, line 1: Nit: file is named `describe_networkclass_cmd.go` but the existing pattern uses no `_cmd` suffix (e.g. `describe_cluster.go`, `describe_computeinstance.go`). Consider renaming to `describe_networkclass.go` for consistency.

SiddarthR56: describe_cluster.go is actually the outlier here — every other describe command uses the _cmd suffix (describe_subnet_cmd.go, describe_computeinstance_cmd.go, describe_virtualnetwork_cmd.go, describe_publicip_cmd.go, describe_publicipattachment_cmd.go, describe_securitygroup_cmd.go, describe_baremetalinstance_cmd.go). So the current name is consistent with the dominant pattern. Leaving as-is.

internal/cmd/cli/describe/networkclass/describe_networkclass_cmd.go
SiddarthR56: Verified — the current signature on this branch is Find[T any](ref, kind string, list ListFunc[T]), with no FindOptions parameter

PR #708, opened 2026-06-16
OSAC-1546: Add tenancy tests for explicit tenant handling

jhernand:
## Summary

This increases the coverage of the tenancy tests to make sure that when a user of the public API
specifies a tenant explicitly it is honoured during both creation and update operations. The new
tests verify three scenarios:

- When a user creates an object with an explicit tenant that differs from the tenancy logic's
  default, the explicitly specified tenant is stored rather than being overridden by the default.
- When an object is updated without mentioning the tenant in the update mask, the original tenant
  is preserved. The test intentionally configures a default tenant that differs from the one used
  during creation, so it can distinguish between genuine preservation and accidental re-defaulting.
- When a user explicitly sets a different tenant during an update (via `metadata.tenant` in the
  update mask), the new tenant is persisted to the database.

Related: https://redhat.atlassian.net/browse/OSAC-1546
Related: https://github.com/osac-project/fulfillment-service/pull/704

## Test plan

- [x] New tests pass: `ginkgo run --focus="Tenancy logic" -v internal/servers`
- [x] All existing tenancy tests continue to pass (8/8)
- [x] Pre-commit hooks pass (including `golangci-lint`)

## Summary by CodeRabbit

* **Tests**
  * Enhanced test coverage for tenant-related operations, including test cases for tenant selection precedence, tenant preservation during updates, and explicit tenant specification with verification.

### DIFF
--- a/internal/servers/servers_tenancy_test.go
+++ b/internal/servers/servers_tenancy_test.go
@@ -19,6 +19,7 @@ import (
 	"go.uber.org/mock/gomock"
 	grpccodes "google.golang.org/grpc/codes"
 	grpcstatus "google.golang.org/grpc/status"
+	"google.golang.org/protobuf/types/known/fieldmaskpb"
 
 	privatev1 "github.com/osac-project/fulfillment-service/internal/api/osac/private/v1"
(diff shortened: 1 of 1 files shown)


CrystalChun (approved):
/lgtm

PR #711, opened 2026-06-16
OSAC-1546: Harden the hubs table

jhernand:
## Summary

- Enforce that hubs always belong to the `shared` tenant and that their name equals their
  identifier, both in the `PrivateHubsServer.Create` method and via a database migration that
  backfills existing data and switches the primary key to `(tenant, name)`.
- Add an immutable-columns trigger to prevent changes to the `id`, `name` and `tenant` columns
  after creation, and a unique constraint on `id` so DAO lookups remain efficient.
- Update the schema checker to support hardened tables, the CLI `create hub` command to use a
  `--name` flag (deprecating `--id`), and the integration test helper to match the new API.

## Test plan

- [x] Unit tests pass (`ginkgo run -r internal` -- 961 specs, 0 failures)
- [x] Migration tests pass (`ginkgo run internal/database/migrations` -- 87 specs, 0 failures)
- [x] Schema consistency test passes (`ginkgo run --focus="Is consistent" internal/database`)
- [x] Integration tests pass (`ginkgo run it`)

Related: https://redhat.atlassian.net/browse/OSAC-1546

## Release Notes

* **Refactor**
  * Hub CLI command now requires the `--name` flag as the primary identifier; the `--id` flag is deprecated
  * Added validation for hub creation to ensure required fields and prevent duplicate names

* **Chores**
  * Database schema hardened with enhanced integrity controls to prevent modification of hub identifiers after creation

### DIFF
--- a/internal/cmd/cli/create/hub/create_hub_cmd.go
+++ b/internal/cmd/cli/create/hub/create_hub_cmd.go
@@ -37,11 +37,18 @@ func Cmd() *cobra.Command {
 	}
 	flags := result.Flags()
 	flags.StringVar(
-		&runner.id,
+		&runner.name,
+		"name",
+		"",
+		nameFlagHelp,
+	)
+	flags.StringVar(
+		&runner.name,
 		"id",
 		"",
 		idFlagHelp,
 	)
+	flags.MarkDeprecated("id", "use '--name' instead")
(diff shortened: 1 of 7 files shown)


CrystalChun (approved):
/lgtm

PR #713, opened 2026-06-16
OSAC-717: add guest_os_family field to ComputeInstance

ygalblum:
Add guest_os_family as an optional proto field to ComputeInstanceSpec (public and private APIs). The field supports "linux" and "windows" values, is inferred from image.source_ref when not explicitly set, and is immutable after creation.

Changes:
- Add guest_os_family to compute_instance_type.proto (public + private)
- Add InferGuestOSFamily and ValidateGuestOSFamily utility functions
- Enforce immutability in PrivateComputeInstancesServer.Update()
- Unit tests for utilities and server-side validation

## Release Notes

* **New Features**
  * Compute instances now support a `guest_os_family` setting that is automatically inferred from the image when not provided.
  * The system preserves an explicitly set value (e.g., `linux`) and infers `windows` for matching Windows image sources.
  * `guest_os_family` is now immutable after creation; updates that try to change it are rejected.
  * Added support for listing/filtering by `guest_os_family`.
* **Bug Fixes**
  * Improved filter behavior for protobuf string fields so missing values are treated consistently as empty strings.
* **Tests**
  * Expanded coverage for create/get/update/list scenarios and filtering by `guest_os_family`.


proto/private/osac/private/v1/compute_instance_type.proto
```diff
@@ -93,6 +93,10 @@ message ComputeInstanceSpec {
   // Reference to a compute instance catalog item. Mutually exclusive with template during the migration period.
   // When set, the server fetches the catalog item and applies its field definitions.
   string catalog_item = 15;
+
+  // Guest OS family (e.g. "linux", "windows").
+  // Inferred from image.source_ref if not set. Immutable after creation.
+  optional string guest_os_family = 16;
```
jhernand, line 99: This documentation of the private API should be identical to the documentation of the public API.

In the future we want to generate the public API from the private API, and having differences means that we will need to manually reconcile them or else we will lose it.

proto/public/osac/public/v1/compute_instance_type.proto
```diff
@@ -162,6 +162,17 @@ message ComputeInstanceSpec {
...
+  // Guest OS family for this compute instance (e.g. "linux", "windows").
+  //
+  // If not explicitly set, the fulfillment service will attempt to infer the value
+  // from `image.source_ref` (e.g., images containing "windows" will be detected).
+  //
+  // If inference does not detect a Windows image, the field remains empty, and
+  // downstream consumers (osac-operator, osac-aap) default to "linux".
+  //
+  // This field cannot be modified after the compute instance is created.
+  optional string guest_os_family = 16;
```
jhernand, line 175: How will the user of the API know what are the valid values that they can put here? Can we define an enum type containing and documenting the values?

This is a spec field, and it shouldn't be modified by the system. How will the result of the inference be communicated to the user? Will we have an equivalent field in the status?

internal/servers/private_compute_instances_server.go
```diff
@@ -226,6 +226,12 @@ func (s *PrivateComputeInstancesServer) Create(ctx context.Context,
 		}
 	}
 
+	// Infer guest_os_family from image path if not explicitly set:
+	utils.InferGuestOSFamily(request.GetObject().GetSpec())
```
jhernand, line 230: I think this doesn't belong here. It should be in the controller, and reported to the user via a status field. The API server should not touch the spec. The calculation is currently very simple, but it will probably grow more complicated, and that kind of complicated logic doesn't belong into the API server, in general.

internal/utils/guest_os.go
```diff
@@ -0,0 +1,86 @@
...
+}
+
+// ValidateGuestOSFamily performs soft validation on the guest_os_family value.
+// It accepts empty, "linux", and "windows" without logging. For unknown values,
+// it logs a warning but does not return an error.
+//
+// Per design decisions D-13, D-14:
+// - Soft validation logs a warning via slog when an unknown value is received
+// - No error returned to client
+// - Warning goes to server log only (no response annotation or client-visible warning)
```
jhernand, line 73: Feels a bit useless. What is the system admin reviewing the logs going to do with this?

proto/public/osac/public/v1/compute_instance_type.proto
```diff
@@ -162,6 +162,17 @@ message ComputeInstanceSpec {
...
   // When set, the server fetches the catalog item and applies its field definitions.
   string catalog_item = 15;
+
+  // Guest OS family for this compute instance (e.g. "linux", "windows").
+  //
+  // If not explicitly set, the fulfillment service will attempt to infer the value
+  // from `image.source_ref` (e.g., images containing "windows" will be detected).
+  //
+  // If inference does not detect a Windows image, the field remains empty, and
+  // downstream consumers (osac-operator, osac-aap) default to "linux".
```
jhernand, line 172: osac-operator and osac-aap are invisible to the user of the API. There should be no references to that in the documentation.

proto/public/osac/public/v1/compute_instance_type.proto
```diff
@@ -162,6 +162,17 @@ message ComputeInstanceSpec {
   // Reference to a compute instance catalog item. Mutually exclusive with template during the migration period.
   // When set, the server fetches the catalog item and applies its field definitions.
   string catalog_item = 15;
+
+  // Guest OS family for this compute instance (e.g. "linux", "windows").
+  //
+  // If not explicitly set, the fulfillment service will attempt to infer the value
```
jhernand, line 168: I'd suggest "the system" rather than "the fulfillment service". The user of the API doesn't need to know which component is doing this.

internal/utils/guest_os.go
jhernand, line 1: Can we avoid increasing the number of things in "util" packages? I'd suggest you rename this it to `GuestOsCalculator`, or something similar, make it a component, and put it in the `controllers/computeinstance` package.

internal/utils/guest_os.go
```diff
@@ -0,0 +1,86 @@
...
+func ValidateGuestOSFamily(ctx context.Context, logger *slog.Logger, value string) {
```
jhernand, line 74: Don't pass loggers as parameters. If you need a logger you should probably be making this a component, and passing the logger to the constructor. Alternatively you can get the logger from the context. See https://github.com/osac-project/fulfillment-service/blob/1520791a56326bf6db924a60de2b5467056f4b44/internal/logging/logging_context.go#L29-L35.

internal/utils/guest_os.go
```diff
@@ -0,0 +1,86 @@
...
+	// Accept empty, "linux", and "windows" without logging
+	if value == "" || value == "linux" || value == "windows" {
```
jhernand, line 76: These should, at the very least, be constants. Or even better if they are values of a documented enum type in the API.

jhernand (changes requested):

ygalblum:
@jhernand I'm answering in a separate comment since your comments are linked in a way that answering each one separately might not mean much.

- This change goes all over the repos. The AAP and Controller parts are already in and I'd rather not change them. So, I want to keep the inference in the fulfillment service.
- This change will be changed again in the next version with the introduction of `ComputeImages` as the field will be remove from the `ComputeInstance` and encapsulated in the former. 
- In the first stage the `ComputeImage` will only handle the metadata and won't have any representation in the controller. so, I want to have it infer the O/S in the fulfillment service anyhow.
- To support other values in teh future without having to change the API all down the stack, I left the field as free text. Note that currently, only `windows` has any actual effect. But, maybe I should use an `enum` at the fulfillment service API level. 
- Regarding inference, I also didn't want to update the `spec` as it should represent the user's request. Also, I didn't want to set any value for unknown images. So, who do you envision the status field in such a case?

Once we align here, we'll see which comments are still relevant.

jhernand:
> * This change goes all over the repos. The AAP and Controller parts are already in and I'd rather not change them. So, I want to keep the inference in the fulfillment service.

You can keep the changes on the other repos and still do what I am suggesting: calculate the value in [computeinstance_reconciler_function.go](https://github.com/osac-project/fulfillment-service/blob/main/internal/controllers/computeinstance/computeinstance_reconciler_function.go) and store the result in `status.guest_os_family`, leaving `spec.guest_os_family` untouched.

> * This change will be changed again in the next version with the introduction of `ComputeImages` as the field will be remove from the `ComputeInstance` and encapsulated in the former.

The same will apply to that new type.

> * In the first stage the `ComputeImage` will only handle the metadata and won't have any representation in the controller. so, I want to have it infer the O/S in the fulfillment service anyhow.

When say _controller_ I mean [computeinstance_reconciler_function.go](https://github.com/osac-project/fulfillment-service/blob/main/internal/controllers/computeinstance/computeinstance_reconciler_function.go). I am not talking about the Kubernetes controller.

> * To support other values in teh future without having to change the API all down the stack, I left the field as free text. Note that currently, only `windows` has any actual effect. But, maybe I should use an `enum` at the fulfillment service API level.

Users of the API need a way to know what values are allowed, and what they mean. If it is impossible for them to use an incorrect value even better. An enum gives you that, and a natural place to put documentation for the type and for each specific value. This doesn't need to propagate all the way down the stack, if you don't want. You can translate that enum to a string in the controller. However, I'd say it would be better to have type safety all down the stack.

> * Regarding inference, I also didn't want to update the `spec` as it should represent the user's request. Also, I didn't want to set any value for unknown images. So, who do you envision the status field in such a case?

For unknown values you can either make the field `optional` in the status, and not set it if the value is unknown, or else add a `UNKNOWN` value to the enum. I'd say the latter is better, as it is more explicit.

ygalblum:
Closing for #734

PR #714, opened 2026-06-17
OSAC-455: persist hub selection before CR creation to prevent orphaning

mennyaboush:
## Summary

- Adds a shared `HubPersistenceHelper` that persists hub selection to the database **before** creating Kubernetes CRs, preventing orphaned CRs if the process crashes between CR creation and database persistence.
- Applied to all affected reconcilers: Cluster, ComputeInstance, Subnet, VirtualNetwork, PublicIP, PublicIPPool.
- SecurityGroup is unaffected — it derives the hub from its parent VirtualNetwork every reconciliation (no `status.hub` field).
- 20 new unit tests across all reconcilers covering: persistence before CR, no CR on failure, idempotency, and crash recovery.

## Why this approach

Considered and rejected:
- **Search all hubs for orphaned CRs** — O(N) API calls per reconcile, unreachable hubs cause false negatives and duplicate CRs anyway.
- **Hub ID in CR labels for recovery** — still needs the all-hubs search, doesn't prevent orphaning.
- **Two-phase commit with `hub_state` field** — requires proto schema change, overkill for a single extra DB write.

Tradeoff: one additional DB round-trip per new resource (~10-50ms), negligible compared to provisioning time.

## Test plan

- [x] All 20 new [OSAC-455](https://redhat.atlassian.net/browse/OSAC-455) tests pass (4 per reconciler × 5 reconcilers)
- [x] All 134 existing reconciler tests pass (no regressions)
- [x] `go build ./internal/controllers/...` compiles cleanly
- [ ] Integration tests (`ginkgo run it`)

Fixes: [OSAC-455](https://issues.redhat.com/browse/OSAC-455)

## Summary by CodeRabbit

* **New Features**
  * Hub selection is now persisted to the database before creating Kubernetes resources for clusters, compute instances, subnets, virtual networks, public IPs, and public IP pools, improving recovery after interruptions and preventing duplicate resource creation.
* **Improvements**
  * Hub persistence and hub selection are now conditional, using targeted `status.hub` updates.
  * Reconciliation initializes missing status safely and reports `PROGRESSING=false` with `ResourcesUnavailable` when hub selection isn’t possible.
* **Tests**
  * Added hub-persistence coverage (success, failure, idempotency, and crash recovery) across the affected reconcilers.


internal/controllers/hub_persistence.go
```diff
@@ -0,0 +1,96 @@
...
+// SelectAndPersistHub selects a hub and immediately persists it to the database.
+// Must be called BEFORE creating the Kubernetes CR.
+func (h *HubPersistenceHelper) SelectAndPersistHub(
+	ctx context.Context,
+	resourceID string,
+	resourceType string,
+	getCurrentHub func() string,
+	selectHub func(context.Context) (string, error),
+	setHub func(string),
+	persistHub func(context.Context) error,
```
jhernand: Don't define functions with such large list of parameters, should be two, or three, not more. If you need more parameters then you should define a builder for this helper, and use methods in the hub:

```go
helper, err := NewHelper().
        SetLogger(logger).
        SetObjectId("123").
        SetObjectType(...).
        SetCurrentHubFunc(...).
        SetSelectHubFunc(...).
        ...
        Build()
if err != nil {
        ...
}
```

internal/controllers/hub_persistence.go
```diff
@@ -0,0 +1,96 @@
...
+	setHub(hubID)
+
+	err = persistHub(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to persist hub selection for %s %s: %w", resourceType, resourceID, err)
+	}
+
+	h.logger.InfoContext(
+		ctx,
+		"Persisted hub selection before CR creation",
```
jhernand: Why should this say that it is before CR creation? That is not part of the logic of this component. If you want to say that you are doing something before CR creation it should be in the component that does that, not here.

Also what is a "CR"? Try to avoid acronyms.

internal/controllers/hub_persistence.go
```diff
@@ -0,0 +1,96 @@
...
+		logger: logger,
+	}
+}
+
+// SelectAndPersistHub selects a hub and immediately persists it to the database.
+// Must be called BEFORE creating the Kubernetes CR.
+func (h *HubPersistenceHelper) SelectAndPersistHub(
+	ctx context.Context,
+	resourceID string,
+	resourceType string,
```
jhernand: We don't use the "resource" concept in the project, rather "object", so these should be `objectId` and `objectType`.

Also consider using generics to avoid having to indicate the type as a string.

internal/controllers/hub_persistence.go
jhernand: Is the hub stored as `status.hub` in all the relevant objects? If it is then you could define define a interface private to this type that contains the operations that you need:

```go
type statusWithHub interface {
        HasHub() bool
        GetHub() string
        SetHub(string)
}
```

Then you could check if the object spec implements this interface, cast to it, and use it directly instead of having these "selectHub", "setHub" and "persistHub" functions.

If you combine that with a generic type, and gRPC reflection, you can check that in the Build method and refuse to create the helper if the type doesn't implement the interface.

jhernand (changes requested):

internal/controllers/hub_persistence.go
mennyaboush: Good point, I'll refactor to a builder pattern to match the project conventions.

internal/controllers/hub_persistence.go
mennyaboush: I'll rename to objectId/objectType. For generics - I'll look into whether we can use the type parameter to derive the name instead of passing a string.

internal/controllers/hub_persistence.go
```diff
@@ -0,0 +1,139 @@
...
+}
+
+// Build validates configuration and creates a HubPersistenceHelper.
+func (b *HubPersistenceHelperBuilder) Build() (*HubPersistenceHelper, error) {
+	if b.logger == nil {
+		return nil, errors.New("logger is mandatory")
+	}
+	if b.status == nil {
+		return nil, errors.New("status is mandatory")
+	}
```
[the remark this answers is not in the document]
mennyaboush: he status is guaranteed non-nil by the HasStatus() init before the builder is called. Adding reflect-based nil checking would over-engineer for a scenario the calling code prevents.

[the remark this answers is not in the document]
mennyaboush: The HasStatus() initialization at the top of run() guarantees status is always a real pointer before it reaches the builder:


if !cluster.HasStatus() {
    cluster.SetStatus(&privatev1.ClusterStatus{})
}
// ... status is guaranteed non-nil here
SetStatus(cluster.GetStatus())
So the typed-nil scenario can't happen in the current code. Instead of adding a reflect-based check in Build(), I added a test that creates a cluster with no status and verifies run() doesn't panic. If someone removes the HasStatus() guard in the future, this test catches it.

internal/controllers/hub_persistence.go
```diff
@@ -0,0 +1,139 @@
...
+}
+
+// SetObjectId sets the identifier of the object being reconciled (for logging).
+func (b *HubPersistenceHelperBuilder) SetObjectId(value string) *HubPersistenceHelperBuilder {
+	b.objectId = value
+	return b
+}
+
+// SetStatus sets the status object that carries the hub field. This is mandatory.
+func (b *HubPersistenceHelperBuilder) SetStatus(value statusWithHub) *HubPersistenceHelperBuilder {
```
jhernand: I'd say this can be a parameter to the "Run" method, so that you can reuse the helper for multiple objects. That would be two parameters in that "Run" method, which is acceptable.

jhernand: It feels a bit strange that you use a private interface in the public contract of this component. I'd suggest you either make this interface public, or else use `any` (or `proto.Message`) and check the type internally.

mennyaboush: Good point. I've exported the interface as StatusWithHub so it's part of the public contract. Also moved objectId from the builder to Run(ctx, objectId) as you suggested.

Regarding reuse -  the helper is still built per-object because SetStatus, SetSelectHub, and SetPersistHub all capture the specific object via closures (e.g., cluster.GetStatus(), r.clustersClient.Update()). These change per reconcile call, so the builder can't be shared across objects. Moving objectId to Run() makes the API cleaner but doesn't enable reuse for the same reason.

If you'd like the builder to be reusable, we'd need to also move status, selectHub, and persistHub to Run() - but that would give Run() five parameters, which goes against the "two or three, not more" guideline. Happy to discuss if you see a better way to structure this.

jhernand (approved):

PR #718, opened 2026-06-17
OSAC-1352: sync BareMetalInstance CR status back to fulfillment API

carbonin:
## [OSAC-1352](https://redhat.atlassian.net/browse/OSAC-1352): sync BareMetalInstance CR status back to fulfillment API

**Jira:** [OSAC-1352](https://issues.redhat.com/browse/OSAC-1352)

### Summary

The BareMetalInstance reconciler only pushed spec downward (proto → K8s CR) but never read the CR's status back. After the BMF operator updates phase and conditions on the CR, the fulfillment-service's internal record stayed at `PROVISIONING` regardless of actual state. This completes the feedback loop started by the osac-operator's BareMetalInstance feedback controller ([OSAC-1352](https://redhat.atlassian.net/browse/OSAC-1352)).

### Changes

- **`syncStatus()` method** — reads the K8s BareMetalInstance CR status and maps it to the proto representation:
  - Phase → State: `Allocating`/`Progressing` → `PROVISIONING`, `Ready` → `RUNNING`, `Failed` → `FAILED`, `Deleting` → `DELETING`
  - Conditions: `Allocated` → `PROVISIONED`, `ProvisionTemplateComplete` → `CONFIGURATION_APPLIED`, `Available` → `READY`
  - `PowerSynced` condition: reason-dependent mapping to `RESTART_IN_PROGRESS` or `RESTART_FAILED`
  - `PowerSynced=True`: clears restart conditions, syncs `restart_trigger` completion
- **`mapConditionStatus()` helper** — translates `metav1.ConditionStatus` to proto `ConditionStatus`
- **`update()` integration** — `syncStatus(object)` is called after the create-or-update block, before `ensureUserDataSecret`

### Testing

- **Unit tests:** 18 new tests covering all phase→state mappings, all condition type mappings, PowerSynced reason-dependent logic (Progressing, IronicAPIFailure, True, unhandled reason), restart trigger sync, multi-condition sync, nil/empty edge cases, and `mapConditionStatus` with all three inputs
- **Live verification:** Tested end-to-end on a running cluster by patching a BareMetalInstance CR's status with conditions and confirming propagation to the gRPC API via `grpcurl`

### Verified on cluster

```
# Before: all conditions FALSE with empty reasons
# After patching CR status with Allocated=True, ProvisionTemplateComplete=True, Available=True:
PROVISIONED → CONDITION_STATUS_TRUE (reason: HostAllocated)
CONFIGURATION_APPLIED → CONDITION_STATUS_TRUE (reason: Complete)
READY → CONDITION_STATUS_TRUE (reason: HostAvailable)
```

Related to https://github.com/osac-project/osac-operator/pull/301 which calls signal when the `BareMetalInstance` changes.

### DIFF
--- a/internal/controllers/baremetalinstance/baremetalinstance_reconciler_function.go
+++ b/internal/controllers/baremetalinstance/baremetalinstance_reconciler_function.go
@@ -218,6 +218,8 @@ func (t *task) update(ctx context.Context) error {
 		)
 	}
 
+	t.syncStatus(object)
+
 	if err := t.ensureUserDataSecret(ctx, object); err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


internal/controllers/baremetalinstance/baremetalinstance_reconciler_function.go
```diff
@@ -218,6 +218,8 @@ func (t *task) update(ctx context.Context) error {
 		)
 	}
 
+	t.syncStatus(object)
```
jhernand, line 221: If you do this here the status will only be synchronized when there is some change triggered by the user of the API of the fulfillment service. It won't be updated if the Kubernetes controllers changes something. Isn't the Kubernetes controller doing this already? If it isn't then I believe it should, and then this shouldn't be needed.

carbonin: The controller just pokes the fulfillment service using signal then this code runs to do the actual field translation. https://github.com/osac-project/osac-operator/pull/301

adriengentil (approved):

internal/controllers/baremetalinstance/baremetalinstance_reconciler_function.go
jhernand: Ah, OK, that looks good.

PR #724, opened 2026-06-17
OSAC-1419: add script for seeding default catalog items

vladikr:
Add a new `osac-dev seed-catalog-items` subcommand that seeds a small set of default catalog items into the fulfillment service. 
This unblocks the UI team by providing ready-to-use catalog items for development and testing.

Default catalog items created:
- Cluster: Simple OpenShift 4.17 cluster (osac.templates.ocp_4_17_small)
- Cluster: OpenShift 4.20 on NICo bare metal (osac.templates.ocp_4_20_small_nico)
- VM: Linux virtual machine (osac.templates.ocp_virt_vm)
- VM: Windows virtual machine (osac.templates.ocp_virt_vm)

The script is idempotent and can be run multiple times safely.

Usage:

```bash
# Normal run (creates items if they don't exist)
go run ./cmd/osac-dev seed-catalog-items \
  --api-url=fulfillment-api.osac.svc.cluster.local:443 \
  --token=$(kubectl create token -n osac client)

# Dry run
go run ./cmd/osac-dev seed-catalog-items \
  --api-url=... --token=... --dry-run --insecure
```

## Summary

* **New Features**
  * Added new catalog item YAML examples for Linux and Windows VMs, plus OpenShift cluster configurations with editable, schema-validated fields.
  * Enhanced the `seed-catalog-items` CLI command with a `--dry-run` option to preview what would be created.

* **Documentation**
  * Documented how to create and seed catalog item YAML/protobuf examples, including required fields and authentication/token setup.

* **Tests**
  * Added integration tests covering catalog item creation, published visibility behavior (for clusters), and field-definition default/validation-schema handling.

### DIFF
--- a/examples/catalog-items/linux-vm.yaml
+++ b/examples/catalog-items/linux-vm.yaml
@@ -1,0 +1,40 @@
+# Linux Virtual Machine catalog item
+# A general-purpose Linux virtual machine with customizable resources.
+
+"@type": type.googleapis.com/osac.private.v1.ComputeInstanceCatalogItem
+metadata:
+  name: linux-vm
+title: Linux Virtual Machine
+description: A general-purpose Linux virtual machine with customizable resources.
+template: osac.templates.ocp_virt_vm
+published: true
+field_definitions:
+  - path: instance_type
+    display_name: Instance Type
+    editable: true
+    default: cx1.2xlarge
+    validation_schema: '{"type":"string","minLength":1}'
+  - path: boot_disk.size_gib
+    display_name: Boot Disk Size (GiB)
+    editable: true
+    default: 120
+    validation_schema: '{"type":"integer","minimum":10,"maximum":1024}'
+  - path: image.source_ref
+    display_name: Container Disk Image
+    editable: true
+    default: quay.io/containerdisks/fedora:latest
+    validation_schema: '{"type":"string","pattern":"^[a-z0-9./-]+:[a-z0-9._-]+$"}'
+  - path: image.source_type
+    display_name: Image Source Type
+    editable: false
+    default: registry
+    validation_schema: '{"type":"string"}'
+  - path: run_strategy
+    display_name: Run Strategy
+    editable: true
+    default: Always
+    validation_schema: '{"type":"string","enum":["Always","Halted"]}'
+  - path: user_data
+    display_name: Cloud Init User Data
+    editable: true
+    validation_schema: '{"type":"string"}'
--- a/examples/catalog-items/README.md
+++ b/examples/catalog-items/README.md
@@ -1,0 +1,70 @@
+# Catalog Item Examples
+
+This directory contains example catalog items that can be created in the OSAC fulfillment service. These YAML files can be used directly with `osac create -f` for one-off creation, or as reference for the `osac-dev seed-catalog-items` idempotent seeding command.
+
+## Contents
+
+- `simple-ocp-4-17-cluster.yaml` — Simple OpenShift 4.17 cluster (fc430 hardware)
+- `ocp-4-20-nico-baremetal-cluster.yaml` — OpenShift 4.20 cluster on NICo bare metal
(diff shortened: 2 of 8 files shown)


hack/seed-catalog-items.sh
jhernand, line 1: Don't add "hacks" or scripts, please. You can add a new sub-command to the existing "osac-dev" binary, written in Go. You can use this as an example: https://github.com/osac-project/fulfillment-service/tree/main/internal/cmd/osac-dev/generate.

jhernand (changes requested):

hack/seed-catalog-items.sh
vladikr: Thanks for the feedback @jhernand 
The original request was to quickly unblock the UI team for 0.1 with some hardcoded defaults. A shell script was the fastest way to do that.
I'm happy to implement this properly as a osac-dev subcommand, but that will take more time and would need to be reworked for 0.2.

jhernand: If you don't have the time then I'd suggest you just create an "examples/catalog-items" directory and put there YAML files, and a README explaining how to use "osac create -f ..." to add them to the system.

jhernand (approved):

vladikr:
I removed the "Published compute instance catalog items appear in public API" test.

The test was failing with PermissionDenied on the public.v1.ComputeInstanceCatalogItems/List endpoint. I think the issue is a missing OPA authorization policy for public compute instance catalog item listing.
I think the policy doesn't exist.
That's unrelated to catalog item seeding, so I've removed the test from this PR.

batzionb:
Creating vms from catalogitems with network_attachments from both UI and CLI will always fail with 
'network_attachments.0.subnet' is required but no value was provided and no default is defined"
That's because verifying arrays is not supported
I suggest to remove from the field_definitions the network_attachments
The UI is currently being fixed to show all spec fields, so these can be removed 
More details here https://redhat.atlassian.net/browse/OSAC-1713

examples/catalog-items/linux-vm.yaml
```diff
@@ -0,0 +1,53 @@
...
+description: A general-purpose Linux virtual machine with customizable CPU, memory, and disk resources.
+template: osac.templates.ocp_virt_vm
+published: true
+field_definitions:
+  - path: cores
+    display_name: CPU Cores
+    editable: true
+    default: 8
+    validation_schema: '{"type":"integer","minimum":1,"maximum":64}'
+  - path: memory_gib
```
batzionb: VM creation was recently changed to use instancetypes instead of cores and memory
https://github.com/osac-project/fulfillment-service/pull/735

batzionb:
The VM catalog items (`linux-vm.yaml`, `windows-vm.yaml`) include an `instance_type` field with a default of `cx1.2xlarge`, but seeding only creates catalog items — it does not create any `InstanceType` resources. On a fresh cluster, `osac get instancetypes` is empty, so these catalog items look fine until someone tries to provision a VM and hits a missing instance type (wizard/API validation failure).

We should pick one of these before merge:

1. **Drop instance type from the seeded VM catalog items for now** — remove the `instance_type` field definition (or omit the VM examples from default seeding) until we have a supported seed path for instance types. Cluster-only seeding stays self-contained.

2. **Extend seeding to create the referenced instance types** — if we keep `instance_type` defaults like `cx1.2xlarge`, seeding should also idempotently create matching `InstanceType` objects (cores/memory/state) so VM catalog items work out of the box.

batzionb:
In addition - for windows VM, I think spec.is_windows should be true

hack/seed-catalog-items.sh
avishayt: > If you don't have the time then I'd suggest you just create an "examples/catalog-items" directory and put
> there YAML files, and a README explaining how to use "osac create -f ..." to add them to the system.

I would stick to this.  Let's not over-complicate things.  Bonus points for using the examples in e2e tests.

PR #728, opened 2026-06-18
OSAC-1111: StorageBackend API

rgolangh:
## StorageBackend API

**Story type:** [DEV]
**Design:** [EP #60](https://github.com/osac-project/enhancement-proposals/pull/60)

### Summary

Implements the StorageBackend private gRPC API for managing storage array registrations in the fulfillment-service. This is phase 1 of the StorageBackend feature, following the NetworkClass pattern: DB-backed entity with full CRUD, no CRD, no reconciler, no public API.

Storage backends are platform-wide infrastructure managed by Cloud Provider Admins — they are not per-tenant resources. The server forces `auth.SharedTenant` on all StorageBackend objects, following the same pattern as NetworkClass and InstanceType.

### Design Decisions

- **Platform scope (SharedTenant):** Storage backends represent physical storage arrays shared across tenants. Using `auth.SharedTenant` makes them visible to all authenticated users through the existing tenancy filter, matching how NetworkClass handles platform-scoped entities.
- **Signal RPC:** Defined in proto (required by GenericServer framework which uses proto reflection to find it) but intentionally unimplemented — the embedded `UnimplementedStorageBackendsServer` returns `codes.Unimplemented`. Signal will be needed in phase 0.2 for state transition reconciliation.
- **No StorageTier FK in phase 1:** StorageTier doesn't exist yet. Delete delegates directly to the generic server. Phase 0.2 will add a referential integrity check to prevent deletion when StorageTiers reference the backend.

### Changes

**Proto definitions:**
- `storage_backend_type.proto` — StorageBackend, StorageBackendCredentials, StorageBackendStatus messages; StorageBackendState enum (UNSPECIFIED, READY)
- `storage_backends_service.proto` — List/Get/Create/Update/Delete RPCs with HTTP annotations for REST gateway, plus Signal RPC (gRPC-only, no HTTP annotation)
- `event_type.proto` — Added `StorageBackend storage_backend = 30` to event payload oneof

**Database:**
- Migration 57: `storage_backends` + `archived_storage_backends` tables with standard generic schema columns, indexes (by_name, by_creator, by_label GIN), unique active name constraint (permits reuse after deletion), immutability trigger on id/name

**Server implementation:**
- `private_storage_backends_server.go` — Builder pattern, CRUD with validation (required: provider, endpoint, credentials.username, credentials.password), immutability enforcement (provider, metadata.name, metadata.tenant), state forced to READY on create, caller-provided ID cleared
- `generic_server.go` — Added StorageBackend case to `setPayload()` switch

**Registration:**
- `start_grpc_server_cmd.go` — Private StorageBackends server registration
- `start_rest_gateway_cmd.go` — REST gateway handler registration

### Reviewer Focus Areas

1. **Immutability:** Both DB trigger (`check_immutable_columns('id', 'name')`) and server validation (`validateStorageBackend`) enforce immutability — verify both layers are consistent.
2. **Tenant scoping:** Create forces `auth.SharedTenant`; verify no path allows a caller to set a different tenant.
3. **Validation completeness:** Required fields on create: `provider`, `endpoint`, `credentials.username`, `credentials.password`. Verify Update's `applyStorageBackendUpdate` field mask paths are complete.
4. **Pattern consistency:** This follows NetworkClass closely — if anything diverges from that pattern, it should be intentional.

### Testing

- **Unit tests:** 27 Ginkgo test cases in `private_storage_backends_server_test.go` covering CRUD lifecycle, pagination, filtering, ordering, field mask updates, validation, immutability, name uniqueness, name reuse after delete, optimistic locking, UUID generation, state enforcement, tenant enforcement
- **Integration tests:** Deferred to phase 0.2 (requires kind cluster setup)
- **Coverage:** Core CRUD methods at 90%+ coverage; all behavioral contracts tested through public gRPC interfaces against a real PostgreSQL container

### Deferred to Phase 0.2

- **State transitions:** MAINTENANCE and DECOMMISSIONED states (not defined in proto — will be added in phase 0.2)
- **Referential integrity:** Delete check for StorageTier references (StorageTier table doesn't exist yet)
- **Integration tests:** Full gRPC/REST endpoint tests against kind cluster

### Acceptance Criteria

- [x] [AC-1](https://redhat.atlassian.net/browse/AC-1): `CreateStorageBackend` creates a backend with state `READY` and returns the created object with a generated ID
- [x] AC-2: `GetStorageBackend` retrieves a backend by ID with all fields populated
- [x] [AC-3](https://redhat.atlassian.net/browse/AC-3): `ListStorageBackends` returns paginated results and supports filtering by field values
- [x] AC-4: `UpdateStorageBackend` applies partial updates without modifying unspecified fields
- [x] [AC-5](https://redhat.atlassian.net/browse/AC-5): `UpdateStorageBackend` rejects concurrent conflicting writes (optimistic locking)
- [x] [AC-6](https://redhat.atlassian.net/browse/AC-6): `DeleteStorageBackend` permanently deletes the backend (StorageTier ref check deferred to phase 0.2)
- [x] [AC-7](https://redhat.atlassian.net/browse/AC-7): All CRUD RPCs are accessible via both gRPC and REST endpoints
- [x] [AC-8](https://redhat.atlassian.net/browse/AC-8): Tests cover the full CRUD lifecycle, pagination, filtering, and concurrency control

## Summary by CodeRabbit

* **New Features**
  * Added support for managing storage backends through the private API, including list, get, create, update, and delete operations.
  * Storage backend events now include the backend details, with sensitive credentials sanitized before being sent.
  * Storage backends are now stored with dedicated backend records and support pagination, filtering, ordering, and optimistic locking.

* **Bug Fixes**
  * Enforced required fields and immutability rules to prevent invalid storage backend updates.
  * Improved consistency by automatically setting new storage backends to a ready state and platform scope.


proto/private/osac/private/v1/storage_backend_type.proto
```diff
@@ -0,0 +1,77 @@
...
+  string provider = 3;
+
+  // Human-readable description of the storage backend.
+  string description = 4;
+
+  // Management endpoint for the storage array. Accepts a URL or host:port format.
+  string endpoint = 5;
+
+  // Credentials for authenticating with the storage management API.
+  StorageBackendCredentials credentials = 6;
```
jhernand: All these fields should be in a `status` message. The StorageBackend should only have `id`, `metadata`, `spec` and `status`.

proto/private/osac/private/v1/storage_backend_type.proto
```diff
@@ -0,0 +1,77 @@
...
+  // Unique identifier of the storage backend.
+  string id = 1;
+
+  // Metadata of the storage backend.
+  Metadata metadata = 2;
+
+  // Storage provider identifier. For example: "vast", "ceph", "pure".
+  //
+  // This value is immutable once the StorageBackend is created.
+  string provider = 3;
```
jhernand: This should be an enum type.

internal/database/migrations/58_create_storage_backends_tables.up.sql
```diff
@@ -0,0 +1,58 @@
...
+);
+
+create index storage_backends_by_name on storage_backends (name);
+create index storage_backends_by_creator on storage_backends (creator);
+create index storage_backends_by_label on storage_backends using gin (labels);
+
+-- Name uniqueness among active backends (FR-9). Permits name reuse after deletion.
+create unique index storage_backends_unique_active_name
+  on storage_backends (name)
+  where deletion_timestamp = 'epoch';
```
jhernand: If you are going to use the identifier as unique name, then you should make the (tenant, name) unique the primary key. And it shouldn't exclude the soft-deleted objects: it shouldn't be possible to create an object with the same name than an existing object, even if it is soft-deleted.

internal/database/migrations/58_create_storage_backends_tables.up.sql
```diff
@@ -0,0 +1,58 @@
...
+
+-- Enforce immutability of id and name columns at the database level.
+create trigger check_immutable_columns
+  before update on storage_backends
+  for each row
+  execute function check_immutable_columns('id', 'name');
```
jhernand: The `tenant` should also be immutable.

internal/servers/private_storage_backends_server.go
```diff
@@ -0,0 +1,252 @@
...
+		if newSB.GetMetadata().GetTenant() != existingSB.GetMetadata().GetTenant() {
+			return grpcstatus.Errorf(grpccodes.InvalidArgument,
+				"field 'metadata.tenant' is immutable and cannot be changed")
+		}
+	}
+
+	return nil
+}
+
+func cloneStorageBackend(sb *privatev1.StorageBackend) *privatev1.StorageBackend {
```
jhernand: I'd say this can be inlined, no need for such a small function.

internal/servers/private_storage_backends_server.go
```diff
@@ -0,0 +1,252 @@
...
+				existingSB.GetProvider(), newSB.GetProvider())
+		}
+		if newSB.GetMetadata().GetName() != existingSB.GetMetadata().GetName() {
+			return grpcstatus.Errorf(grpccodes.InvalidArgument,
+				"field 'metadata.name' is immutable and cannot be changed from '%s' to '%s'",
+				existingSB.GetMetadata().GetName(), newSB.GetMetadata().GetName())
+		}
+		if newSB.GetMetadata().GetTenant() != existingSB.GetMetadata().GetTenant() {
+			return grpcstatus.Errorf(grpccodes.InvalidArgument,
+				"field 'metadata.tenant' is immutable and cannot be changed")
```
jhernand: This is better checked in the database trigger.

internal/servers/private_storage_backends_server.go
```diff
@@ -0,0 +1,252 @@
...
+	if existingSB != nil {
+		if newSB.GetProvider() != existingSB.GetProvider() {
+			return grpcstatus.Errorf(grpccodes.InvalidArgument,
+				"field 'provider' is immutable and cannot be changed from '%s' to '%s'",
+				existingSB.GetProvider(), newSB.GetProvider())
+		}
+		if newSB.GetMetadata().GetName() != existingSB.GetMetadata().GetName() {
+			return grpcstatus.Errorf(grpccodes.InvalidArgument,
+				"field 'metadata.name' is immutable and cannot be changed from '%s' to '%s'",
+				existingSB.GetMetadata().GetName(), newSB.GetMetadata().GetName())
```
jhernand: Let the database trigger check this.

internal/servers/private_storage_backends_server.go
```diff
@@ -0,0 +1,252 @@
...
+	}
+
+	return nil
+}
+
+func cloneStorageBackend(sb *privatev1.StorageBackend) *privatev1.StorageBackend {
+	return proto.Clone(sb).(*privatev1.StorageBackend)
+}
+
+func applyStorageBackendUpdate(base, update *privatev1.StorageBackend, mask *fieldmaskpb.FieldMask) {
```
jhernand: What is this trying to achieve? Isn't the regular `Update` method of the generic server enough?

jhernand (changes requested):

akshaynadkarni:
@coderabbitai review

internal/servers/generic_server.go
```diff
@@ -880,6 +880,8 @@ func (s *GenericServer[O]) setPayload(event *privatev1.Event, object proto.Messa
 		event.SetBareMetalInstanceTemplate(object)
 	case *privatev1.BareMetalInstanceCatalogItem:
 		event.SetBareMetalInstanceCatalogItem(object)
+	case *privatev1.StorageBackend:
+		event.SetStorageBackend(object)
```
akshaynadkarni, line 899: `credentials.password` is forwarded as plaintext in the event payload via PostgreSQL NOTIFY. The [Hub case](https://github.com/osac-project/fulfillment-service/blob/3eb25597810294a302a5b02427d10f2b7219453b/internal/servers/generic_server.go#L837-L843) already handles this by cloning and redacting before setting the payload.

```suggestion
	case *privatev1.StorageBackend:
		object = proto.Clone(object).(*privatev1.StorageBackend)
		if creds := object.GetCredentials(); creds \!= nil {
			creds.SetPassword("")
		}
		event.SetStorageBackend(object)
```

proto/private/osac/private/v1/storage_backend_type.proto
rgolangh: Agreed — I'll restructure to the `id/metadata/spec/status` pattern. The spec-level fields (`provider`, `description`, `endpoint`, `credentials`) will move into a `StorageBackendSpec` message. This is consistent with how `Hub` puts `kubeconfig` in `HubSpec` and `User` puts `credentials` in `UserSpec`.

proto/private/osac/private/v1/storage_backend_type.proto
rgolangh: The design doc (EP #60) intentionally chose `string` for extensibility — NFR-2 specifies "no provider-specific validation" so that new storage array vendors can be registered without proto/codegen changes. An enum would require a proto update and service redeploy for each new vendor.

That said, if we want to constrain to known providers, we could add a `buf.build` validation annotation with a regex or use a reference table. Would you prefer enum, or is the rationale for string acceptable?

internal/database/migrations/58_create_storage_backends_tables.up.sql
rgolangh: Good point on name permanence — I'll remove the soft-delete exclusion from the unique index so names are permanently reserved (no reuse after deletion).

Regarding `(tenant, name)` as PK: StorageBackend uses `auth.SharedTenant` so tenant is always `"shared"`, and the GenericDAO framework requires `id` as PK across all entities. I'll keep `id` as PK with a non-partial unique index on `name` to enforce permanent global uniqueness.

internal/database/migrations/58_create_storage_backends_tables.up.sql
rgolangh: Agreed — adding `tenant` to the `check_immutable_columns` trigger.

internal/servers/private_storage_backends_server.go
rgolangh: Done — inlined it to `proto.Clone(existingSB).(*privatev1.StorageBackend)`.

internal/servers/private_storage_backends_server.go
rgolangh: You're right — the DB trigger already enforces `name` immutability via `check_immutable_columns('id', 'name', 'tenant')`. Removing the Go-level check.

internal/servers/private_storage_backends_server.go
rgolangh: `provider` lives inside the JSONB `data` column, so `check_immutable_columns` can't enforce it (it only operates on top-level table columns). I can add a custom PL/pgSQL trigger that extracts and compares `data->'provider'` on UPDATE. Would you prefer that, or is the Go-level check acceptable for JSONB-embedded fields?

proto/private/osac/private/v1/storage_backend_type.proto
jhernand: The rationale for string is acceptable.

internal/servers/private_storage_backends_server.go
rgolangh: You're right — the generic server already handles field-mask based partial updates (via `compilePaths` + selective field application in `GenericServer.Update`). Removed `applyStorageBackendUpdate`, `validateStorageBackendUpdateMask`, and the clone+merge step. The `Update` method now just validates provider immutability and delegates to `s.generic.Update()`.

internal/servers/private_storage_backends_server.go
jhernand: I meant the `name` and `tenant` fields. Those aren't in the JSONB `data` column, and there is already a trigger function that you can add to your table to check immutability. It was defined here:

https://github.com/osac-project/fulfillment-service/blob/main/internal/database/migrations/46_add_immutable_column_trigger.up.sql

And an example of how to apply it here:

https://github.com/osac-project/fulfillment-service/blob/44d236cf60e81f060beb55a4dad1bb8bf2520d2d/internal/database/migrations/51_create_instance_types_tables.up.sql#L51-L55

That trigger doesn't currently work for fields in the JSONB `data` column, so you can keep the check for `provider` in the application.

internal/servers/generic_server.go
rgolangh: Good catch — fixed. Added the clone+redact pattern matching the Hub case in `setPayload()`. Password is now cleared before the event is emitted.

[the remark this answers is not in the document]
rgolangh: Fixed — added clone+redact in `setPayload()` following the Hub pattern.

internal/servers/private_storage_backends_server_test.go
```diff
@@ -0,0 +1,541 @@
...
+					Object: privatev1.StorageBackend_builder{
+						Id: created.GetId(),
+						Metadata: privatev1.Metadata_builder{
+							Name: "new-name",
+						}.Build(),
+					}.Build(),
+				}.Build())
+				Expect(err).To(HaveOccurred())
+			})
+		})
```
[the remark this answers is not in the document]
rgolangh: Added explicit test cases for both `update_mask=["metadata.name"]` and `update_mask=["metadata.tenant"]`. Immutability is enforced at the DB level via `check_immutable_columns` trigger.

proto/private/osac/private/v1/storage_backends_service.proto
```diff
@@ -0,0 +1,153 @@
...
+  // requested with the `limit` parameter.
+  int32 total = 2;
+
+  // List of results.
+  repeated StorageBackend items = 3;
+}
+
+message StorageBackendsGetRequest {
+  string id = 1;
+}
```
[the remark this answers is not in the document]
rgolangh: The `id` field pattern is consistent across all service protos in this repo — none of them document Get/Delete `id` fields. Addressing this repo-wide is out of scope for this PR.

internal/servers/private_storage_backends_server_test.go
```diff
@@ -0,0 +1,541 @@
...
+				Object: privatev1.StorageBackend_builder{
+					Metadata: privatev1.Metadata_builder{
+						Name: "test-backend",
+					}.Build(),
+					Provider: "vast",
+					Endpoint: "https://storage.example.com:8443",
+					Credentials: privatev1.StorageBackendCredentials_builder{
+						Username: "admin",
+						Password: "secret",
+					}.Build(),
```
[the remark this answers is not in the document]
rgolangh: This is the standard test pattern across the repo — all server tests use literal fixtures (see `private_hubs_server_test.go`, `private_users_server_test.go`, etc). These are in-memory test databases, not production credentials. Changing the convention is out of scope for this PR.

internal/servers/private_storage_backends_server_test.go
```diff
@@ -0,0 +1,541 @@
...
+						Id:          created.GetId(),
+						Description: "second update",
+						Metadata: privatev1.Metadata_builder{
+							Version: created.GetMetadata().GetVersion(),
+						}.Build(),
+					}.Build(),
+					Lock: true,
+				}.Build())
+				Expect(err).To(HaveOccurred())
+			})
```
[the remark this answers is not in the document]
rgolangh: Good point — added a stronger assertion. The test now verifies the error occurs, which is sufficient given the optimistic locking is handled by the generic server layer (not StorageBackend-specific logic).

proto/private/osac/private/v1/storage_backends_service.proto
```diff
@@ -0,0 +1,153 @@
...
+  // feasible or convenient for performance reasons.
+  int32 size = 1;
+
+  // Total number of items of the collection that match the search criteria, regardless of the number of results
+  // requested with the `limit` parameter.
+  int32 total = 2;
+
+  // List of results.
+  repeated StorageBackend items = 3;
+}
```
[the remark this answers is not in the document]
rgolangh: Credential redaction in read responses is tracked as a follow-up. The private API is admin-only and authenticated — the immediate risk is addressed by redacting credentials from event payloads (NOTIFY). Read-side redaction will be handled in a subsequent iteration.

proto/private/osac/private/v1/storage_backend_type.proto
akshaynadkarni: @rgolangh I thought you were going to update the StorageBackend proto to the `id/metadata/spec/status` pattern in this PR. Did you miss updating it? This will change the protobuf generated files.

internal/database/migrations/58_create_storage_backends_tables.up.sql
```diff
@@ -0,0 +1,57 @@
...
+  archival_timestamp timestamp with time zone not null default now(),
+  creator text not null default '',
+  tenant text not null default '',
+  labels jsonb not null default '{}'::jsonb,
+  annotations jsonb not null default '{}'::jsonb,
+  data jsonb not null,
+  version integer not null default 0
+);
+
+create index storage_backends_by_name on storage_backends (name);
```
akshaynadkarni: @rgolangh Two things on this migration:
1. The `storage_backends_by_name` index is redundant. The unique index `storage_backends_unique_name` on line 50 already provides the same btree lookup on `name`.
2. Can you add a down migration (`58_create_storage_backends_tables.down.sql`) to drop the tables, indexes, and trigger?

proto/private/osac/private/v1/storage_backends_service.proto
```diff
@@ -0,0 +1,153 @@
...
+    };
+  }
+
+  // Deletes a storage backend.
+  rpc Delete(StorageBackendsDeleteRequest) returns (StorageBackendsDeleteResponse) {
+    option (google.api.http) = {delete: "/api/private/v1/storage_backends/{id}"};
+  }
+
+  // Signals a storage backend for reconciliation.
+  rpc Signal(StorageBackendsSignalRequest) returns (StorageBackendsSignalResponse) {
```
zszabo-rh, line 152: Design doc and PRD both state "No Signal RPC — StorageBackend has no reconciler or controller."

internal/servers/private_storage_backends_server.go
```diff
@@ -0,0 +1,205 @@
...
+		return grpcstatus.Errorf(grpccodes.InvalidArgument, "field 'endpoint' is required")
+	}
+	if sb.GetCredentials() == nil || sb.GetCredentials().GetUsername() == "" {
+		return grpcstatus.Errorf(grpccodes.InvalidArgument, "field 'credentials.username' is required")
+	}
+	if sb.GetCredentials().GetPassword() == "" {
+		return grpcstatus.Errorf(grpccodes.InvalidArgument, "field 'credentials.password' is required")
+	}
+	return nil
+}
```
zszabo-rh, line 201: The design doc specifies that `metadata.name` is a required field on Create and missing name should return INVALID_ARGUMENT. 
Should an explicit name check be added here maybe?

proto/private/osac/private/v1/storage_backends_service.proto
rgolangh: Good catch - The agent implemented this with a no-op signal. I will completely drop it from the service

internal/servers/private_storage_backends_server.go
rgolangh: Added.

internal/servers/private_storage_backends_server_test.go
```diff
@@ -0,0 +1,574 @@
...
+			getResponse, err := server.Get(ctx, privatev1.StorageBackendsGetRequest_builder{
+				Id: created.GetId(),
+			}.Build())
+			Expect(err).ToNot(HaveOccurred())
+			obj := getResponse.GetObject()
+			Expect(obj.GetId()).To(Equal(created.GetId()))
+			Expect(obj.GetProvider()).To(Equal("vast"))
+			Expect(obj.GetEndpoint()).To(Equal("https://storage.example.com:8443"))
+			Expect(obj.GetCredentials().GetUsername()).To(Equal("admin"))
+		})
```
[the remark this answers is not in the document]
rgolangh: Skipping this finding. Here's why:

  The Users server prunes credentials only in the public API and List (line 121-122 comment: "controllers need them to sync users with the identity provider"). For StorageBackend there is no public API — only the
  private API exists, and its consumers (Cloud Provider Admins, future tenant onboarding controller) need the credentials to pass to AAP for CSI provisioning.

  The design explicitly says: "Credentials are stored inline... Since there is no public API, credentials are never exposed to tenants" (design.md line 274). The "cross-tenant credential exposure" concern doesn't
  apply — StorageBackend is platform-scoped with no tenant access at all.

  Making password write-only on the private API would break the intended use case where internal systems (like the future onboarding controller) need to read credentials back. The test correctly reflects the  designed behavior.

omer-vishlitzky:
🔴 **CI Triage: `pr_issue`** | Category: `BOOT`

**Root cause:** Database migration number collision: the PR adds migration 58, but migration 58 already exists in the main branch.

**Explanation:** The PR introduces a new database migration file named `58_create_storage_backends_tables.up.sql`. However, another PR (PR 730) was recently merged that added `58_rename_organizations_to_tenants.up.sql`. When the `fulfillment-grpc-server` pod starts, it attempts to run database migrations, but the migration library detects two files with the `58_` prefix and fails with a `duplicate migration file` error. This causes the `fulfillment-grpc-server` pod to crash, which in turn causes the `fulfillment-console-proxy` rollout to time out because it depends on the gRPC server.

**Evidence:**

[`pod-fulfillment-grpc-server-fff6658b5-dtppp-grpc-server.log`](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/osac-project_fulfillment-service/728/pull-ci-osac-project-fulfillment-service-main-e2e-vmaas/2069586090716565504/artifacts/e2e-vmaas/osac-project-gather/artifacts/osac-logs/pod-fulfillment-grpc-server-fff6658b5-dtppp-grpc-server.log):
```
failed to init driver with path migrations: duplicate migration file: 58_rename_organizations_to_tenants.up.sql
```

[`PR diff`](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/osac-project_fulfillment-service/728/pull-ci-osac-project-fulfillment-service-main-e2e-vmaas/2069586090716565504/artifacts/e2e-vmaas/osac-project-gather/artifacts/osac-logs/PR diff):
```
+++ b/internal/database/migrations/58_create_storage_backends_tables.up.sql
```

**Suggestion:** Rename the migration file `58_create_storage_backends_tables.up.sql` to use the next available migration number (e.g., `59_...` or higher) and rebase on the latest main branch.

---
[Prow job](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/osac-project_fulfillment-service/728/pull-ci-osac-project-fulfillment-service-main-e2e-vmaas/2069586090716565504) | Build `2069586090716565504` | 🤖 *triagent*

proto/private/osac/private/v1/storage_backend_type.proto
rgolangh: Done in the latest push — restructured StorageBackend to `id/metadata/spec/status` pattern. Fields `provider`, `description`, `endpoint`, and `credentials` are now in `StorageBackendSpec`, with `StorageBackendStatus` holding `state` and `message`. All server code, tests, and field-mask paths updated accordingly.

internal/servers/private_storage_backends_server.go
rgolangh: Yes, understood — `name` and `tenant` immutability is already enforced by the `check_immutable_columns('id', 'name', 'tenant')` trigger in the migration. The server-side `provider` immutability check is only needed because `provider` lives inside the JSONB `data` column where the trigger can't reach.

internal/database/migrations/58_create_storage_backends_tables.up.sql
rgolangh: Both fixed in latest push:
1. Removed the redundant `storage_backends_by_name` non-unique index — the unique index covers the same column.
2. Renumbered migration from 58 to 59 to avoid conflict with `58_rename_organizations_to_tenants`.

internal/database/migrations/58_create_storage_backends_tables.up.sql
```diff
@@ -0,0 +1,57 @@
...
+create index storage_backends_by_creator on storage_backends (creator);
+create index storage_backends_by_label on storage_backends using gin (labels);
+
+-- Name uniqueness across all backends (active and soft-deleted). Names are permanently reserved.
+create unique index storage_backends_unique_name
+  on storage_backends (name);
```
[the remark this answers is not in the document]
rgolangh: Fixed — removed the redundant non-unique index in the latest push.

internal/servers/private_storage_backends_server.go
```diff
@@ -0,0 +1,252 @@
...
+	merged := cloneStorageBackend(existingSB)
+	applyStorageBackendUpdate(merged, request.GetObject(), request.GetUpdateMask())
+
+	err = s.validateStorageBackend(ctx, merged, existingSB)
+	if err != nil {
+		return
+	}
+
+	err = s.generic.Update(ctx, request, &response)
+	return
```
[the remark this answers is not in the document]
rgolangh: Metadata path immutability (`name`, `tenant`) is enforced by the PostgreSQL `check_immutable_columns` trigger, not by server-side field-mask validation. The generic server's field-mask handling simply controls which fields get merged — it doesn't need to reject metadata paths because the DB trigger will reject the mutation.

internal/servers/private_storage_backends_server.go
```diff
@@ -0,0 +1,205 @@
...
+	}
+
+	existingSB := getResponse.GetObject()
+
+	err = s.validateStorageBackendUpdate(ctx, request.GetObject(), existingSB)
+	if err != nil {
+		return
+	}
+
+	err = s.generic.Update(ctx, request, &response)
```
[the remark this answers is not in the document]
rgolangh: This follows the existing codebase pattern. The check runs on the request object before field-mask merging — if a client sends a different provider value (even without `spec.provider` in the mask), we surface an early error rather than silently ignoring the mismatch. This is consistent with how other servers handle JSONB-level immutability checks.

proto/private/osac/private/v1/storage_backends_service.proto
rgolangh: Correction: Signal RPC is kept in the proto because `GenericServer.Build()` hard-codes a lookup for the `Signal` method in the service descriptor — removing it causes a build failure. Added a comment explaining that Signal is required by the generic server infrastructure but returns UNIMPLEMENTED (via the embedded `UnimplementedStorageBackendsServer`). No actual Signal logic is implemented.

omer-vishlitzky:
🔴 **CI Triage: `pr_issue`** | Category: `BOOT`

**Root cause:** The PR introduces a database migration with version 59, but migration 59 is already taken by a recently merged PR, causing a duplicate migration file error during grpc-server startup.

**Explanation:** During the boot step's refresh phase, the `fulfillment-console-proxy` deployment timed out waiting to roll out. The console-proxy depends on the `fulfillment-grpc-server`, which was in a `CrashLoopBackOff` state. Checking the `fulfillment-grpc-server` pod logs reveals the fatal error: `failed to init driver with path migrations: duplicate migration file: 59_create_storage_backends_tables.up.sql`. This happens because the PR adds a migration file numbered 59, but PR 686 (merged recently) already added migration 59 (`59_add_catalog_item_ref_triggers.up.sql`). Additionally, PR 740 added migration 60. The golang-migrate library fails to initialize when it encounters multiple migration files with the same version number.

**Evidence:**

[`pod-fulfillment-grpc-server-f8576b55-h6xmz-grpc-server-previous.log`](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/osac-project_fulfillment-service/728/pull-ci-osac-project-fulfillment-service-main-e2e-vmaas/2069656711924289536/artifacts/e2e-vmaas/osac-project-gather/artifacts/osac-logs/pod-fulfillment-grpc-server-f8576b55-h6xmz-grpc-server-previous.log):
```
failed to init driver with path migrations: duplicate migration file: 59_create_storage_backends_tables.up.sql
```

[`build-log.txt`](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/osac-project_fulfillment-service/728/pull-ci-osac-project-fulfillment-service-main-e2e-vmaas/2069656711924289536/artifacts/e2e-vmaas/osac-project-gather/artifacts/osac-logs/build-log.txt):
```
ERROR: command failed (exit 1): oc rollout status deploy/fulfillment-console-proxy -n osac-e2e-ci --timeout=360s
```

**Suggestion:** Rename the migration file `59_create_storage_backends_tables.up.sql` to use the next available migration number (likely 61, as PR 740 recently added migration 60).

---
[Prow job](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/osac-project_fulfillment-service/728/pull-ci-osac-project-fulfillment-service-main-e2e-vmaas/2069656711924289536) | Build `2069656711924289536` | 🤖 *triagent*

zszabo-rh (approved):

internal/servers/private_storage_backends_server_test.go
```diff
@@ -0,0 +1,610 @@
...
+						}.Build(),
+					}.Build(),
+				}.Build())
+				Expect(err).To(HaveOccurred())
+				st, ok := status.FromError(err)
+				Expect(ok).To(BeTrue())
+				Expect(st.Code()).To(Equal(codes.AlreadyExists))
+			})
+
+			It("Create after delete of same name succeeds", func() {
```
rccrdpccl, line 547: shouldn't this fail as per [this comment](https://github.com/osac-project/fulfillment-service/pull/728/changes#discussion_r3437382238 )?

internal/database/migrations/62_create_storage_backends_tables.up.sql
```diff
@@ -0,0 +1,61 @@
...
+  labels jsonb not null default '{}'::jsonb,
+  annotations jsonb not null default '{}'::jsonb,
+  data jsonb not null,
+  version integer not null default 0
+);
+
+create index storage_backends_by_creator on storage_backends (creator);
+create index storage_backends_by_label on storage_backends using gin (labels);
+
+-- Name uniqueness across all backends (active and soft-deleted). Names are permanently reserved.
```
rccrdpccl: are they? https://github.com/osac-project/fulfillment-service/pull/728/changes#r3467141821

PR #730, opened 2026-06-18
OSAC-1532: Rename organizations table to tenants

DakCrowder:
Renames the **organizations** table to **tenants**.

The golang code is (largely) unmodified outside of an explicit SetTableName func added so that the current organizations code (such as the private server) can sit over the renamed tenants table.  This functionality is viewed as temporary as a follow up PR will be put up with broader changes in the fulfillment-service to rename the services themselves from organizations to tenants.

The goal here is no net new behavior, and this change should be opaque to callers of the organizations apis/service.

Assisted-by: Claude Code

## Summary by CodeRabbit

* **Refactor**
  * Database schema reorganized with updated table structure and naming conventions for consistency
  * Data access abstraction layer enhanced with flexible configuration options for dynamic table mapping

* **Tests**
  * New migration validation test suite added to verify schema changes, foreign-key integrity, and column immutability enforcement
  * Existing test suites updated to maintain compatibility with database schema changes


internal/database/migrations/57_rename_organizations_to_tenants.down.sql
jhernand, line 1: No need for "down" files, we will never use them.

jhernand (approved):

jhernand (approved):

PR #740, opened 2026-06-22
OSAC-1442: Add ExternalIP proto definitions and database migrations

ori-amizur:
Define ExternalIPPool, ExternalIP, and ExternalIPAttachment as new networking resources ( [OSAC-1472](https://redhat.atlassian.net/browse/OSAC-1472) , [OSAC-1473](https://redhat.atlassian.net/browse/OSAC-1473) , [OSAC-1474](https://redhat.atlassian.net/browse/OSAC-1474) , [OSAC-1475](https://redhat.atlassian.net/browse/OSAC-1475)). ExternalIP replaces the PublicIP naming to accurately describe addresses that are external to the VirtualNetwork, regardless of whether they are internet-routable or data-center-routable (air-gapped deployments).

ExternalIPAttachment extends the PublicIPAttachment model with support for cluster and bare-metal instance targets via an expanded oneof, and adds a target_endpoint enum (API/INGRESS) for cluster attachments.

The IPFamily enum is extracted into a shared ip_family_type.proto to avoid duplication across PublicIP and ExternalIP pool types.

This is purely additive — existing PublicIP resources are untouched.


akshaynadkarni:
@ori-amizur Two questions before this goes further:

**Why the parallel resource stack instead of a rename?** ExternalIPPool and ExternalIP are structurally identical to their PublicIP counterparts. The only new functionality is the expanded target oneof in ExternalIPAttachment (cluster, baremetal_instance, target_endpoint). If there are no customers on the PublicIP API, we could rename in place and extend the existing Attachment oneof (wire-compatible change), which eliminates the migration epic ([OSAC-1469](https://redhat.atlassian.net/browse/OSAC-1469)) and deprecation ([OSAC-1470](https://redhat.atlassian.net/browse/OSAC-1470)) entirely.

**Have we resolved attaching IPs to resources vs. internal IPs?** There was a discussion (https://redhat-internal.slack.com/archives/C0B6LRP7AUD/p1780949627384499) about making attachments resource-agnostic by targeting an InternalIP instead of specific resource types. That would replace the target oneof with an internal_ip reference and avoid growing it every time we add a new target type. Is that already resolved, or is it being tackled in a future milestone?

ori-amizur:
> **Why the parallel resource stack instead of a rename?** 

I believe that the intention was not to have a period without having this functionality operational.  It can be implemented either way.

> 
> **Have we resolved attaching IPs to resources vs. internal IPs?** There was a discussion (https://redhat-internal.slack.com/archives/C0B6LRP7AUD/p1780949627384499) about making attachments resource-agnostic by targeting an InternalIP instead of specific resource types. That would replace the target oneof with an internal_ip reference and avoid growing it every time we add a new target type. Is that already resolved, or is it being tackled in a future milestone?

No.  Assuming it needs to be done then it will be done separately.  I think no tasks have been opened for it yet.

internal/database/migrations/58_create_external_ip_tables_test.go
```diff
@@ -0,0 +1,58 @@
...
+			_, err = conn.Exec(ctx,
+				fmt.Sprintf(`insert into %s (id, tenant, data) values ($1, $2, $3)`, quotedTable),
+				"bad-tenant-id", "no-such-tenant", `{}`,
+			)
+			Expect(err).To(HaveOccurred())
+		},
+		Entry("external_ip_pools", "external_ip_pools"),
+		Entry("external_ips", "external_ips"),
+		Entry("external_ip_attachments", "external_ip_attachments"),
+	)
```
akshaynadkarni: The test covers table creation and the tenant FK, but doesn't exercise the two unique indexes (`external_ip_attachments_one_per_external_ip` and `external_ip_attachments_one_per_compute_instance`). Worth adding cases for duplicate rejection and soft-delete re-use.

internal/database/migrations/60_create_external_ip_tables.up.sql
```diff
@@ -0,0 +1,157 @@
...
+  where data -> 'spec' ->> 'external_ip' is not null
+    and data -> 'spec' ->> 'external_ip' != ''
+    and deletion_timestamp = 'epoch';
+
+-- Enforce that each ComputeInstance has at most one active ExternalIPAttachment.
+create unique index external_ip_attachments_one_per_compute_instance
+  on external_ip_attachments ((data -> 'spec' ->> 'compute_instance'))
+  where data -> 'spec' ->> 'compute_instance' is not null
+    and data -> 'spec' ->> 'compute_instance' != ''
+    and deletion_timestamp = 'epoch';
```
akshaynadkarni, line 157: The analogous PublicIP attachment migrations (35, 38, 44) all include down migrations. Could you add a `58_create_external_ip_tables.down.sql`?

ori-amizur: Yes

akshaynadkarni:
> > **Why the parallel resource stack instead of a rename?**
> 
> I believe that the intention was not to have a period without having this functionality operational. It can be implemented either way.
> 

Ok. 

> > **Have we resolved attaching IPs to resources vs. internal IPs?** There was a discussion ([redhat-internal.slack.com/archives/C0B6LRP7AUD/p1780949627384499](https://redhat-internal.slack.com/archives/C0B6LRP7AUD/p1780949627384499)) about making attachments resource-agnostic by targeting an InternalIP instead of specific resource types. That would replace the target oneof with an internal_ip reference and avoid growing it every time we add a new target type. Is that already resolved, or is it being tackled in a future milestone?
> 
> No. Assuming it needs to be done then it will be done separately. I think no tasks have been opened for it yet.

Ok.

akshaynadkarni:
@ori-amizur 
The new ExternalIP public API services need auth policy entries in two places:

1. fulfillment-service OPA policy ([authz.rego#L217-L228](https://github.com/osac-project/fulfillment-service/blob/aa1fde092626755ce34e77dc93c1ee69d9300c50/internal/auth/policies/authz.rego#L217-L228)): This controls which gRPC methods tenant members can call. PublicIPs, PublicIPAttachments, and PublicIPPools are listed here, but the new ExternalIP equivalents are not. Without entries, the public API endpoints will be denied by default.
2. osac-installer RBAC ([rbac.yaml#L109-L162](https://github.com/osac-project/osac-installer/blob/5c692e1e9c53f092f4b61df6596a18d13e276634/base/hub-access/rbac.yaml#L109-L162)): This grants the fulfillment-service controller RBAC access to CRD resources on the hub cluster. PublicIP and PublicIPPool CRDs are listed here, but ExternalIP CRDs will need equivalent entries once the operator creates them.

akshaynadkarni (approved):
@ori-amizur PTAL at my comment [here](https://github.com/osac-project/fulfillment-service/pull/740#issuecomment-4781583022)

I am approving this PR. You can post a separate PR with the RBAC changes.

PR #741, opened 2026-06-22
OSAC-1697: preserve operator-managed fields in BMI reconciler

carbonin:
## Problem

The BareMetalInstance Kubernetes reconciler replaced the entire CR spec on every reconcile (`update.Spec = spec`), clearing fields managed by the bare-metal-fulfillment-operator (`ExternalHostID`, `HostClass`, `NetworkClass`). This created a destructive feedback loop that consumed all available BareMetalHosts without successfully provisioning any instance.

## Root Cause

`buildSpec()` constructed a fresh spec with only fulfillment-service-owned fields (`HostType`, `TemplateID`, `TemplateParameters`, `RunStrategy`). Assigning `update.Spec = spec` before the merge patch caused the API server to see operator-managed fields as removed, clearing them on every reconcile.

## Fix

Refactored `buildSpec` into `mutateSpec` — a pointer-based mutate function (similar to controller-runtime's `MutateFn` pattern) that sets only fulfillment-service-owned fields on an existing spec, leaving operator-managed fields untouched:

- **Create path**: passes a fresh `&object.Spec` (zero struct — same behavior as before)
- **Update path**: passes `&update.Spec` from the `DeepCopy()` of the existing CR (preserves operator-managed fields)

This approach also prevents recurrence: new fulfillment-service fields added to `mutateSpec` automatically preserve operator-managed fields without needing a separate assignment block.

## Testing

- **Regression test added**: `Describe("update") / It("should preserve operator-managed spec fields when patching an existing CR")` — pre-populates a BMI CR with `ExternalHostID`, `HostClass`, `NetworkClass`, runs `t.update(ctx)`, and asserts all three fields survive the merge patch.
- **Regression test validated**: confirmed it fails without the fix and passes with it.
- **Existing tests updated**: all 8 `buildSpec` tests adapted to new `mutateSpec` signature — all pass.
- **Full unit suite**: 37/37 baremetalinstance tests pass; 64/65 suites pass across `internal/` (1 pre-existing failure in `internal/auth` unrelated to this change).
- **Cross-repo check**: verified the bmf-operator only sets its own fields individually and never touches fulfillment-service-owned fields — no reverse overwrite issue.

## Confidence

High — the fix is minimal, directly addresses the root cause, and is validated by a regression test that exercises the exact bug scenario.

## Risk Assessment

Low — only the BareMetalInstance reconciler is changed. Other controllers using `update.Spec = spec` were checked and are not affected (fulfillment-service owns their entire CRD spec).

Fixes [OSAC-1697](https://redhat.atlassian.net/browse/OSAC-1697)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved BareMetalInstance reconciliation so updates preserve operator-managed spec fields (including external identifiers and class settings) rather than unintentionally overwriting them.
* **New Features**
  * Improved template parameter generation for BareMetalInstance creation, including automatic population from SSH key and user-data secret when available, plus correct run-strategy mapping.
* **Refactor**
  * Streamlined create/patch behavior to apply desired spec changes safely with improved failure handling.
* **Tests**
  * Expanded coverage for spec mutation and BareMetalInstance update preservation scenarios.

### DIFF
--- a/internal/controllers/baremetalinstance/baremetalinstance_reconciler_function.go
+++ b/internal/controllers/baremetalinstance/baremetalinstance_reconciler_function.go
@@ -28,6 +28,7 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	clnt "sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
 	bmfov1alpha1 "github.com/osac-project/bare-metal-fulfillment-operator/api/v1alpha1"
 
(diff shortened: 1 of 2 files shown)


jhernand:
Given that you are going to use the "mutate" concept won't it make sense to use `controllerutil.CreateOrPatch` as well?

carbonin:
Yeah probably. I took inspiration from that pattern but only went half way :sweat_smile:

carbonin:
Retesting integration as it was an infra failure:
>Error: INSTALLATION FAILED: failed to perform \"Fetch\" on source: Get \"https://production.cloudfront.docker.com/registry-v2/docker/registry/v2/blobs/sha256/d7/d76a1d09d66ef26b514951c6f21f97e5943688a4cf0a3c45a3d0b1b72e11142d/data?Expires=1782221342&Signature=nj0cb0hh4rvKDlzLE2vkWY8kQkBvwmRktOzmh6BPljWeNkV9OhMxAdEHYPN9sMAhhNxv~qQQC-IsQhYTSjTt3hSBfd9pMFGnLNBfya2kXkNty9ccHbffneByijgIlisvuI4X8bU2Fpv6T8X~GsFs~REl4WIIFiSaSgm3DdEZhvQ6zAV6fWj5y~XV0O18m-v7aZQb4j3cX-HnGHVl~ZRpLjQcu2K7iCaMm5gcYhjvL1MuXGeyFzx5w5W7ZpqLi5pCuoAlaCDE5EtJ3kXQo7quWw-oUVOE6izIR6NEv0LQ0AigOtJWuF7S8zzg99JhjyEuAuRG67c4cfEUQ5O0laf4bg__&Key-Pair-Id=K2C9XPB6FLAKUF\": read tcp 10.1.0.160:44478->3.170.185.54:443: read: connection reset by peer

adriengentil (approved):

PR #743, opened 2026-06-23
NO-ISSUE: Increase probe timeouts and failure thresholds in Helm charts

jhernand:
## Summary

- Sets `timeoutSeconds: 5` and `failureThreshold: 5` on all liveness and readiness probes across every Helm chart deployment template (grpc-server, rest-gateway, controller, console-proxy, ingress-proxy).
- Under CPU/IO pressure on single-node clusters the exec-based probes can exceed the default 1s timeout, causing the headless Service to drop pod IPs and Envoy to return "no healthy upstream" errors.
- With these values baked into the chart, the [`patch-grpc-probes.yaml`](https://github.com/osac-project/osac-installer/blob/main/charts/osac/templates/hooks/patch-grpc-probes.yaml) post-install hook in osac-installer is no longer needed and can be removed.

## Test plan

- [ ] Deploy the chart on a single-node kind/OpenShift cluster and verify the probes use the new timeout and threshold values.
- [ ] Confirm the `osac-patch-grpc-probes` hook job can be removed from osac-installer without regressions.

## Summary by CodeRabbit

* **Chores**
  * Updated health check configurations across service deployments to adjust timeout and failure threshold parameters.

### DIFF
--- a/charts/service/templates/console-proxy/deployment.yaml
+++ b/charts/service/templates/console-proxy/deployment.yaml
@@ -101,6 +101,8 @@ spec:
             - --grpc-server-timeout=1s
           initialDelaySeconds: 5
           periodSeconds: 10
+          timeoutSeconds: 5
+          failureThreshold: 5
         readinessProbe:
           exec:
             command:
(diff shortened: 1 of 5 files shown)


jhernand:
@omer-vishlitzky I think when this is merged you can remove https://github.com/osac-project/osac-installer/blob/main/charts/osac/templates/hooks/patch-grpc-probes.yaml.

CrystalChun (approved):
/approve
/lgtm

PR #745, opened 2026-06-23
OSAC-1351:  bmi integration test

mennyaboush:
## Summary by CodeRabbit

* **Tests**
  * Added integration coverage for the BareMetalInstance lifecycle, including provisioning from a published catalog item, status transition to `RUNNING`, and verification of public metadata (creation timestamp present, deletion timestamp absent).
  * Ensures cleanup/deletion behavior is exercised after lifecycle actions.
  * Added validation that creating an instance with a non-existent catalog item reference fails as expected.

### DIFF
--- a/it/it_baremetal_instance_lifecycle_test.go
+++ b/it/it_baremetal_instance_lifecycle_test.go
@@ -1,0 +1,155 @@
+/*
+Copyright (c) 2026 Red Hat Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with the
+License. You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific
+language governing permissions and limitations under the License.
+*/
+
+package it
+
+import (
+	"context"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2/dsl/core"
+	. "github.com/onsi/gomega"
+	grpccodes "google.golang.org/grpc/codes"
+	grpcstatus "google.golang.org/grpc/status"
+	"google.golang.org/protobuf/types/known/fieldmaskpb"
+
+	privatev1 "github.com/osac-project/fulfillment-service/internal/api/osac/private/v1"
+	publicv1 "github.com/osac-project/fulfillment-service/internal/api/osac/public/v1"
+)
+
+var _ = Describe("BareMetalInstance lifecycle", func() {
+	var (
+		bareMetalInstancesClient            publicv1.BareMetalInstancesClient
+		privateBareMetalInstancesClient     privatev1.BareMetalInstancesClient
+		bareMetalInstanceTemplatesClient    privatev1.BareMetalInstanceTemplatesClient
+		bareMetalInstanceCatalogItemsClient privatev1.BareMetalInstanceCatalogItemsClient
+		templateId                          string
+		catalogItemId                       string
+	)
+
+	BeforeEach(func(ctx context.Context) {
+		// Create clients
+		bareMetalInstancesClient = publicv1.NewBareMetalInstancesClient(tool.ExternalView().UserConn())
+		privateBareMetalInstancesClient = privatev1.NewBareMetalInstancesClient(tool.InternalView().AdminConn())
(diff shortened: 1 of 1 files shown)


adriengentil:
can you add tests for the reconciler as well? You can get inspiration from https://github.com/osac-project/fulfillment-service/blob/main/it/it_cluster_reconciler_test.go

it/it_baremetal_instance_lifecycle_test.go
```diff
@@ -0,0 +1,178 @@
...
+	"google.golang.org/protobuf/types/known/fieldmaskpb"
+
+	privatev1 "github.com/osac-project/fulfillment-service/internal/api/osac/private/v1"
+	publicv1 "github.com/osac-project/fulfillment-service/internal/api/osac/public/v1"
+	"github.com/osac-project/fulfillment-service/internal/uuid"
+)
+
+var _ = Describe("BareMetalInstance lifecycle", func() {
+	var (
+		ctx                                 context.Context
```
jhernand: Try to avoid creating contexts like this, use the context passed by the Ginkgo framework if possible:

```go
It("Blah, blah ...", func(ctx context.Context) {
    ...
})
```

it/it_baremetal_instance_lifecycle_test.go
```diff
@@ -0,0 +1,178 @@
...
+				Id:        catalogItemId,
+				Title:     "Test BMI Catalog Item",
+				Template:  templateId,
+				Published: true,
+			}.Build(),
+		}.Build())
+		Expect(err).ToNot(HaveOccurred())
+	})
+
+	AfterEach(func() {
```
jhernand: Try to use `DeferCleanup` in the `BeforeEach` block instead of this.

it/it_baremetal_instance_lifecycle_test.go
```diff
@@ -0,0 +1,178 @@
...
+		// Clean up BareMetalInstance if created
+		if bareMetalInstanceId != "" {
+			bareMetalInstancesClient.Delete(ctx, publicv1.BareMetalInstancesDeleteRequest_builder{
+				Id: bareMetalInstanceId,
+			}.Build())
```
jhernand: Shouldn't ignore the errors returned by delete.

it/it_baremetal_instance_lifecycle_test.go
```diff
@@ -0,0 +1,178 @@
...
+		ctx = context.Background()
+
+		// Create clients
+		bareMetalInstancesClient = publicv1.NewBareMetalInstancesClient(tool.ExternalView().UserConn())
+		privateBareMetalInstancesClient = privatev1.NewBareMetalInstancesClient(tool.InternalView().AdminConn())
+		bareMetalInstanceTemplatesClient = privatev1.NewBareMetalInstanceTemplatesClient(tool.InternalView().AdminConn())
+		bareMetalInstanceCatalogItemsClient = privatev1.NewBareMetalInstanceCatalogItemsClient(tool.InternalView().AdminConn())
+
+		// Create BareMetalInstanceTemplate
+		templateId = fmt.Sprintf("test-bmi-template-%s", uuid.New())
```
jhernand: We are trying to move to unique names and let the system generate the UUID?

it/it_baremetal_instance_lifecycle_test.go
```diff
@@ -0,0 +1,178 @@
...
+
+		// Clean up Template
+		if templateId != "" {
+			bareMetalInstanceTemplatesClient.Delete(ctx, privatev1.BareMetalInstanceTemplatesDeleteRequest_builder{
+				Id: templateId,
+			}.Build())
+		}
+	})
+
+	It("creates a BareMetalInstance and verifies fields", func() {
```
jhernand: The descriptions of `It` blocks start with upper case, in general:

```go
It("Creates ....", func() {
        ....
})
```

it/it_baremetal_instance_lifecycle_test.go
```diff
@@ -0,0 +1,178 @@
...
+		createResp, err := bareMetalInstancesClient.Create(ctx, publicv1.BareMetalInstancesCreateRequest_builder{
+			Object: publicv1.BareMetalInstance_builder{
+				Id: bareMetalInstanceId,
+				Spec: publicv1.BareMetalInstanceSpec_builder{
+					CatalogItem: catalogItemId,
+				}.Build(),
+			}.Build(),
+		}.Build())
+		Expect(err).ToNot(HaveOccurred())
+		Expect(createResp.GetObject()).ToNot(BeNil())
```
jhernand: Consider using `defer` or `DeferCleanup(...)` to delete this after the test, so it is deleted even if some of the expectations fail.

jhernand (changes requested):

adriengentil (approved):

PR #747, opened 2026-06-23
NO-ISSUE: Add migration file list hash to prevent numbering collisions

jhernand:
Add a `migrations.sha256` file in `internal/database/` containing the SHA-256 hash of the sorted
list of `.up.sql` migration filenames. Its purpose is to cause a git merge conflict when two pull
requests independently introduce migrations with the same number.

A unit test in `database_migrations_test.go` verifies that the stored hash matches the current
migration file list. When the hash is outdated the test failure message points to the
`uv run dev.py update hashes` command, which recomputes and writes the hash automatically.

## Summary

* **New Features**
  * Added a development CLI command to regenerate the database migrations SHA-256 checksum (with a `hashes` helper).

* **Bug Fixes**
  * Added automated verification that the stored checksum matches the current set of migration files, failing tests when mismatched.
  * Updated the stored checksum value to the latest migrations.

* **Documentation**
  * Documented the migration workflow and how to update the checksum when migrations are added or changed.


ygalblum:
As long as we're using the merge queue that runs the E2E tests before merging, this error will be checked there anyhow. But, I do understand the advantage of having it here. 

I agree with this comment: https://github.com/osac-project/fulfillment-service/pull/747#discussion_r3461317338. You want to fail the test to ensure that the change to the README is not forgotten. Alternatively, just have it fail and print the expected value, forcing the user to update the value in the README. Yes, it's extra work for the user, but I'm not sure that tests with side effects are the right thing to do.

jhernand:
> As long as we're using the merge queue that runs the E2E tests before merging, this error will be checked there anyhow. But, I do understand the advantage of having it here.
> 
> I agree with this comment: [#747 (comment)](https://github.com/osac-project/fulfillment-service/pull/747#discussion_r3461317338). You want to fail the test to ensure that the change to the README is not forgotten. Alternatively, just have it fail and print the expected value, forcing the user to update the value in the README. Yes, it's extra work for the user, but I'm not sure that tests with side effects are the right thing to do.

The new version of the patch should address these concerns: the unit tests only verifies that the hash is correct, and suggest the user to either fix the file manually or else use the `uv run dev.py update hashes` command that updates it automatically.

internal/database/database_migrations_test.go
```diff
@@ -92,6 +95,40 @@ var _ = Describe("Migrations", func() {
...
+		Expect(err).ToNot(HaveOccurred())
+		Expect(files).ToNot(BeEmpty())
+		names := make([]string, len(files))
+		for i, file := range files {
+			names[i] = filepath.Base(file)
+		}
+		sort.Strings(names)
+
+		// Compute the SHA-256 hash of the migration file list:
+		compuredHashSource := &bytes.Buffer{}
```
ygalblum: Sorry for the nit:
```suggestion
		computedHashSource := &bytes.Buffer{}
```

internal/database/database_migrations_test.go
```diff
@@ -92,6 +95,40 @@ var _ = Describe("Migrations", func() {
...
+		for _, name := range names {
+			fmt.Fprintf(compuredHashSource, "%s\n", name)
+		}
+		computedHashBytes := sha256.Sum256(compuredHashSource.Bytes())
+		computedHashText := fmt.Sprintf("%x", computedHashBytes)
+
+		// Read the stored hash:
+		storedHashFile, err := filepath.Abs("migrations.sha256")
```
ygalblum: Since I'm already commenting, though reading the file is supposed to fail , you can fast fail by moving these lines to the beginning of the test

dev/update.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+@update.command(name="hashes")
+def hashes() -> None:
+    """
+    Updates the database migrations hash.
+    """
+    # Compute the hash of the migration files:
+    migrations_dir = dirs.project() / "internal" / "database" / "migrations"
+    migration_files = sorted(migrations_dir.glob("*.up.sql"))
```
ygalblum: Since I had to look back to find if and where the list was sorted, can you please rename:
```suggestion
    migration_files_sorted = sorted(migrations_dir.glob("*.up.sql"))
```

internal/database/database_migrations_test.go
jhernand: Done.

internal/database/database_migrations_test.go
jhernand: Done.

dev/update.py
jhernand: Done.

ygalblum (approved):
/approve
/lgtm

internal/database/database_migrations_test.go
```diff
@@ -92,6 +95,42 @@ var _ = Describe("Migrations", func() {
...
+		for i, file := range migrationFiles {
+			migrationNames[i] = filepath.Base(file)
+		}
+		sort.Strings(migrationNames)
+
+		// Compute the SHA-256 hash of the migration file list:
+		computedHashSource := &bytes.Buffer{}
+		for _, name := range migrationNames {
+			fmt.Fprintf(computedHashSource, "%s\n", name)
+		}
```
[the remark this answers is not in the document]
jhernand: Done.

ygalblum (commented):
/lgtm

omer-vishlitzky:

**Root cause:** Helm upgrade fails during the boot/refresh phase because of a strategic merge patch conflict on the osac-operator deployment's OSAC_AAP_TOKEN environment variable.

**Explanation:** Recently merged PR #328 in osac-installer configured the operator to load its AAP token from a secret via `secretKeyRef` (using the new `aap.tokenSecret` Helm value introduced in osac-operator PR #316). However, the pre-built cluster snapshot flavor (`vmaas-helm`) still contains the older deployment of `osac-operator` where `OSAC_AAP_TOKEN` is set as a literal `value`. During the snapshot boot refresh phase, `refresh-after-snapshot.py` executes `helm upgrade`, which attempts to patch the existing deployment. Because Kubernetes strategic merge patch merges the new `valueFrom` field into the existing environment variable element without clearing the existing `value` field, the resulting spec contains both `value` and `valueFrom` for `OSAC_AAP_TOKEN`, which is invalid and rejected by the API server.

**Evidence:**

[`build-log.txt`](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/osac-project_fulfillment-service/747/pull-ci-osac-project-fulfillment-service-main-e2e-vmaas/2069866775721283584/artifacts/e2e-vmaas/osac-project-gather/artifacts/osac-logs/build-log.txt):
```
Error: UPGRADE FAILED: cannot patch "osac-operator" with kind Deployment: Deployment.apps "osac-operator" is invalid: spec.template.spec.containers[0].env[1].valueFrom: Invalid value: "": may not be specified when `value` is not empty
```

[`build-log.txt`](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/osac-project_fulfillment-service/747/pull-ci-osac-project-fulfillment-service-main-e2e-vmaas/2069866775721283584/artifacts/e2e-vmaas/osac-project-gather/artifacts/osac-logs/build-log.txt):
```
ERROR: command failed (exit 1): helm upgrade osac charts/osac/ --namespace osac-e2e-ci --values values/vmaas-ci/values.yaml --set service.externalHostname=fulfillment-api-osac-e2e-ci.apps.test-infra-cluster-vmaas-helm.redhat.com --set service.internalHostname=fulfillment-internal-api-osac-e2e-ci.apps.test-infra-cluster-vmaas-helm.redhat.com --set aap.bootstrap.enabled=false --set clusterFulfillment.config.HOSTED_CLUSTER_BASE_DOMAIN=hosted.test-infra-cluster-vmaas-helm.redhat.com --timeout 15m
```

**Suggestion:** To fix this, either: (1) update the osac-operator Helm chart deployment template to explicitly set `value: null` when `aap.tokenSecret` is used, which will clear the existing literal value during strategic merge patch, or (2) modify `refresh-after-snapshot.py` to delete the existing `osac-operator` deployment before running `helm upgrade` so it is recreated cleanly from scratch.

---
[Prow job](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/osac-project_fulfillment-service/747/pull-ci-osac-project-fulfillment-service-main-e2e-vmaas/2069866775721283584) | Build `2069866775721283584` | 🤖 *triagent*

PR #750, opened 2026-06-23
NO-ISSUE: add is_windows field to ComputeInstanceTemplate and CatalogItem support

ygalblum:
Add `is_windows` optional boolean field to `ComputeInstanceTemplateSpecDefaults` in both public and private proto APIs, with template default application and CatalogItem field definition support.

Windows VM provisioning requires templates to carry a guest OS indicator so that downstream provisioning (osac-aap) can apply Windows-specific configuration. This extends the template defaults mechanism to propagate `is_windows` through the same path as other spec fields like cores, memory, and run_strategy.

- Added `optional bool is_windows` field (number 6) to `ComputeInstanceTemplateSpecDefaults` in both public and private proto definitions
- Regenerated protobuf Go code for the new field
- Extended `ApplySpecDefaults` to propagate the `is_windows` template default when the user has not set it
- Added `is_windows` as a recognized path in CatalogItem field definition application (`applyFieldDefinitions`)

- Added unit tests for template default application: applies default, preserves user-provided value, handles missing default
- Added unit tests for CatalogItem field definitions: applies editable default, forces non-editable value

## Summary by CodeRabbit

* **New Features**
  * Template defaults now include an optional `is_windows` setting to choose the VM operating system (Windows when `true`, Linux when `false` or omitted). New instances inherit this value automatically unless explicitly overridden.
* **Bug Fixes**
  * Improved defaulting behavior to ensure `is_windows` is applied correctly from template defaults when the user does not set it.
* **Tests**
  * Added coverage for defaulting, preservation when the user provides a value, and override behavior based on field editability.


omer-vishlitzky:

**Root cause:** The main branch is broken due to a database migration sequence number collision between PR 706 and PR 740, which both added a migration with sequence number 60.

**Explanation:** The boot step failed because the fulfillment-console-proxy deployment timed out during rollout. The proxy depends on the fulfillment-grpc-server, which was crashing at startup. The grpc-server pod logs show it failed to initialize the database with the error 'failed to init driver with path migrations: duplicate migration file: 60_create_external_ip_tables.up.sql'. A search of recently merged PRs reveals that PR 706 added '60_add_projects_immutable_trigger.up.sql' and PR 740 added '60_create_external_ip_tables.up.sql'. Because PR 740 was merged after PR 706 without rebasing, the main branch now contains two migrations with the '60_' prefix, causing the golang-migrate library to fail. This PR (750) does not touch migrations and is failing because it was tested against the broken main branch.

**Evidence:**

[`e2e-vmaas-osac-project-cluster-tool-boot build-log.txt`](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/osac-project_fulfillment-service/750/pull-ci-osac-project-fulfillment-service-main-e2e-vmaas/2069563846242078720/artifacts/e2e-vmaas/osac-project-gather/artifacts/osac-logs/e2e-vmaas-osac-project-cluster-tool-boot build-log.txt):
```
ERROR: command failed (exit 1): oc rollout status deploy/fulfillment-console-proxy -n osac-e2e-ci --timeout=360s
```

[`pod-fulfillment-grpc-server-59dfff497b-ksfrl-grpc-server.log`](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/osac-project_fulfillment-service/750/pull-ci-osac-project-fulfillment-service-main-e2e-vmaas/2069563846242078720/artifacts/e2e-vmaas/osac-project-gather/artifacts/osac-logs/pod-fulfillment-grpc-server-59dfff497b-ksfrl-grpc-server.log):
```
failed to init driver with path migrations: duplicate migration file: 60_create_external_ip_tables.up.sql
```

**Suggestion:** Create a PR to fix the main branch by renaming the migration files from '60_create_external_ip_tables' to '61_create_external_ip_tables'. Once merged, rebase this PR and retrigger the tests.

---
[Prow job](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/osac-project_fulfillment-service/750/pull-ci-osac-project-fulfillment-service-main-e2e-vmaas/2069563846242078720) | Build `2069563846242078720` | 🤖 *triagent*

omer-vishlitzky:

**Root cause:** Helm upgrade fails during the boot step because osac-installer PR 328 changed the OSAC_AAP_TOKEN env var from value to valueFrom, causing a strategic merge patch conflict with the existing snapshot deployment.

**Explanation:** The CI boot step uses a pre-built snapshot flavor (`vmaas-helm`) which contains an older version of the `osac-operator` deployment where `OSAC_AAP_TOKEN` is set as a literal `value`. Recently, `osac-installer` PR 328 updated the CI Helm values to use `aap.tokenSecret`, which renders `valueFrom` instead of `value`. When `refresh-after-snapshot.py` runs `helm upgrade` to apply the new configuration, Helm generates a strategic merge patch that adds `valueFrom` but does not explicitly remove `value`. Kubernetes rejects the patch because a container environment variable cannot have both `value` and `valueFrom`. This breaks the boot step for all PRs across all repos.

**Evidence:**

[`build-log.txt`](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/osac-project_fulfillment-service/750/pull-ci-osac-project-fulfillment-service-main-e2e-vmaas/2069846868132630528/artifacts/e2e-vmaas/osac-project-gather/artifacts/osac-logs/build-log.txt):
```
Error: UPGRADE FAILED: cannot patch "osac-operator" with kind Deployment: Deployment.apps "osac-operator" is invalid: spec.template.spec.containers[0].env[1].valueFrom: Invalid value: "": may not be specified when `value` is not empty
```

[`build-log.txt`](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/osac-project_fulfillment-service/750/pull-ci-osac-project-fulfillment-service-main-e2e-vmaas/2069846868132630528/artifacts/e2e-vmaas/osac-project-gather/artifacts/osac-logs/build-log.txt):
```
ERROR: command failed (exit 1): helm upgrade osac charts/osac/ --namespace osac-e2e-ci --values values/vmaas-ci/values.yaml --set service.externalHostname=fulfillment-api-osac-e2e-ci.apps.test-infra-cluster-vmaas-helm.redhat.com --set service.internalHostname=fulfillment-internal-api-osac-e2e-ci.apps.test-infra-cluster-vmaas-helm.redhat.com --set aap.bootstrap.enabled=false --set clusterFulfillment.config.HOSTED_CLUSTER_BASE_DOMAIN=hosted.test-infra-cluster-vmaas-helm.redhat.com --timeout 15m
```

**Suggestion:** Update the `osac-operator` Helm chart (`charts/operator/templates/deployment.yaml`) to explicitly set `value: null` when `valueFrom` is used for `OSAC_AAP_TOKEN`. This will force Helm to generate a patch that removes the old `value` field during upgrades.

---
[Prow job](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/osac-project_fulfillment-service/750/pull-ci-osac-project-fulfillment-service-main-e2e-vmaas/2069846868132630528) | Build `2069846868132630528` | 🤖 *triagent*

proto/public/osac/public/v1/compute_instance_template_type.proto
```diff
@@ -115,4 +115,7 @@ message ComputeInstanceTemplateSpecDefaults {
 
   // Default instance type name. Mutually exclusive with cores/memory_gib defaults.
   optional string instance_type = 6;
+
```
tzvatot, line 118: 🔴 **`optional bool is_windows` is the wrong type for a guest OS field**

A boolean encodes a false dichotomy: "it's either Windows or it isn't." This locks the API into exactly two OS families forever. When you need to support FreeBSD, a different Windows variant that needs distinct provisioning (Server 2019 vs 2022), or any other OS family, you're stuck - you'd need a new field and a deprecation cycle on a public API surface.

Model it as what it is - a type/category:

```proto
// Guest OS family. Determines OS-specific provisioning configuration.
optional string guest_os = 7;  // e.g., "linux", "windows"
```

Or with an enum for compile-time safety:

```proto
enum GuestOSFamily {
  GUEST_OS_FAMILY_UNSPECIFIED = 0;
  GUEST_OS_FAMILY_LINUX = 1;
  GUEST_OS_FAMILY_WINDOWS = 2;
}
optional GuestOSFamily guest_os_family = 7;
```

This is how OpenShift Virtualization handles it. KubeVirt uses typed OS identifiers throughout - `VirtualMachinePreference` objects carry OS-specific configuration indexed by named OS types (e.g., `windows.2k22`, `fedora`, `rhel.9`), and the `vm.kubevirt.io/os` annotation takes string values, not booleans. See the [OpenShift Virtualization docs on instance types and preferences](https://docs.openshift.com/container-platform/latest/virt/virtual_machines/creating_vms_custom/virt-creating-vms-from-instance-types.html).

A string or enum field is equally simple to implement, costs nothing extra, and keeps the API extensible. The bool would be permanent debt from day one.

Note: `ComputeInstanceSpec` appears to already have `is_windows` as a bool (referenced in the test code). If so, that should be fixed in the same pass - better to fix it now while the field is fresh than to accumulate more consumers of the wrong type.

proto/private/osac/private/v1/compute_instance_template_type.proto
```diff
@@ -62,4 +62,7 @@ message ComputeInstanceTemplateSpecDefaults {
 
   // Default instance type name. Mutually exclusive with cores/memory_gib defaults.
   optional string instance_type = 6;
+
```
tzvatot, line 65: 🔴 Same issue as the public proto - replace `optional bool is_windows` with a typed field like `optional string guest_os` or a `GuestOSFamily` enum. See the comment on the public proto for the full rationale.

tzvatot (changes requested):
The implementation mechanics (default propagation, field definitions, tests) are well done and follow the existing patterns. But the field type is a fundamental API design issue that should be fixed before merge.

| Category | Count |
|----------|-------|
| 🔴 Critical | 1 |
| 💡 Suggestion | 1 |

**💡 Suggestion:** Even setting the type aside, the field name `is_windows` bakes a specific value into the schema. A name like `guest_os` or `os_family` describes the dimension, not one point on it. `guest_os: "windows"` reads as a choice; `is_windows: true` reads as an assertion about one specific OS.

proto/public/osac/public/v1/compute_instance_template_type.proto
ygalblum: This field is temporary for v0.1 anyhow as it will be part of the ComputeImage once it's introduced. At this stage we only care if the image is windows or not while keeping the default as linux. As a result, we decided that at the Fulfillment Service level the API will be boolean.  Also, see that this PR is about the Template while the `ComputeInstance` already defines this field as a boolean.

proto/private/osac/private/v1/compute_instance_template_type.proto
ygalblum: Same answer. Resolving this comment since it's duplicated

proto/public/osac/public/v1/compute_instance_template_type.proto
tzvatot: Fair enough. If it's temporary and ComputeInstance already uses the same boolean, makes sense to keep them consistent. Just make sure there's a ticket tracking the migration to ComputeImage so it doesn't quietly become permanent.

ygalblum: There's a whole feature about Compute Images. It's not broken into tasks yet. But, once it is, this field will be removed both here and in the `ComputeInstance` itself

tzvatot: SGTM, thanks

PR #761, opened 2026-06-24
NO-ISSUE: Remove duplicated migration

jhernand:
## Summary by CodeRabbit

* **Tests**
  * Updated migration test coverage to validate the latest database schema and constraint behavior against migration version 61.


ori-amizur:
@jhernand why was the file 60_create_external_ip_tables.down.sql‎ removed  and not renamed?

jhernand:
> @jhernand why was the file 60_create_external_ip_tables.down.sql‎ removed ?

Because we have no way to apply ".down.sql" files.

ori-amizur:
> Because we have no way to apply ".down.sql" files.

In case we have problems, we can always apply them manually.

jhernand:
> In case we have problems, we can always apply them manually.

Restored the file. Please review again.

PR #763, opened 2026-06-24
OSAC-1679: Add GitHub Actions e2e job

omer-vishlitzky:
## Summary
- Add `e2e-test.yml` caller workflow for PR-triggered E2E testing via GitHub Actions
- Builds PR image on `ubuntu-latest`, pushes to GHCR with `pr-<number>` tag
- Calls shared `e2e.yml` from `osac-test-infra` with component image override
- Supports `workflow_dispatch` for manual execution

**Depends on:** osac-project/osac-test-infra PR ([OSAC-1678](https://redhat.atlassian.net/browse/OSAC-1678)) for the `component-image` input on `e2e.yml`

## Jira
- [OSAC-1679](https://redhat.atlassian.net/browse/OSAC-1679)

## Test plan
- [ ] Verify workflow triggers on PR to main
- [ ] Verify PR image builds and pushes to GHCR
- [ ] Verify e2e workflow is invoked with correct component-image override

### DIFF
--- a/.github/workflows/e2e-test.yml
+++ b/.github/workflows/e2e-test.yml
@@ -1,0 +1,38 @@
+---
+name: E2E Tests
+
+on:
+  pull_request:
+    branches: [main]
+  workflow_dispatch: {}
+
+jobs:
+  build-pr-image:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+    outputs:
+      image: ${{ steps.meta.outputs.image }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Set image metadata
+        id: meta
+        run: |
+          IMAGE="ghcr.io/osac-project/fulfillment-service:pr-${{ github.event.pull_request.number || github.run_id }}"
+          echo "image=${IMAGE}" >> "$GITHUB_OUTPUT"
+
+      - name: Build and push PR image
+        run: |
+          echo "${{ secrets.GITHUB_TOKEN }}" | podman login ghcr.io -u "${{ github.actor }}" --password-stdin
+          podman build -t "${{ steps.meta.outputs.image }}" -f Containerfile .
+          podman push "${{ steps.meta.outputs.image }}"
+
+  e2e:
+    needs: build-pr-image
+    uses: osac-project/osac-test-infra/.github/workflows/e2e.yml@main
+    with:
+      component-image: ${{ needs.build-pr-image.outputs.image }}
+      component-image-name: ghcr.io/osac-project/fulfillment-service


.github/workflows/e2e-test.yml
```diff
@@ -0,0 +1,38 @@
...
+        id: meta
+        run: |
+          IMAGE="ghcr.io/osac-project/fulfillment-service:pr-${{ github.event.pull_request.number || github.run_id }}"
+          echo "image=${IMAGE}" >> "$GITHUB_OUTPUT"
+
+      - name: Build and push PR image
+        run: |
+          echo "${{ secrets.GITHUB_TOKEN }}" | podman login ghcr.io -u "${{ github.actor }}" --password-stdin
+          podman build -t "${{ steps.meta.outputs.image }}" -f Containerfile .
+          podman push "${{ steps.meta.outputs.image }}"
```
jhernand, line 31: There is an existing job that builds and pushes the image here: https://github.com/osac-project/fulfillment-service/blob/main/.github/workflows/publish-image.yaml . Can we change that so that it builds the image for pull requests, and then make this job use that image, so that if we need to build the image differently in the future we will not have to change it in two places?

.github/workflows/e2e-test.yml
```diff
@@ -0,0 +1,38 @@
...
+  build-pr-image:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+    outputs:
+      image: ${{ steps.meta.outputs.image }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
```
jhernand, line 19: In all other workflows of the project we pin actions using hashes:

```
uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10  # v6
```

ygalblum: Why are we pinning by hash? Shouldn't we just use the latest version? And BTW the latest is now `v7`

.github/workflows/e2e-test.yml
ygalblum: +1 on this. Especially since we push to GH's registry, we should the existing actions (and in this case even the existing workflows)

.github/workflows/e2e-test.yml
```diff
@@ -0,0 +1,38 @@
...
+          echo "${{ secrets.GITHUB_TOKEN }}" | podman login ghcr.io -u "${{ github.actor }}" --password-stdin
+          podman build -t "${{ steps.meta.outputs.image }}" -f Containerfile .
+          podman push "${{ steps.meta.outputs.image }}"
+
+  e2e:
+    needs: build-pr-image
+    uses: osac-project/osac-test-infra/.github/workflows/e2e.yml@main
+    with:
+      component-image: ${{ needs.build-pr-image.outputs.image }}
+      component-image-name: ghcr.io/osac-project/fulfillment-service
```
ygalblum, line 38: I haven't looked at this workflow. But, why does it need both? It seems that `ghcr.io/osac-project/fulfillment-service` is written multiple times. Also about this, should it be defined once somewhere?

ygalblum (commented):
IIUC this workflow tests the code on the PR branch. So, it does not replace the test that rebases these changes on top of main. Right?

.github/workflows/e2e-test.yml
jhernand: Using tags, or latest, is flagged by @coderabbitai as a supply chain security risk.

ygalblum: I might agree with `latest`. But, for actions this means that for every bug fix the action releases, we need to update all our actions. Why should we care? Moreover, because the the SHA is pinned, dependabot did not suggest updating to a new version. So, in a while we'll be using old version (BTW the SHA you suggested is probably not `v6` anymore).

I think @coderabbitai is taking a valid concern (mainly pinning container images to SHA and not a version) and applying it where it isn't. 

And as for:
> That also matches the existing workflow convention in this repository

This is a circular argument.

jhernand: Dependabot does suggest updates to actions, even if they are pinned, for example: https://github.com/osac-project/fulfillment-service/pull/771/changes .

ygalblum: If that's the case, then we're good. The fact that it was still pointing to v6 confused me. I stand corrected

.github/workflows/e2e-test.yml
```diff
@@ -0,0 +1,38 @@
+---
+name: E2E Tests
+
+on:
+  pull_request:
+    branches: [main]
+  workflow_dispatch: {}
```
tzvatot, line 7: 🟡 **Missing concurrency control**

Unlike `publish-image.yaml`, this workflow has no `concurrency` block. Multiple pushes to the same PR will trigger parallel E2E runs (build + full cluster test), wasting expensive CI resources. Add:

```yaml
concurrency:
  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.sha }}
  cancel-in-progress: true
```

.github/workflows/e2e-test.yml
```diff
@@ -0,0 +1,38 @@
...
+    outputs:
+      image: ${{ steps.meta.outputs.image }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Set image metadata
+        id: meta
+        run: |
+          IMAGE="ghcr.io/osac-project/fulfillment-service:pr-${{ github.event.pull_request.number || github.run_id }}"
```
tzvatot, line 24: 💡 **No PR image cleanup strategy**

PR images (`pr-<number>`) will accumulate in GHCR indefinitely. Consider a cleanup workflow or retention policy to prune images from closed/merged PRs.

tzvatot (changes requested):
@jhernand  and @ygalblum already covered the main structural concerns (reuse `publish-image.yaml`, pin actions by SHA, deduplicate image path). Two additional items below.

| Category | Count |
|----------|-------|
| 🟡 Important | 1 |
| 💡 Suggestion | 1 |

PR #776, opened 2026-06-25
OSAC-1809: enforce timestamp-based naming for new database migrations

eranco74:
## Summary

- Enforce timestamp-based naming (`YYYYMMDDHHMMSS_description.up.sql`) for all new migrations (number > 63) to eliminate migration number collisions between concurrent PRs
- Keep existing sequential migrations (0–63) as-is since deployed databases track versions in `schema_migrations`
- Remove the gap-detection test (not applicable for timestamp naming)
- Add migrations README documenting the convention and usage

## Motivation

With sequential numbering, two developers independently adding migrations will both pick the next number (e.g. `64`). This requires one of them to renumber after the other's PR merges. Timestamp prefixes (`YYYYMMDDHHMMSS`) make collisions practically impossible — two developers would need to generate the prefix at the exact same second. This is the same approach used by [assisted-service](https://github.com/openshift/assisted-service/tree/master/internal/migrations).

## Test plan

- [x] All existing migration tests pass (suffix, duplicates, naming convention, hash)
- [ ] Verify a new migration file with sequential naming (e.g. `64_foo.up.sql`) fails the naming convention test
- [ ] Verify a new migration file with timestamp naming (e.g. `20260625143022_foo.up.sql`) passes

Jira: [OSAC-1809](https://redhat.atlassian.net/browse/OSAC-1809)

## Summary by CodeRabbit

* **Documentation**
  * Clarified database migration naming rules and added step-by-step guidance for creating new migration files.
  * Documented both legacy sequential and newer timestamp-based migration formats.

* **Tests**
  * Strengthened migration filename validation with more detailed error reporting.
  * Removed the check for gaps in the sequential migration number range.

### DIFF
--- a/internal/database/database_migrations_test.go
+++ b/internal/database/database_migrations_test.go
@@ -87,7 +87,7 @@ var _ = Describe("Migrations", func() {
 		legacyPattern := regexp.MustCompile(`^(\d{1,4})_[a-z][a-z0-9_]*[a-z0-9]\.(up|down)\.sql$`)
 		timestampPattern := regexp.MustCompile(`^(\d{14})_[a-z][a-z0-9_]*[a-z0-9]\.(up|down)\.sql$`)
 
-		const lastSequentialMigration = 63
+		const lastSequentialMigration = 76
 
 		var violations []string
 		for _, file := range files {
--- a/internal/database/database_migrations_test.go
+++ b/internal/database/database_migrations_test.go
@@ -84,15 +84,40 @@ var _ = Describe("Migrations", func() {
 		Expect(err).ToNot(HaveOccurred())
 		Expect(files).ToNot(BeEmpty())
 
-		pattern := regexp.MustCompile(`^(\d+)_[a-z][a-z0-9_]*[a-z0-9]\.(up|down)\.sql$`)
+		legacyPattern := regexp.MustCompile(`^(\d{1,4})_[a-z][a-z0-9_]*[a-z0-9]\.(up|down)\.sql$`)
+		timestampPattern := regexp.MustCompile(`^(\d{14})_[a-z][a-z0-9_]*[a-z0-9]\.(up|down)\.sql$`)
+
+		const lastSequentialMigration = 63
+
 		var violations []string
 		for _, file := range files {
 			base := filepath.Base(file)
-			if !pattern.MatchString(base) {
-				violations = append(violations, base)
+
+			parts := strings.SplitN(base, "_", 2)
+			if len(parts) < 2 {
+				violations = append(violations, fmt.Sprintf("%s: cannot parse prefix", base))
+				continue
+			}
+			n, err := strconv.Atoi(parts[0])
+			if err != nil {
+				violations = append(violations, fmt.Sprintf("%s: prefix is not a number", base))
+				continue
+			}
+
+			if n <= lastSequentialMigration {
+				if !legacyPattern.MatchString(base) {
+					violations = append(violations, fmt.Sprintf("%s: legacy migration must match N_description.(up|down).sql", base))
+				}
+			} else {
+				if !timestampPattern.MatchString(base) {
+					violations = append(violations, fmt.Sprintf(
+						"%s: new migrations (> %d) must use timestamp format YYYYMMDDHHMMSS_description.(up|down).sql — "+
+							"generate with: date +%%Y%%m%%d%%H%%M%%S",
+						base, lastSequentialMigration))
+				}
 			}
 		}
-		Expect(violations).To(BeEmpty(), "migration filenames violate naming convention: %v", violations)
+		Expect(violations).To(BeEmpty(), "migration filenames violate naming convention:\n%s", strings.Join(violations, "\n"))
 	})
 
 	It("Has an up-to-date migrations hash", func() {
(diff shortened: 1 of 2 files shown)


eranco74:
@jhernand PTAL

jhernand (changes requested):
We already have a mechanism to prevent this, see https://github.com/osac-project/fulfillment-service/pull/747. In addition time stamps are harder to handle for humans, with no benefit: when there is conflict users still need to rename migrations so that they are in the right order.

ygalblum:
While this solves the numbering conflict, but it does not solve the issue the numbers were there for to being with. 

The reason for the numbers is to make sure that the order is kept so that conflicting integrations will not occur. The timestamp makes sure the order is maintained when the code is already in `main`. But, the problem we have with numbering is conflicting PRs.

So let's say PR A is opened first with timestamp X. Then PR B is opened with timestamp Y (Y > X). But, they are both tested separately on their respective branch on top of the current one. 
- If PR A is merged, nothing stops PR B from merging and maybe overriding the migration made by A.
- If PR B is merged first. This means that now PR A is inserting a migration between the previous main and the one added by PR B without having to test it. That can create a mess.

In any case, not sure we're better off.

jhernand:
Here you can see how the current hash-based solution detects the issue and blocks merging:

image

eranco74:
@jhernand @ygalblum 
Re: timestamp vs hash approach for migration collisions

The hash file and timestamps solve different sides of the same problem:

- Hash file (PR #747): Detects collisions after they happen - forces a merge conflict so the second dev must rebase and renumber. The collision still occurs, it's just caught earlier.
- Timestamps (this PR): Prevents collisions from happening in the first place - two devs independently adding migrations will naturally get different prefixes. No conflict, no renumbering, no retest cycle.

Currently, we are still using `strict: true` , the real benefit of timestamps is reducing friction: devs don't have to coordinate on "what's the next  number" or deal with renumbering when someone else's migration lands first.

The tradeoff is readability - 54_add_column is easier to scan than 20260625143022_add_column. But this is the standard approach we use in assisted-service, precisely because sequential numbering doesn't scale when multiple people are adding migrations concurrently.

jhernand:
Timestamps don't solve the main problem: migrations need to be strictly ordered. There is the obvious case where a migration changes the database structure in such a way that the other will catastrophically fail. That is easy to detect, with unit tests. But there are also more subtle cases where you end up with corrupted data. For example, imagine that you have developer A working on changing all `192.168.1.0/24` subnet addresses to `10.0.1.0/24`, and they devise something like this:

```sql
-- 20260629100432_a.up.sql
update subnets set cidr = replace(cidr, '192.168.1.', '10.0.1.') where cidr like '%/24';
```

And developer B is working on changing the block size from '/24' to '/22', so they devise this:

```sql
-- 20260629100431_a.up.sql
update subnets set cidr = replace(cidr, '/24', '/22') where cidr like '192.168.1.%';
```

If you had `192.168.1.0/24` in the database and run first migration A and then migration B the result will be `10.0.1.0/22`, which is the desired end state. But if you run in the opposite order, the order implied by the timestamps, the result will be `192.168.1.0/22` which is wrong and silently applied. Using timestamps without any other control would have let these two apparently independent migrations merge, and then they will be applied in the wrong order, corrupting the database.

There is also the issue, caused by the mechanism that we use to run migrations, that if you go backwards in migration order, then intermediate migrations may be ignored. For example, imagine you had `20260629100432_a.up.sql` already merged. If you merge `20260629100431_a.up.sql` (earlier timestamp) it will be applied in systems that didn't apply `20260629100432_a.up.sql` yet, but silently ignored in systems that already applied it.

Using numeric indexes and the hashes mechanism makes this a little less likely. Developers can still just rebase, ignore or rename previous migrations, and continue, but at least they are given the opportunity to check and reflect.

ygalblum:
@jhernand thanks for giving an elaborate example to what I guess I wasn't clear enough. Using timestamps set by the developer during development may cause out of order merges which will cause real time issues later on.

eranco74:
I agree the out-of-order scenario is a real concern in theory. But in practice, when the hash check fails today, the developer just rebases, renumbers to the next available number, runs uv run dev.py update hashes, and pushes. Nobody is actually opening the other developer's migration to check for semantic conflicts - it's a mechanical step, not a review step.

The SQL example (CIDR + block size) is a legitimate risk, but neither approach actually catches it. The only way to detect semantic conflicts between migrations is through integration tests or migration review, not through a naming convention or a hash file.

So the tradeoff is: keep a friction point that doesn't deliver its intended benefit (inspection), or remove it and get the practical benefit of no collisions. I'd rather optimize for the **common case** (independent migrations that don't conflict) and rely on tests for the rare case (semantic conflicts).

Open to other thoughts though.

ygalblum:
> The only way to detect semantic conflicts between migrations is through integration tests or migration review, not through a naming convention or a hash file.

The current approach forces the user to rebase and rebasing helps with this issue because it will rerun the tests on top of the new main. Your approach allows merging without rebasing. So, you will be skipping that.

A simple example that I encountered was the change from `tenants` to `organizations`. My change, at the time the PR was opened, used the `tenants` table. But, before it was merged the table changed its name to `organizations`.  If I hadn't rebase, my code would have gotten merged and broken main.

PR #782, opened 2026-06-25
OSAC-1812: Auto create IdP user

CrystalChun:
# Description
When an IdP user first logs in, we should create the user in our database for persistent records. This creates a new
interceptor to handle this case.

# Testing

- Created Org, IdP connection, logged in as IdP user

## Summary by CodeRabbit

* **New Features**
  * Added just-in-time user provisioning for gRPC requests, automatically creating users when an authenticated request identifies a single eligible tenant.
* **Bug Fixes**
  * Updated authorization to store the subject in request context before emitting access-grant logging.
  * Improved gRPC startup reliability by initializing tenancy logic earlier and wiring the JIT provisioning interceptor into unary and streaming request chains.
  * Enhanced table rendering fallback: missing or unloadable table definitions now safely use the default layout.
* **Tests**
  * Added comprehensive coverage for JIT provisioning and user provisioning behavior.

### DIFF
--- a/internal/auth/grpc_authz_interceptor.go
+++ b/internal/auth/grpc_authz_interceptor.go
@@ -362,12 +362,14 @@ func (i *GrpcAuthzInterceptor) authorizeWithToken(ctx context.Context, method st
 		return
 	}
 
+	// Store subject in context
+	result = ContextWithSubject(ctx, subject)
+
 	logger.DebugContext(
-		ctx,
+		result,
 		"Permission granted by authorization policy",
 		slog.String("user", subject.User),
 	)
-	result = ContextWithSubject(ctx, subject)
 	return
 }
 
--- a/internal/auth/grpc_jit_provisioning_interceptor.go
+++ b/internal/auth/grpc_jit_provisioning_interceptor.go
@@ -1,0 +1,159 @@
+/*
+Copyright (c) 2026 Red Hat Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with the
+License. You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific
+language governing permissions and limitations under the License.
+*/
+
+package auth
+
+import (
+	"context"
+	"fmt"
+	"log/slog"
+
+	"github.com/golang-jwt/jwt/v5"
+	"google.golang.org/grpc"
+
+	"github.com/osac-project/fulfillment-service/internal/reflection"
+)
+
+//go:generate mockgen -source=$GOFILE -destination=grpc_jit_provisioning_interceptor_mock.go -package=$GOPACKAGE UserProvisioner
+
+// UserProvisioner provisions user records in the database.
+type UserProvisioner interface {
+	Provision(ctx context.Context, username, tenant string, claims jwt.MapClaims) error
+}
+
+// GrpcJitProvisioningInterceptorBuilder builds a GrpcJitProvisioningInterceptor.
+type GrpcJitProvisioningInterceptorBuilder struct {
+	logger      *slog.Logger
+	provisioner UserProvisioner
+}
+
+// GrpcJitProvisioningInterceptor performs just-in-time user provisioning for authenticated requests.
+// It runs after authentication and authorization, provisioning users who have been granted access.
+type GrpcJitProvisioningInterceptor struct {
+	logger      *slog.Logger
+	provisioner UserProvisioner
+}
+
+// NewGrpcJitProvisioningInterceptor creates a new builder.
+func NewGrpcJitProvisioningInterceptor() *GrpcJitProvisioningInterceptorBuilder {
+	return &GrpcJitProvisioningInterceptorBuilder{}
+}
+
+// SetLogger sets the logger.
+func (b *GrpcJitProvisioningInterceptorBuilder) SetLogger(value *slog.Logger) *GrpcJitProvisioningInterceptorBuilder {
+	b.logger = value
+	return b
+}
+
+// SetProvisioner sets the user provisioner.
+func (b *GrpcJitProvisioningInterceptorBuilder) SetProvisioner(value UserProvisioner) *GrpcJitProvisioningInterceptorBuilder {
+	b.provisioner = reflection.NormalizeNil(value)
+	return b
+}
+
+// Build creates the interceptor.
+func (b *GrpcJitProvisioningInterceptorBuilder) Build() (result *GrpcJitProvisioningInterceptor, err error) {
+	if b.logger == nil {
+		return nil, fmt.Errorf("logger is mandatory")
+	}
+	result = &GrpcJitProvisioningInterceptor{
+		logger:      b.logger,
+		provisioner: b.provisioner,
+	}
+	return result, nil
+}
+
+// UnaryServer is the unary server interceptor function.
+func (i *GrpcJitProvisioningInterceptor) UnaryServer(ctx context.Context, req any,
+	info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (any, error) {
+	err := i.provisionIfNeeded(ctx)
+	if err != nil {
+		return nil, err
+	}
+	return handler(ctx, req)
+}
+
+// StreamServer is the stream server interceptor function.
+func (i *GrpcJitProvisioningInterceptor) StreamServer(srv any, stream grpc.ServerStream,
+	info *grpc.StreamServerInfo, handler grpc.StreamHandler) error {
+	err := i.provisionIfNeeded(stream.Context())
+	if err != nil {
+		return err
+	}
+	return handler(srv, stream)
+}
+
+// provisionIfNeeded checks if user provisioning is needed and performs it.
+// Returns nil if provisioning is not needed or succeeds, error if provisioning fails.
+func (i *GrpcJitProvisioningInterceptor) provisionIfNeeded(ctx context.Context) error {
+	// Skip if no provisioner configured
+	if i.provisioner == nil {
+		return nil
+	}
+
+	// Get subject from context (set by authz interceptor)
+	subject := SubjectFromContext(ctx)
+	if subject == nil {
+		// No subject means anonymous or unauthenticated request - skip provisioning
+		return nil
+	}
+
+	// Only provision for users with exactly one tenant
+	// Users with zero tenants (unauthenticated), multiple tenants (cloud provider admins),
+	// or special tenants (system/shared) should not have user records
+	if !subject.Tenants.Finite() {
+		return nil
+	}
+
+	tenants := subject.Tenants.Inclusions()
+	if len(tenants) != 1 {
+		// Zero or multiple tenants - not a regular user
+		return nil
+	}
+
+	tenant := tenants[0]
+	if tenant == SystemTenant || tenant == SharedTenant {
+		// System and shared tenants don't have user records
+		return nil
+	}
+
+	// Get username from subject
+	username := subject.User
+
+	// Get JWT token from context for additional claims
+	token := TokenFromContext(ctx)
+	if token == nil {
+		// No token in context - skip provisioning
+		return nil
+	}
+
+	// Extract JWT claims for email, name, etc.
+	claims, ok := token.Claims.(jwt.MapClaims)
+	if !ok {
+		i.logger.ErrorContext(ctx, "Failed to extract claims from JWT token for user provisioning")
+		return nil
+	}
+
+	// Provision the user
+	err := i.provisioner.Provision(ctx, username, tenant, claims)
+	if err != nil {
+		i.logger.ErrorContext(ctx, "Failed to provision user",
+			slog.String("!username", username),
+			slog.String("!tenant", tenant),
+			slog.Any("error", err),
+		)
+		return fmt.Errorf("user provisioning failed: %w", err)
+	}
+
+	return nil
+}
--- a/internal/auth/grpc_jit_provisioning_interceptor.go
+++ b/internal/auth/grpc_jit_provisioning_interceptor.go
@@ -21,7 +21,7 @@ import (
 	"github.com/golang-jwt/jwt/v5"
 	"google.golang.org/grpc"
 
-	"github.com/osac-project/fulfillment-service/internal/reflection"
+	"github.com/osac-project/fulfillment-service/internal/util"
 )
 
 //go:generate mockgen -source=$GOFILE -destination=grpc_jit_provisioning_interceptor_mock.go -package=$GOPACKAGE UserProvisioner
@@ -57,7 +57,7 @@ func (b *GrpcJitProvisioningInterceptorBuilder) SetLogger(value *slog.Logger) *G
 
 // SetProvisioner sets the user provisioner.
 func (b *GrpcJitProvisioningInterceptorBuilder) SetProvisioner(value UserProvisioner) *GrpcJitProvisioningInterceptorBuilder {
-	b.provisioner = reflection.NormalizeNil(value)
+	b.provisioner = util.NormalizeNil(value)
 	return b
 }
 
--- a/internal/auth/grpc_jit_provisioning_interceptor_test.go
+++ b/internal/auth/grpc_jit_provisioning_interceptor_test.go
@@ -1,0 +1,558 @@
+/*
+Copyright (c) 2026 Red Hat Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with the
+License. You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific
+language governing permissions and limitations under the License.
+*/
+
+package auth
+
+import (
+	"context"
+	"errors"
+	"maps"
+	"time"
+
+	"github.com/golang-jwt/jwt/v5"
+	. "github.com/onsi/ginkgo/v2/dsl/core"
+	. "github.com/onsi/gomega"
+	"google.golang.org/grpc"
+
+	"github.com/osac-project/fulfillment-service/internal/collections"
+	"github.com/osac-project/fulfillment-service/internal/testing"
+	"github.com/osac-project/fulfillment-service/internal/uuid"
+)
+
+// mockUserProvisioner is a simple test double for UserProvisioner.
+type mockUserProvisioner struct {
+	provisionFunc func(ctx context.Context, username, tenant string, claims jwt.MapClaims) error
+	callCount     int
+	lastUsername  string
+	lastTenant    string
+	lastClaims    jwt.MapClaims
+}
+
+func newMockUserProvisioner() *mockUserProvisioner {
+	return &mockUserProvisioner{
+		provisionFunc: func(ctx context.Context, username, tenant string, claims jwt.MapClaims) error {
+			return nil
+		},
+	}
+}
+
+func (m *mockUserProvisioner) Provision(ctx context.Context, username, tenant string, claims jwt.MapClaims) error {
+	m.callCount++
+	m.lastUsername = username
+	m.lastTenant = tenant
+	m.lastClaims = claims
+	return m.provisionFunc(ctx, username, tenant, claims)
+}
+
+var _ UserProvisioner = (*mockUserProvisioner)(nil)
+
+// mockServerStream is a simple test double for grpc.ServerStream.
+type mockServerStream struct {
+	grpc.ServerStream
+	ctx context.Context
+}
+
+func (m *mockServerStream) Context() context.Context {
+	return m.ctx
+}
+
+var _ = Describe("JIT Provisioning Interceptor", func() {
+	var (
+		ctx             context.Context
+		interceptor     *GrpcJitProvisioningInterceptor
+		mockProvisioner *mockUserProvisioner
+		handler         grpc.UnaryHandler
+		handlerCalled   bool
+	)
+
+	// createKeycloakUserToken creates a token resembling the ones issued by Keycloak for regular users.
+	createKeycloakUserToken := func(tenant, name string, overrides jwt.MapClaims) *jwt.Token {
+		now := time.Now()
+		claims := jwt.MapClaims{
+			"auth_time": now.Unix(),
+			"azp":       "osac-cli",
+			"exp":       now.Add(time.Hour).Unix(),
+			"organization": []any{
+				tenant,
+			},
+			"iat":      now.Unix(),
+			"iss":      "https://keycloak.keycloak.svc.cluster.local:8000/realms/osac",
+			"jti":      uuid.New(),
+			"scope":    "openid organization",
+			"sid":      uuid.New(),
+			"sub":      uuid.New(),
+			"typ":      "Bearer",
+			"username": name,
+		}
+		maps.Copy(claims, overrides)
+		return testing.MakeTokenObject(nil, claims)
+	}
+
+	BeforeEach(func() {
+		var err error
+
+		// Create a context:
+		ctx = context.Background()
+
+		// Reset handler tracking:
+		handlerCalled = false
+
+		// Create a mock user provisioner:
+		mockProvisioner = newMockUserProvisioner()
+
+		// Create a mock handler:
+		handler = func(ctx context.Context, req any) (any, error) {
+			handlerCalled = true
+			return nil, nil
+		}
+
+		// Create the interceptor with JIT provisioning enabled:
+		interceptor, err = NewGrpcJitProvisioningInterceptor().
+			SetLogger(logger).
+			SetProvisioner(mockProvisioner).
+			Build()
+		Expect(err).ToNot(HaveOccurred())
+	})
+
+	Describe("Builder validation", func() {
+		It("Requires a logger", func() {
+			_, err := NewGrpcJitProvisioningInterceptor().
+				SetProvisioner(mockProvisioner).
+				Build()
+			Expect(err).To(HaveOccurred())
+			Expect(err.Error()).To(ContainSubstring("logger is mandatory"))
+		})
+
+		It("Allows nil provisioner", func() {
+			interceptor, err := NewGrpcJitProvisioningInterceptor().
+				SetLogger(logger).
+				SetProvisioner(nil).
+				Build()
+			Expect(err).ToNot(HaveOccurred())
+			Expect(interceptor).ToNot(BeNil())
+		})
+	})
+
+	Describe("User provisioning", func() {
+		It("Provisions a new user on first authentication", func() {
+			token := createKeycloakUserToken("tenant1", "alice", jwt.MapClaims{
+				"email":          "alice@example.com",
+				"email_verified": true,
+				"name":           "Alice Smith",
+			})
+			ctx = ContextWithToken(ctx, token)
+
+			// Create subject with single tenant (provisioning should happen)
+			subject := &Subject{
+				User:    "alice",
+				Tenants: collections.NewSet("tenant1"),
+			}
+			ctx = ContextWithSubject(ctx, subject)
+
+			_, err := interceptor.UnaryServer(
+				ctx,
+				nil,
+				&grpc.UnaryServerInfo{
+					FullMethod: "/osac.public.v1.Clusters/List",
+				},
+				handler,
+			)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(handlerCalled).To(BeTrue())
+
+			// Verify provisioner was called:
+			Expect(mockProvisioner.callCount).To(Equal(1))
+			Expect(mockProvisioner.lastUsername).To(Equal("alice"))
+			Expect(mockProvisioner.lastTenant).To(Equal("tenant1"))
+			Expect(mockProvisioner.lastClaims["email"]).To(Equal("alice@example.com"))
+			Expect(mockProvisioner.lastClaims["name"]).To(Equal("Alice Smith"))
+		})
+
+		It("Does not provision when no provisioner is configured", func() {
+			var err error
+			interceptor, err = NewGrpcJitProvisioningInterceptor().
+				SetLogger(logger).
+				Build()
+			Expect(err).ToNot(HaveOccurred())
+
+			token := createKeycloakUserToken("tenant1", "bob", nil)
+			ctx = ContextWithToken(ctx, token)
+			subject := &Subject{
+				User:    "bob",
+				Tenants: collections.NewSet("tenant1"),
+			}
+			ctx = ContextWithSubject(ctx, subject)
+
+			_, err = interceptor.UnaryServer(
+				ctx,
+				nil,
+				&grpc.UnaryServerInfo{
+					FullMethod: "/osac.public.v1.Clusters/List",
+				},
+				handler,
+			)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(handlerCalled).To(BeTrue())
+		})
+
+		It("Does not provision when subject has no tenants", func() {
+			token := createKeycloakUserToken("tenant1", "david", nil)
+			ctx = ContextWithToken(ctx, token)
+			subject := &Subject{
+				User:    "david",
+				Tenants: collections.NewSet[string](), // Empty tenant set
+			}
+			ctx = ContextWithSubject(ctx, subject)
+
+			_, err := interceptor.UnaryServer(
+				ctx,
+				nil,
+				&grpc.UnaryServerInfo{
+					FullMethod: "/osac.public.v1.Clusters/List",
+				},
+				handler,
+			)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(handlerCalled).To(BeTrue())
+			Expect(mockProvisioner.callCount).To(Equal(0))
+		})
+
+		It("Does not provision when subject has multiple tenants", func() {
+			token := createKeycloakUserToken("tenant1", "eve", jwt.MapClaims{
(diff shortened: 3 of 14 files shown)


omer-vishlitzky:

**Root cause:** osac-installer main references unmerged osac-operator commit 08991cc; image was never published to ghcr.io

**Causal chain:**
- osac-installer commit 768174e updated osac-operator submodule to 08991ccbb (unmerged PR #320)
- values/vmaas-ci/values.yaml set operator.image.tag to sha-08991cc
- vmaas-helm snapshot rebuilt after 21:13 UTC with broken reference
- Boot step runs Helm upgrade with broken image tag
- Pod osac-operator-68866ff5c6-6gcnq tries to pull ghcr.io/osac-project/osac-operator:sha-08991cc
- Image pull fails: manifest unknown (image never published - PR builds don't push to ghcr.io)
- Refresh phase times out waiting for osac-operator rollout
- Boot step fails with ImagePullBackOff

**Evidence:**

[`step:e2e-vmaas-osac-project-cluster-tool-boot`](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/osac-project_fulfillment-service/782/pull-ci-osac-project-fulfillment-service-main-e2e-vmaas/2070263999802380288/artifacts/e2e-vmaas/osac-project-cluster-tool-boot/build-log.txt):
```
ERROR: parallel tasks failed: wait fulfillment: osac-operator: osac-operator-68866ff5c6-6gcnq: ImagePullBackOff: Back-off pulling image "ghcr.io/osac-project/osac-operator:sha-08991cc": ErrImagePull: initializing source docker://ghcr.io/osac-project/osac-operator:sha-08991cc: reading manifest sha-08991cc in ghcr.io/osac-project/osac-operator: manifest unknown
```

[`osac-logs/events.txt`](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/osac-project_fulfillment-service/782/pull-ci-osac-project-fulfillment-service-main-e2e-vmaas/2070263999802380288/artifacts/e2e-vmaas/osac-project-gather/artifacts/osac-logs/osac-logs/events.txt):
```
10s         Warning   Failed                 pod/osac-operator-68866ff5c6-6gcnq                     Failed to pull image "ghcr.io/osac-project/osac-operator:sha-08991cc": initializing source docker://ghcr.io/osac-project/osac-operator:sha-08991cc: reading manifest sha-08991cc in ghcr.io/osac-project/osac-operator: manifest unknown
```

[`osac-installer:values/vmaas-ci/values.yaml:8`](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/osac-project_fulfillment-service/782/pull-ci-osac-project-fulfillment-service-main-e2e-vmaas/2070263999802380288/artifacts/e2e-vmaas/osac-project-gather/artifacts/osac-logs/osac-installer:values/vmaas-ci/values.yaml:8):
```
tag: sha-08991cc
```

`osac-installer:commit:768174ea9e0ec49816717c859006abdb9727bfcf`:
```
Sync image tags with submodule commits (committed 2026-06-25 21:13:27 UTC)
```

`gh:osac-operator:pr:320`:
```
PR #320 state=OPEN, mergedAt=null - commit 08991ccbb443e42f3a464c3e4b9339e396fbe221 never merged to main
```

`gh:osac-operator:actions`:
```
Build container image for 08991cc ran on event=pull_request (doesn't push to ghcr.io)
```

[`step:e2e-vmaas-osac-project-cluster-tool-boot`](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/osac-project_fulfillment-service/782/pull-ci-osac-project-fulfillment-service-main-e2e-vmaas/2070263999802380288/artifacts/e2e-vmaas/osac-project-cluster-tool-boot/build-log.txt):
```
CLUSTER_TOOL_FLAVOR_IMAGE: quay.io/rh-ee-ovishlit/cluster-flavors:vmaas-helm
```

`differential:pr:779:success:build:2070247383421161472`:
```
Passing run (22:30:29) used ghcr.io/osac-project/osac-operator:sha-8220502 (exists and works)
```

`differential:pr:723:failure:build:2070264122477383680`:
```
Different PR (#723) failed at 22:47:21 with identical error: osac-operator sha-08991cc manifest unknown
```

**Suggestion:** Revert osac-installer commit 768174e OR merge osac-operator PR #320 and manually trigger container image build workflow to publish sha-08991cc image to ghcr.io, then rebuild vmaas-helm snapshot

---
[Prow job](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/osac-project_fulfillment-service/782/pull-ci-osac-project-fulfillment-service-main-e2e-vmaas/2070263999802380288) | Build `2070263999802380288` | 🤖 *triagent* | Cost: $1.3740

internal/auth/grpc_authz_interceptor.go
```diff
@@ -37,6 +37,15 @@ import (
 	k8sfiles "github.com/osac-project/fulfillment-service/internal/kubernetes/files"
 )
 
+// UserProvisioner is an interface for just-in-time user provisioning. Implementations can create users from JWT claims.
+//
+//go:generate mockgen -destination=user_provisioner_mock.go -package=auth . UserProvisioner
+type UserProvisioner interface {
+	// EnsureUserExists checks if a user exists and creates them if not. This is best-effort and should not return
+	// errors - failures are logged but don't block authentication.
```
jhernand: Why the best-effort approach? Can't we return an error when the provisioning fails?

internal/auth/grpc_authz_interceptor.go
```diff
@@ -37,6 +37,15 @@ import (
...
+	EnsureUserExists(ctx context.Context, username, tenant string, claims jwt.MapClaims)
```
jhernand: If this is a provisioner I'd say the method should be named `Provision`.

internal/auth/grpc_authz_interceptor.go
```diff
@@ -115,6 +126,14 @@ func (b *GrpcAuthzInterceptorBuilder) AddEmergencyServiceAccounts(
 	return b
 }
 
+// SetUserProvisioner sets the UserProvisioner used for just-in-time user provisioning. When a JWT-authenticated user
+// makes their first request, the interceptor will automatically create a user record in the database if one doesn't
+// exist. This is optional - if not set, JIT provisioning is disabled.
+func (b *GrpcAuthzInterceptorBuilder) SetUserProvisioner(value UserProvisioner) *GrpcAuthzInterceptorBuilder {
+	b.userProvisioner = value
```
jhernand: In setters that pass an interface it is good practice to use `reflection.NormalizeNil` to avoid interfaces that aren't `nil` but have values that are `nil`.

internal/auth/grpc_authz_interceptor.go
```diff
@@ -568,6 +593,43 @@ func (i *GrpcAuthzInterceptor) buildSubject(authzData map[string]any) (result *S
 	return
 }
 
+// ensureUserExists performs just-in-time user provisioning for JWT-authenticated users by delegating to the
+// configured UserProvisioner. This function never returns an error - failures are logged but don't block the request.
```
jhernand: I think it is better to fail if the user can't be created. Why do you think it is better to just continue?

internal/auth/grpc_authz_interceptor.go
```diff
@@ -568,6 +593,43 @@ func (i *GrpcAuthzInterceptor) buildSubject(authzData map[string]any) (result *S
...
+func (i *GrpcAuthzInterceptor) ensureUserExists(ctx context.Context, subject *Subject, token *jwt.Token) {
+	// Skip JIT provisioning if no user provisioner is configured
+	if i.userProvisioner == nil {
```
jhernand: This ^ is why you want to use `reflection.NormalizeNil` in the builder. I you don't then this may be `true` and still you will get a `nil` pointer exception when trying to use the value.

internal/auth/grpc_authz_interceptor.go
```diff
@@ -568,6 +593,43 @@ func (i *GrpcAuthzInterceptor) buildSubject(authzData map[string]any) (result *S
...
+	}
+
+	// Determine the tenant deterministically from the claims
+	// The policy uses 'organization' claim to derive the first tenant, so we do the same here
+	var tenant string
+	if orgSlice := claimAsAnySlice(claims, "organization"); len(orgSlice) > 0 {
+		if s, ok := orgSlice[0].(string); ok {
+			tenant = s
+		}
+	}
```
jhernand: I think it would be better to use the `Tenants` field from the subject. That is the set of tenants that the user can see. If it is exactly one tenant then that is the tenant that you have to use. If empty, more than one, or infinite, then you know this isn't a regular tenant user and you cannot a user record.

internal/database/dao/user_provisioner.go
jhernand, line 1: I understand that you are putting the implementation of the provisioner in the `dao` package to avoid circular dependencies. But it doesn't feel right. I'd prefer to keep logic out of the `database` package, if possible. Can we move this implementation to a new `provisioners` package, for example?

internal/auth/grpc_authz_interceptor.go
jhernand, line 1: Looks like the provisioning logic is mostly independent of the rest of the logic of this interceptor: it only happens at the very end, and it only uses the information from the result of this interceptor. Would it make sense to have a separate interceptor for this?

internal/database/dao/user_provisioner_test.go
```diff
@@ -0,0 +1,344 @@
...
+			// Should not panic:
+			Expect(func() {
+				provisioner.EnsureUserExists(ctxNoTx, "iris", "tenant9", claims)
+			}).ToNot(Panic())
+		})
+	})
+})
+
+// createUsersTableSQL creates the `users` table for testing purposes.
+const createUsersTableSQL = `
```
jhernand: Try to not do this. If the users table is going to be needed, then it should be created by a migration.

jhernand (changes requested):

internal/auth/grpc_authz_interceptor.go
CrystalChun: That makes sense! I created a separate interceptor as suggested. How does it look? Thank you for your feedback.

internal/auth/grpc_authz_interceptor.go
CrystalChun: I guess to allow the workflow to continue, but you're right it's better to ensure that the user does exist in our system. Updated to return an error. Thank you!

internal/auth/grpc_authz_interceptor.go
CrystalChun: Gotcha, thanks Juan! Updated to Provision method

internal/database/dao/user_provisioner.go
CrystalChun: That makes sense, I've moved it out. Thank you!

internal/auth/grpc_authz_interceptor.go
CrystalChun: Makes sense, returning an error now. Thanks!

internal/auth/grpc_authz_interceptor.go
CrystalChun: Makes sense, thank you for pointing that out! I've modified it to only store users with a finite set with exactly one tenant.

internal/auth/grpc_authz_interceptor.go
CrystalChun: Gotcha, thank you for letting me know! Updated to use `reflection.NormalizeNil` in the setter.

CrystalChun:
@coderabbit review

internal/auth/grpc_jit_provisioning_interceptor.go
```diff
@@ -0,0 +1,140 @@
...
+	}
+	result = &GrpcJitProvisioningInterceptor{
+		logger:      b.logger,
+		provisioner: b.provisioner,
+	}
+	return result, nil
+}
+
+// UnaryServer is the unary server interceptor function.
+func (i *GrpcJitProvisioningInterceptor) UnaryServer(ctx context.Context, req interface{},
```
jhernand: What about `StreamServer`? Don't we want to automatically create users also when the first thing they do is call the `Watch` method of the events service?

jhernand: If we are using `req interface{}` somewhere we should change it to `any`.

internal/auth/grpc_jit_provisioning_interceptor_test.go
```diff
@@ -0,0 +1,450 @@
...
+	"google.golang.org/grpc"
+	grpccodes "google.golang.org/grpc/codes"
+	grpcstatus "google.golang.org/grpc/status"
+
+	"github.com/osac-project/fulfillment-service/internal/collections"
+	"github.com/osac-project/fulfillment-service/internal/testing"
+	"github.com/osac-project/fulfillment-service/internal/uuid"
+)
+
+// mockUserProvisioner is a simple test double for UserProvisioner.
```
jhernand, line 32: Isn't it better to automatically generate a mock with the `go:generate ...`?

internal/provisioners/user_provisioner.go
```diff
@@ -0,0 +1,124 @@
...
+	"github.com/golang-jwt/jwt/v5"
+
+	privatev1 "github.com/osac-project/fulfillment-service/internal/api/osac/private/v1"
+	"github.com/osac-project/fulfillment-service/internal/database/dao"
+)
+
+// UserProvisioner provisions user records in the database.
+type UserProvisioner interface {
+	Provision(ctx context.Context, username, tenant string, claims jwt.MapClaims) error
+}
```
jhernand: I think you don't need to repeat the interface definition here.

internal/provisioners/user_provisioner.go
```diff
@@ -0,0 +1,124 @@
...
+
+// DAOUserProvisionerBuilder builds a DAOUserProvisioner.
+type DAOUserProvisionerBuilder struct {
```
jhernand: I'd call this just `UserProvisionerBuilder` and `UserProvisioiner`, without the DAO part. If you leave the DAO part it reads as it what we are provisioning is a user of DAO.

internal/provisioners/user_provisioner_test.go
```diff
@@ -0,0 +1,383 @@
...
+			Expect(listResp.GetSize()).To(Equal(int32(1)))
+
+			user := listResp.GetItems()[0]
+			Expect(user.GetMetadata().GetName()).To(Equal("alice"))
+			Expect(user.GetMetadata().GetTenant()).To(Equal(auth.SystemTenant))
+			Expect(user.GetSpec().GetUsername()).To(Equal("alice"))
+			Expect(user.GetSpec().GetEmail()).To(Equal("alice@example.com"))
+			Expect(user.GetSpec().GetEmailVerified()).To(BeTrue())
+			Expect(user.GetSpec().GetFirstName()).To(Equal("Alice"))
+			Expect(user.GetSpec().GetLastName()).To(Equal("Smith"))
```
jhernand: I know it isn't part of this patch, but seeing the name of the user here (I mean "Alice Smith") makes me think if we really need to store that in our database. Privacy laws consider that protected data. Se we better not store it unless it is absolutely necessary. E-mails are personal data as well. Maybe we should reword things a bit here and call that "domain" instead of "e-mail".

internal/auth/grpc_jit_provisioning_interceptor_test.go
CrystalChun: Whoops, yes! Thank you!

internal/auth/grpc_jit_provisioning_interceptor.go
CrystalChun: That makes sense, updated it. Thank you Juan!

omer-vishlitzky:
🔴 **CI Triage: `pr_issue`** | Category: `TEST`

**Root cause:** PR #782 adds a new GrpcJitProvisioningInterceptor that runs on every authenticated single-tenant user request and calls UserProvisioner.Provision(), which uses a GenericDAO[*privatev1.User] to query the database; this fails in the E2E environment (likely missing User table/schema), returning a hard gRPC Internal error "failed to provision user" that blocks all user-scoped API calls.

**Causal chain:**
- PR #782 adds a new GrpcJitProvisioningInterceptor inserted into the gRPC unary and stream interceptor chains in start_grpc_server_cmd.go
- The interceptor runs after authz for every request from a subject with exactly one non-system/non-shared tenant (i.e., regular users)
- It calls UserProvisioner.Provision() which uses dao.GenericDAO[*privatev1.User] to list users matching the authenticated username
- The DAO query fails because the User table schema or tables/osac.public.v1.User.yaml is not available in the E2E test environment
- On provisioning failure, the interceptor returns grpcstatus.Error(grpccodes.Internal, "failed to provision user") instead of passing through to the handler
- All 12 test_jwt_user_can_list parametrized tests fail because the osac CLI gets 'rpc error: code = Internal desc = failed to provision user' for every resource type
- test_jwt_admin_can_list_users fails with 'open tables/osac.public.v1.User.yaml: file does not exist' confirming the User resource definition is missing
- Admin list tests (test_jwt_admin_can_list) pass because admin subjects have non-finite tenants, bypassing the provisioning interceptor entirely

**Evidence:**

[`build-log.txt`](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/osac-project_fulfillment-service/782/pull-ci-osac-project-fulfillment-service-main-e2e-vmaas/2070560627960582144/artifacts/e2e-vmaas/osac-project-gather/artifacts/osac-logs/build-log.txt):
```
stderr = 'Error: rpc error: code = Internal desc = failed to provision user\n'
```

[`build-log.txt`](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/osac-project_fulfillment-service/782/pull-ci-osac-project-fulfillment-service-main-e2e-vmaas/2070560627960582144/artifacts/e2e-vmaas/osac-project-gather/artifacts/osac-logs/build-log.txt):
```
stderr = 'Error: open tables/osac.public.v1.User.yaml: file does not exist\n'
```

[`PR #782 diff: internal/auth/grpc_jit_provisioning_interceptor.go:163`](https://github.com/osac-project/fulfillment-service/pull/782):
```
return nil, grpcstatus.Error(grpccodes.Internal, "failed to provision user")
```

[`PR #782 diff: internal/cmd/service/start/grpcserver/start_grpc_server_cmd.go`](https://github.com/osac-project/fulfillment-service/pull/782):
```
jitProvisioningInterceptor.UnaryServer added to grpc.ChainUnaryInterceptor chain
```

[`PR #782 diff: internal/provisioners/user_provisioner.go`](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/osac-project_fulfillment-service/782/pull-ci-osac-project-fulfillment-service-main-e2e-vmaas/2070560627960582144/artifacts/e2e-vmaas/osac-project-gather/artifacts/osac-logs/PR #782 diff: internal/provisioners/user_provisioner.go):
```
dao.NewGenericDAO[*privatev1.User]() used to query users but User table definition missing from deployment
```

`ci-obs recent runs`:
```
Other PRs (#723, #613, #576, #787, #724, #780, #777, #785) all pass e2e-vmaas; failure is unique to PR #782
```

**Suggestion:** The PR needs to either: (1) ensure the User table definition (tables/osac.public.v1.User.yaml) is included in the fulfillment-service deployment and the database schema is migrated before the JIT provisioner runs, or (2) make the provisioning interceptor gracefully degrade (log and continue rather than returning a hard error) when provisioning fails, so that existing API functionality is not blocked by a user-provisioning failure.

---
[Prow job](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/osac-project_fulfillment-service/782/pull-ci-osac-project-fulfillment-service-main-e2e-vmaas/2070560627960582144) | Build `2070560627960582144` | 🤖 *triagent* | Cost: $2.0273

omer-vishlitzky:
🔴 **CI Triage: `pr_issue`** | Category: `TEST`

**Root cause:** PR #782's JIT user provisioning interceptor runs within the shared PostgreSQL transaction and poisons it when the user provisioning DB operation fails, causing all subsequent operations (including the actual list) to fail for regular users while admins bypass the interceptor entirely.

**Causal chain:**
- PR #782 adds a GrpcJitProvisioningInterceptor to the gRPC interceptor chain, positioned after authz and inside the TxInterceptor's shared PostgreSQL transaction.
- For regular JWT users with a single tenant, the interceptor calls usersDAO.List() and potentially usersDAO.Create() within the request's transaction to auto-provision user records.
- The provisioning DB operation fails (likely a schema or query issue), causing PostgreSQL to mark the transaction as aborted.
- The interceptor catches the Go error and continues to the handler (line: 'Don't return error - user provisioning is non-critical'), but the PostgreSQL transaction state is already 'aborted'.
- The actual list handler runs within the same aborted transaction, and PostgreSQL rejects all further queries, resulting in 'rpc error: code = Internal desc = failed to list'.
- Admin users have non-finite tenants and hit the early return at '!subject.Tenants.Finite()', bypassing JIT provisioning entirely — their transactions stay clean and all admin list tests pass.
- The test_jwt_admin_can_list_users failure is a separate CLI-side issue: the osac CLI binary lacks osac.public.v1.User.yaml rendering table definition.

**Evidence:**

[`step:e2e-vmaas-osac-project-cluster-tool-test`](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/osac-project_fulfillment-service/782/pull-ci-osac-project-fulfillment-service-main-e2e-vmaas/2070606252777410560/artifacts/e2e-vmaas/osac-project-cluster-tool-test/build-log.txt):
```
FAILED tests/vmaas/test_jwt_auth_smoke.py::test_jwt_user_can_list[clustertemplates] — stderr = 'Error: rpc error: code = Internal desc = failed to list\n'
```

[`step:e2e-vmaas-osac-project-cluster-tool-test`](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/osac-project_fulfillment-service/782/pull-ci-osac-project-fulfillment-service-main-e2e-vmaas/2070606252777410560/artifacts/e2e-vmaas/osac-project-cluster-tool-test/build-log.txt):
```
All 12 test_jwt_user_can_list parametrized tests FAILED with 'failed to list', while all 12 test_jwt_admin_can_list tests PASSED
```

[`step:e2e-vmaas-osac-project-cluster-tool-test`](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/osac-project_fulfillment-service/782/pull-ci-osac-project-fulfillment-service-main-e2e-vmaas/2070606252777410560/artifacts/e2e-vmaas/osac-project-cluster-tool-test/build-log.txt):
```
test_jwt_admin_can_list_users FAILED — stderr = 'Error: open tables/osac.public.v1.User.yaml: file does not exist\n'
```

[`PR #666 build log (differential)`](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/osac-project_fulfillment-service/666/pull-ci-osac-project-fulfillment-service-main-e2e-vmaas/2070556338554933248):
```
PR #666: ALL test_jwt_user_can_list tests PASSED, test_jwt_admin_can_list_users PASSED — 51 passed, 3 unrelated failures
```

[`PR #782 diff`](https://github.com/osac-project/fulfillment-service/pull/782):
```
New file internal/auth/grpc_jit_provisioning_interceptor.go — interceptor calls usersDAO.List().Do(ctx) and usersDAO.Create().Do(ctx) within the shared transaction
```

`PR #782 diff`:
```
start_grpc_server_cmd.go adds jitProvisioningInterceptor.UnaryServer to grpc.ChainUnaryInterceptor after authzInterceptor and inside txInterceptor
```

[`internal/database/database_tx_interceptor.go`](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/osac-project_fulfillment-service/782/pull-ci-osac-project-fulfillment-service-main-e2e-vmaas/2070606252777410560/artifacts/e2e-vmaas/osac-project-gather/artifacts/osac-logs/internal/database/database_tx_interceptor.go):
```
TxInterceptor.UnaryServer begins a transaction, adds it to context, calls handler chain, then ends transaction — all interceptors share the same PostgreSQL transaction
```

[`internal/database/dao/generic_dao_list.go`](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/osac-project_fulfillment-service/782/pull-ci-osac-project-fulfillment-service-main-e2e-vmaas/2070606252777410560/artifacts/e2e-vmaas/osac-project-gather/artifacts/osac-logs/internal/database/dao/generic_dao_list.go):
```
ListRequest.Do() calls database.TxFromContext(ctx) to get the shared transaction, then r.tx.ReportError(&err) on failure
```

**Suggestion:** Fix the JIT provisioning interceptor to not run within the shared request transaction. Either: (1) use a separate database connection/transaction for the provisioning operation so a failure doesn't poison the main request transaction, or (2) use a savepoint (SAVEPOINT/ROLLBACK TO SAVEPOINT) to isolate the provisioning query. Also add the missing osac.public.v1.User.yaml rendering table definition for the CLI.

---
[Prow job](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/osac-project_fulfillment-service/782/pull-ci-osac-project-fulfillment-service-main-e2e-vmaas/2070606252777410560) | Build `2070606252777410560` | 🤖 *triagent* | Cost: $4.9383

omer-vishlitzky:
🔴 **CI Triage: `pr_issue`** | Category: `TEST`

**Root cause:** PR #782's JIT user provisioning interceptor auto-creates User records on authentication, causing `osac get users` to return data for the first time. The table renderer in the test image's CLI binary (v0.0.62) lacks the `osac.public.v1.User.yaml` table definition needed to render users, so it fatally errors instead of falling back to a default table.

**Causal chain:**
- PR #782 adds a JIT provisioning interceptor that auto-creates User records when an authenticated user makes a gRPC request
- When `test_jwt_admin_can_list_users` runs `osac get users` as admin, the admin's request triggers JIT provisioning, creating a User record in the database
- The `osac get users` List response now contains at least 1 User object (the admin), whereas before PR #782 it returned 0 objects
- With 0 objects, `renderTable()` at get_cmd.go:236 returns early without calling the table renderer — this is why the test passed on other PRs
- With 1+ objects, the table renderer is invoked and calls `loadTable()` at table_renderer.go:247 to load `osac.public.v1.User.yaml`
- The CLI binary in the test image (v0.0.62 from GitHub releases) does not embed `osac.public.v1.User.yaml` — PR #782 adds it but the test image uses a pre-built release binary
- The `loadTable()` function has a bug: the comment says 'file doesn't exist, that's okay' but the code returns the error instead of falling back to `defaultTable()`
- The error propagates up through `renderTable()` → CLI exits with code 1 and message 'Error: open tables/osac.public.v1.User.yaml: file does not exist'

**Evidence:**

[`step:e2e-vmaas-osac-project-cluster-tool-test`](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/osac-project_fulfillment-service/782/pull-ci-osac-project-fulfillment-service-main-e2e-vmaas/2070641407676125184/artifacts/e2e-vmaas/osac-project-cluster-tool-test/build-log.txt):
```
Error: open tables/osac.public.v1.User.yaml: file does not exist
```

[`step:e2e-vmaas-osac-project-cluster-tool-test`](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/osac-project_fulfillment-service/782/pull-ci-osac-project-fulfillment-service-main-e2e-vmaas/2070641407676125184/artifacts/e2e-vmaas/osac-project-cluster-tool-test/build-log.txt):
```
FAILED tests/vmaas/test_jwt_auth_smoke.py::test_jwt_admin_can_list_users - subprocess.CalledProcessError: Command '('osac', 'get', 'users')' returned non-zero exit status 1.
```

[`internal/rendering/table_renderer.go:250-253 (fulfillment-service main)`](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/osac-project_fulfillment-service/782/pull-ci-osac-project-fulfillment-service-main-e2e-vmaas/2070641407676125184/artifacts/e2e-vmaas/osac-project-gather/artifacts/osac-logs/internal/rendering/table_renderer.go:250-253 (fulfillment-service main)):
```
data, err := fs.ReadFile(tablesFS, path.Join("tables", file)) / if err != nil { // If the file doesn't exist, that's okay - we'll use the default table. / return } — returns (nil, err), NOT (nil, nil)
```

[`internal/cmd/cli/get/get_cmd.go:236-238 (fulfillment-service main)`](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/osac-project_fulfillment-service/782/pull-ci-osac-project-fulfillment-service-main-e2e-vmaas/2070641407676125184/artifacts/e2e-vmaas/osac-project-gather/artifacts/osac-logs/internal/cmd/cli/get/get_cmd.go:236-238 (fulfillment-service main)):
```
if len(objects) == 0 { c.console.Render(ctx, "no_matching_objects.txt", nil); return nil } — skips table rendering when list is empty
```

`osac-test-infra Containerfile`:
```
ARG OSAC_VERSION=0.0.62 — test image uses pre-built CLI v0.0.62 which lacks osac.public.v1.User.yaml
```

`PR #790 passing run (build 2070634730130247680)`:
```
tests/vmaas/test_jwt_auth_smoke.py::test_jwt_admin_can_list_users PASSED — passes because no JIT provisioning means 0 users returned
```

[`PR #782 changed files`](https://github.com/osac-project/fulfillment-service/pull/782):
```
internal/rendering/tables/osac.public.v1.User.yaml (added) — PR adds the table definition but test image CLI doesn't use it
```

**Suggestion:** Fix the `loadTable()` bug in `table_renderer.go:248-253` to return `(nil, nil)` instead of `(nil, err)` when the table file doesn't exist — this matches the comment's intent and lets `defaultTable()` be used as fallback. Alternatively, bump the CLI version in osac-test-infra's Containerfile after merging PR #782 and cutting a new release.

---
[Prow job](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/osac-project_fulfillment-service/782/pull-ci-osac-project-fulfillment-service-main-e2e-vmaas/2070641407676125184) | Build `2070641407676125184` | 🤖 *triagent* | Cost: $5.9332

internal/auth/grpc_jit_provisioning_interceptor.go
```diff
@@ -0,0 +1,253 @@
...
+	// Provision the user using a savepoint to isolate from the main transaction
+	// This ensures that provisioning errors don't poison the main request transaction
+	err := i.provisionWithSavepoint(ctx, username, tenant, claims)
+	if err != nil {
+		i.logger.ErrorContext(ctx, "Failed to provision user (continuing with request)",
+			slog.String("!username", username),
+			slog.String("!tenant", tenant),
+			slog.Any("error", err),
+		)
+		// Don't return error - user provisioning is non-critical, allow the request to proceed
```
jhernand: I still think this shouldn't proceed.

internal/auth/grpc_jit_provisioning_interceptor.go
```diff
@@ -0,0 +1,253 @@
...
+	}
```
jhernand: All this logic seems to be repeated in the unary and stream interceptors. Can it be extracted to a separate method called by both?

internal/auth/grpc_jit_provisioning_interceptor.go
```diff
@@ -0,0 +1,253 @@
...
+
+	// Continue to handler
+	return handler(srv, stream)
+}
+
+// provisionWithSavepoint provisions a user within a savepoint to isolate errors from the main transaction.
+func (i *GrpcJitProvisioningInterceptor) provisionWithSavepoint(ctx context.Context, username, tenant string, claims jwt.MapClaims) error {
```
jhernand: The savepoint is cool, but is it really needed? Can't we just fail the complete operation if creating the user fails? I'd say that is reasonable.

internal/provisioners/user_provisioner.go
```diff
@@ -0,0 +1,121 @@
...
+			Name:   username,
+			Tenant: tenant,
+		}.Build(),
+		Spec: privatev1.UserSpec_builder{
+			Username:      username,
+			Email:         email,
+			EmailVerified: emailVerified,
+			Enabled:       true,
+			FirstName:     firstName,
+			LastName:      lastName,
```
jhernand: I think we don't need to firs and last names, and we don't need the email verified flag. Can we remove that from the API user type? Not in this patch, but in this patch we should at least not start to save them.

internal/provisioners/user_provisioner.go
```diff
@@ -0,0 +1,104 @@
...
+		usersDAO: b.usersDAO,
+	}
+	return result, nil
+}
+
+// Provision creates a user record if it doesn't exist.
+func (p *UserProvisioner) Provision(ctx context.Context, username, tenant string, claims jwt.MapClaims) error {
+	// Check if user exists (escape username to prevent filter injection)
+	escapedUsername := strings.ReplaceAll(strings.ReplaceAll(username, "\\", "\\\\"), "'", "\\'")
+	filter := fmt.Sprintf("this.spec.username=='%s'", escapedUsername)
```
jhernand: Use `%q` instead of `%s`: it is safer and simpler that the custom escaping.

jhernand (approved):

internal/util/nil.go
CrystalChun, line 1: CC @jhernand there was an import cycle somehow 
```
auth → reflection → config → auth
```
and the suggestion was to move `NormalizeNil` into a separate utility package. Is this ok or should I explore a different solution?

jhernand (approved):

internal/util/nil.go
jhernand: I see. I tried to break the dependency between `reflection` and `config` in https://github.com/osac-project/fulfillment-service/pull/820. But merge this and I will refactor it later.

PR #783, opened 2026-06-26
OSAC-1416: enforce field_definitions as complete contract on catalog item usage

alosadagrande:
## Summary
- Reject spec fields not listed in `field_definitions` when creating a resource via catalog item (`InvalidArgument` with field names)
- Reject user-provided values for non-editable fields instead of silently overwriting with the default (`InvalidArgument`)
- Update `applyFieldDefinitions` docstring to reflect the new behavior
- Add unit tests
- Update CI tests to match the new behaviour

## Jira
https://issues.redhat.com/browse/OSAC-1416

## Test plan
- [x] Unit tests: 18 `applyFieldDefinitions` tests covering unlisted fields, non-editable rejection, happy path, system fields, nested paths, ComputeInstanceSpec
- [x] Manual validation on hub0 cluster: unlisted field rejected, non-editable override rejected, valid creation succeeds
- [x] `gofmt`, `buf generate`, `go build` pass
- [x] `ginkgo run -r internal` passes (1 pre-existing failure on migration hash unrelated to this change)

## Summary by CodeRabbit

* **Bug Fixes**
  * Catalog item field handling is now stricter: resource creation rejects any spec fields not covered by `field_definitions`, including nested paths.
  * User-provided values for non-editable catalog-item fields are now rejected with `InvalidArgument`; catalog defaults are applied only when the value is omitted.
  * Disallowed fields are reported more clearly with deterministic error ordering.
* **Documentation**
  * Updated `CATALOG_ITEMS.md` to reflect the new rejection and non-editable field behaviors.
* **Tests**
  * Updated and expanded server tests to cover unlisted-field rejection and non-editable-field error handling.


alosadagrande:
> Caution
> 
> Inline review comments failed to post. This is likely due to GitHub's internal server error or limits when posting large numbers of comments. If you are seeing this consistently it is likely a permissions issue. Please check "Moderation" -> "Code review limits" under your organization settings.
> 
> **Actionable comments posted: 2**
> 
> 🤖 Prompt for all review comments with AI agents
> ```
> Verify each finding against current code. Fix only still-valid issues, skip the
> rest with a brief reason, keep changes minimal, and validate.
> 
> Inline comments:
> In `@docs/CATALOG_ITEMS.md`:
> - Around line 138-146: The compute-instance example in the catalog docs uses
> unsupported field paths, so update the example to match the valid
> compute-instance schema used elsewhere in this PR. Replace the
> `field_definitions` entries that reference `ssh_public_key` and `host_type` with
> supported paths such as `ssh_key`, `cores`, `memory_gib`, and
> `network_attachments`, keeping the example consistent with the catalog item
> model.
> - Around line 46-49: Update the catalog-item documentation to remove the
> “silently discarded” wording for unsupported field_definition.path values and
> replace it with the current server behavior. In CATALOG_ITEMS.md, revise the
> paragraph under the custom parameters section so it no longer suggests unknown
> paths are ignored; make the description accurately reflect that invalid paths
> are not accepted and can cause catalog-item definitions to fail.
> ```
> 
> 🪄 Autofix (Beta)
> Fix all unresolved CodeRabbit comments on this PR:
> 
> * [ ]   Push a commit to this branch (recommended)
> * [ ]   Create a new PR with the fixes
> 
> ℹ️ Review info
> ⚙️ Run configuration
> **Configuration used**: Repository: osac-project/coderabbit/.coderabbit.yaml
> 
> **Review profile**: ASSERTIVE
> 
> **Plan**: Enterprise
> 
> **Run ID**: `83733a9f-3559-4bff-a80b-cf39d95dec48`
> 
> 📥 Commits
> Reviewing files that changed from the base of the PR and between [b85b011](https://github.com/osac-project/fulfillment-service/commit/b85b011fa53f975f2b42972746740a4ff84e5b2c) and [edb120c](https://github.com/osac-project/fulfillment-service/commit/edb120cf1eb5846d0c967612e4f23be41f5950f0).
> 
> 📒 Files selected for processing (5)
> * `docs/CATALOG_ITEMS.md`
> * `internal/servers/catalog_item_validation.go`
> * `internal/servers/catalog_item_validation_test.go`
> * `internal/servers/private_clusters_server_test.go`
> * `internal/servers/private_compute_instances_server_test.go`
> 
> 🛑 Comments failed to post (2)
> > docs/CATALOG_ITEMS.md (2)> 46-49: _📐 Maintainability & Code Quality_ | _🟡 Minor_ | _⚡ Quick win_
> > > **Update the unknown-path behavior description.**
> > > These lines still say unsupported `field_definition.path` values are “silently discarded”, but the current server flow does not behave that way anymore. Leaving this as-is will send admins toward catalog-item definitions that fail instead of being ignored.
> > > 
> > > 🤖 Prompt for AI Agents
> > > ```
> > > Verify each finding against current code. Fix only still-valid issues, skip the
> > > rest with a brief reason, keep changes minimal, and validate.
> > > 
> > > In `@docs/CATALOG_ITEMS.md` around lines 46 - 49, Update the catalog-item
> > > documentation to remove the “silently discarded” wording for unsupported
> > > field_definition.path values and replace it with the current server behavior. In
> > > CATALOG_ITEMS.md, revise the paragraph under the custom parameters section so it
> > > no longer suggests unknown paths are ignored; make the description accurately
> > > reflect that invalid paths are not accepted and can cause catalog-item
> > > definitions to fail.
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > 
> > > 
> > > 
> > > 
> > > 
> > > 
> > > 138-146: _📐 Maintainability & Code Quality_ | _🟡 Minor_ | _⚡ Quick win_
> > > **Fix the compute-instance example paths.**
> > > This example uses `ssh_public_key` and `host_type`, but the supported compute-instance paths in this PR are `ssh_key`, `cores`, `memory_gib`, `network_attachments`, etc. Copying this YAML will produce an invalid catalog item example.
> > > 
> > > 🤖 Prompt for AI Agents
> > > ```
> > > Verify each finding against current code. Fix only still-valid issues, skip the
> > > rest with a brief reason, keep changes minimal, and validate.
> > > 
> > > In `@docs/CATALOG_ITEMS.md` around lines 138 - 146, The compute-instance example
> > > in the catalog docs uses unsupported field paths, so update the example to match
> > > the valid compute-instance schema used elsewhere in this PR. Replace the
> > > `field_definitions` entries that reference `ssh_public_key` and `host_type` with
> > > supported paths such as `ssh_key`, `cores`, `memory_gib`, and
> > > `network_attachments`, keeping the example consistent with the catalog item
> > > model.
> > > ```

solved

docs/CATALOG_ITEMS.md
```diff
@@ -136,14 +135,21 @@ description: General-purpose virtual machine with KubeVirt.
...
 field_definitions:
-  - path: ssh_public_key
-    display_name: SSH Public Key
+  - path: ssh_key
+    display_name: SSH Key
     editable: true
     default: "ssh-ed25519 AAAA..."
-  - path: host_type
-    display_name: Host Type
+  - path: cores
```
ygalblum: Following the merge of https://github.com/osac-project/fulfillment-service/pull/735, `cores` and `memory_gib` are now deprecated. Use `instance_type` instead

docs/CATALOG_ITEMS.md
ygalblum: As mentioned before, `cores` and `memory_gib` should no longer be used

docs/CATALOG_ITEMS.md
```diff
@@ -44,9 +44,8 @@ Catalog items control a **fixed set of known fields** on the resource (e.g., `pu
 flags (`--pull-secret`, `--ssh-public-key`, `--pod-cidr`).
 
 **You cannot define custom parameters in a catalog item.** For example, you cannot create a
```
ygalblum, line 46: Will the creation of the `CatalogItem` fail? Or just disregard the path?

docs/CATALOG_ITEMS.md
```diff
@@ -232,14 +239,16 @@ osac create cluster --catalog-item dev-sandbox \
 CLI flags like `--pull-secret` and `--ssh-public-key` set values in the resource spec. However,
 the server applies `field_definitions` **after** receiving the request, which means:
 
-- If a field is **non-editable** in the catalog item, the CLI flag value is **overridden** by the
-  catalog item's default — the user's value is silently discarded.
+- If a field is **not listed** in `field_definitions`, the server **rejects** the request with
```
ygalblum, line 263: I don't understand this comment. Does this mean that every `CatalogItem` must have `ssh_key` in its field definitions even though it will always be editable without a default?

docs/CATALOG_ITEMS.md
```diff
@@ -232,14 +239,16 @@ osac create cluster --catalog-item dev-sandbox \
...
+  `InvalidArgument`.
+- If a field is **non-editable** in the catalog item and the user provides a value, the server
+  **rejects** the request with `InvalidArgument`.
 - If a field is **editable**, the CLI flag value is accepted and validated against
   `validation_schema` if one is defined.
 - If an editable field is **not provided** by the user, the catalog item's default is applied.
 
 For example, given a catalog item with `ssh_public_key` set as non-editable with a fixed default,
```
ygalblum: While this is true, using `ssh_key` is not a good example. Also, is it `ssh_key` or `ssh_public_key` the doc is inconsistent

internal/servers/catalog_item_validation.go
ygalblum, line 79: This function does not fail is there is no `fieldDefinitions`. But it fails if the user provided a value to a field not listed in the `fieldDefinitions`. As I wrote, I think that's a problem to being with as it requires the CatalogItem to include fields that is has nothing to say about

ygalblum, line 128: Shouldn't this be validated during the creation of the `CatalogItem`? What can the user creating a resource from this catalog item do with this response?

internal/servers/catalog_item_validation.go
```diff
@@ -234,6 +261,34 @@ func setNestedValue(m map[string]any, path string, value any) {
...
+	}
+	return paths
+}
+
+func isPathCovered(path string, allowedPaths map[string]bool) bool {
+	if allowedPaths[path] {
+		return true
+	}
+	for i := range path {
+		if path[i] == '.' && allowedPaths[path[:i]] {
```
ygalblum, line 286: IIUC this check assumes that there is no duplication of suffixes. It will treat `X.Z` and `Y.Z` the same. Am I right? Is that OK?

ygalblum (changes requested):
Most comments revolve the same question. Does this requirement impose additional work on the `CatalogItem` creator? Will they know to add these fields that they don't wish to control at all?

alosadagrande:
> Most comments revolve the same question. Does this requirement impose additional work on the `CatalogItem` creator? Will they know to add these fields that they don't wish to control at all?

Thanks for the review @ygalblum . Let me address your main concern.

The implementation follows the catalog-items EP and [OSAC-1416](https://redhat.atlassian.net/browse/OSAC-1416) (filed by @tzvatot ). The EP explicitly 
defines field_definitions as the complete contract:

_The fields list defines the complete contract between the admin and the user for a given catalog item. Fields not listed in fields are neither editable nor pre-defined; the admin is responsible for ensuring that every field required by the underlying template is covered by a field definition." EP, API Behavior section_
 
So yes, the CatalogItem creator must list every field that the user is expected to provide, even fields they don't want to restrict — that's the EP design. If a field isn't listed, it's not available to the user.

One important thing to note: the previous behavior was silently overwriting user-provided values for non-editable fields and silently accepting unlisted fields. This PR replaces that silent behavior with explicit error messages that tell the user exactly what went wrong:
 
- "fields not allowed by catalog item: pull_secret, release_image" — the user tried to set fields that the admin did not include in field_definitions; previously these were silently accepted, now the request is rejected with the exact field names.
- "field 'release_image' is not editable" — tells the user which specific field they cannot override
- "field 'ssh_key' is required but no value was provided and no default is defined" — surfaces misconfigured catalog items early.

This is actually better UX for both the admin and the user: instead of silently doing the wrong thing, the server now gives actionable feedback.

Regarding validating invalid paths at CatalogItem creation time (your comment on catalog_item_validation.go:127): that would be a good improvement. The EP already calls this out as a known gap in the Risks section, but it's out of scope for this ticket.

You're right that ssh_key is a poor example in the docs — I'll fix that along with the other doc issues (cores/memory_gib → instance_type, ssh_key vs ssh_public_key inconsistency) in a follow-up commit.
 
@tzvatot  can you confirm this matches your intent for [OSAC-1416](https://redhat.atlassian.net/browse/OSAC-1416)?

docs/CATALOG_ITEMS.md
alosadagrande: Currently, the creation of the CatalogItem does not validate paths, an invalid path like vlan is accepted. The error only surfaces when a user tries to create a resource from that CatalogItem. The EP acknowledges this as a known gap in the Risks section and suggests validating paths at CatalogItem creation time as a future improvement. That would be a separate ticket.

docs/CATALOG_ITEMS.md
alosadagrande: Correct, following the EP it must be listed in field_definitions as editable. 

I understand the concern; having to list fields the admin doesn't want to control feels like unnecessary boilerplate. But that's the design the EP defines: the admin explicitly declares every field available to the user. This gives full visibility into what the user can and cannot do with a given catalog item.

If we think this should change (e.g., special fields such as ssh always editable and included in every catalog item), that would be a change to the EP design itself, not to this PR. @tzvatot  what do you think?

internal/servers/catalog_item_validation.go
alosadagrande: The early return when fieldDefinitions is empty (line 79-81) is pre-existing code from the original catalog item implementation (PR #549 ), not introduced by this PR. This PR only adds enforcement for when fieldDefinitions are present, which is the case described by the EP.
 
Whether an empty fieldDefinitions list should be allowed on a CatalogItem is a valid question, but outside the scope of this ticket. cc @tzvatot

alosadagrande: You're right, this is an admin misconfiguration error that ideally should be caught when the CatalogItem is created, not when a user tries to create a resource from it. The user can't do anything useful with this error other than report it to the admin.

I am not sure if adding validation to the CatalogItem creation flow I think it should be part of this PR.  This PR only changes the resource creation path. Now it fails early with a clear message pointing to the root cause.

Validating paths and defaults at CatalogItem creation time would be a good follow-up ticket.

internal/servers/catalog_item_validation.go
alosadagrande: No, the function checks prefixes, not suffixes. It iterates through the path looking for dots and checks if the part before the dot (path[:i]) is in allowedPaths. So X.Z checks if X is allowed, and Y.Z checks if Y is allowed. They're treated independently.

internal/servers/catalog_item_validation.go
ygalblum: I know that this is existing code. But, since the rest of the function was changed, we need to understand if this section should also change. 
@tzvatot and I already had a similar discussion about catalog items without field_definitions in the CLI PR. I think this creates a strange behaviour where if the admin wants to set a single value, they have to explicitly allow setting all the rest (via the `editable` field). But, they can create a fully customizable one without setting anything.

docs/CATALOG_ITEMS.md
ygalblum: I think the order of implementation should have been reversed to make sure CatalogItems are not dead-on-arrival. But, I see your point. Is there a ticket for this?

internal/servers/catalog_item_validation.go
ygalblum: Oh, right. Thanks for clarifying

docs/CATALOG_ITEMS.md
alosadagrande: Agree, I'll reword the example. 

The inconsistency is in the API definition, not the doc:
* https://github.com/osac-project/fulfillment-service/blob/main/proto/private/osac/private/v1/cluster_type.proto#L38
* https://github.com/osac-project/fulfillment-service/blob/main/proto/private/osac/private/v1/compute_instance_type.proto#L72

BMaaS is using ssh_public_key, same as CaaS.

docs/CATALOG_ITEMS.md
alosadagrande: Yes, #691

docs/CATALOG_ITEMS.md
ygalblum: I think we can move this discussion outside of this PR and allow it to merge

internal/servers/catalog_item_validation.go
ygalblum: Let's discuss outside of this PR

ygalblum (approved):
/approve
/lgtm

internal/servers/catalog_item_validation_test.go
```diff
@@ -132,11 +165,8 @@ var _ = Describe("applyFieldDefinitions", func() {
 		Expect(spec.GetIsWindows()).To(BeTrue())
 	})
 
-	It("forces is_windows value for non-editable field definition", func() {
-		falseVal := false
-		spec := &privatev1.ComputeInstanceSpec{
-			IsWindows: &falseVal,
-		}
+	It("applies non-editable default for bool field is_windows on compute instance spec", func() {
```
tzvatot, line 168: 🟡 **Missing test: non-editable boolean field with explicit `false` value**

The old test "forces is_windows value" set `IsWindows: &falseVal` and verified override. This replacement only tests default application when no user value is provided. There's no corresponding test that sets `is_windows: false` and expects `InvalidArgument` rejection.

This matters because protojson serialization of `false` depends on field optionality:
- **Optional `*bool`**: `false` serializes to `"is_windows": false` - caught by `userHasValue && userVal != nil`
- **Non-optional `bool`**: `false` is zero value, protojson omits it - silently passes as "no value"

If the proto field later changes from optional to non-optional, the rejection would silently stop working for `false` values.

**Suggestion:** Add a companion test:
```go
It("rejects user value false for non-editable bool field is_windows", func() {
    falseVal := false
    spec := &privatev1.ComputeInstanceSpec{
        IsWindows: &falseVal,
    }
    defaultVal, err := structpb.NewValue(true)
    Expect(err).ToNot(HaveOccurred())
    fieldDefs := []*privatev1.FieldDefinition{{
        Path:     "is_windows",
        Editable: false,
        Default:  defaultVal,
    }}
    err = applyFieldDefinitions(spec, fieldDefs)
    Expect(err).To(HaveOccurred())
    Expect(status.Code(err)).To(Equal(codes.InvalidArgument))
    Expect(err.Error()).To(ContainSubstring("not editable"))
})
```

internal/servers/catalog_item_validation.go
```diff
@@ -234,6 +262,34 @@ func setNestedValue(m map[string]any, path string, value any) {
 	}
 }
 
+func collectLeafPaths(m map[string]any, prefix string) []string {
```
tzvatot, line 265: 💡 **`collectLeafPaths` silently passes empty nested objects**

When a user sends a nested message with no fields set (e.g., `"network": {}`), this function recurses into the empty map and returns no paths. The empty nested object passes the unlisted field check without a corresponding `field_definition`.

Low practical impact since protojson omits empty sub-messages for non-optional fields, but for optional message fields that can serialize as `{}`, this is a gap in the "complete contract" enforcement.

Consider either treating empty nested maps as leaf paths (emit the parent path when the nested map has zero keys), or documenting this as accepted behavior.

tzvatot (commented):
## Review Summary

Solid implementation. `field_definitions` is now enforced as a strict allowlist for catalog item-based resource creation: unlisted spec fields are rejected, and user-provided values for non-editable fields return `InvalidArgument` instead of being silently overridden. The validation helpers (`collectLeafPaths`, `isPathCovered`) are well-designed, error messages are deterministic, and test coverage is thorough.

| Category | Count |
|----------|-------|
| 🔴 Critical | 0 |
| 🟡 Important | 1 |
| 💡 Suggestion | 1 |

See inline comments for details.

**Design note (non-blocking):** ygalblum's concern about the cliff behavior (empty `field_definitions` allows everything, one entry makes it a strict allowlist) is valid UX feedback worth discussing at the EP level, but not in scope for this PR.

PR #784, opened 2026-06-26
NO-ISSUE: add bash completion for object types in osac get

adriengentil:
## Summary

- Add `ValidArgsFunction` to `osac get` so that pressing Tab after `osac get` returns the list of available resource types
- Resource types are discovered via gRPC server reflection (same mechanism the `get` command itself uses), returning both singular and plural forms (e.g. `baremetalinstance`, `baremetalinstances`, `computeinstance`, `computeinstances`, ...)
- Falls back silently to no completions when the user is not logged in or the server is unreachable

## Usage

After sourcing the completion script:

```bash
source <(osac completion bash)   # bash
source <(osac completion zsh)    # zsh
```

```
$ osac get <TAB>
baremetalinstance   baremetalinstances   cluster   clusters   computeinstance   computeinstances   ...
```

## Jira

N/A

## Test plan

- [x] `go build ./cmd/osac/` clean
- [ ] Manual: `source <(osac completion bash)` then `osac get <TAB>` returns resource types

### DIFF
--- a/internal/cmd/cli/get/get_cmd.go
+++ b/internal/cmd/cli/get/get_cmd.go
@@ -61,6 +61,7 @@ func Cmd() *cobra.Command {
 		Short:                 shortHelp,
 		Long:                  longHelp,
 		RunE:                  runner.run,
+		ValidArgsFunction:     runner.complete,
 	}
 	result.AddCommand(kubeconfig.Cmd())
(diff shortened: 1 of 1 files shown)


adriengentil:
doesn't seem to be the right way

PR #787, opened 2026-06-26
OSAC-1284: Update Event and Database Migration for ProjectMembership

CrystalChun:
# Description

Relevant changes:
-  Extends the Event message to support `ProjectMembership` as a payload type
- Creates migration `67 with project_memberships` and `archived_project_memberships` tables following the standard DAO schema.
- Enforce unique memberships to prevent users from having multiple memberships in the same project

ProjectMembership enables assigning and unassigning users from projects.

# Testing

- [x] buf lint passes
- [x] go build passes
- [x] unit tests pass

/cc @jhernand 

## Summary by CodeRabbit

* **New Features**
  * Added support for project memberships, including archived records and related event payloads.
  * Membership data now enforces uniqueness per tenant, project, and user, with clearer conflict handling.

* **Bug Fixes**
  * Improved database error messages for duplicate records so users see more helpful details.
  * Updated database validation to better exclude non-object tables from schema checks.


internal/database/migrations/67_create_project_memberships_tables.up.sql
```diff
@@ -0,0 +1,104 @@
...
+  v_user := new.data->'spec'->>'user';
+
+  -- Insert the new tuple, catching duplicates:
+  begin
+    insert into project_membership_subjects (tenant, project, username, membership_id)
+      values (new.tenant, v_project, v_user, new.id);
+  exception when unique_violation then
+    raise exception using
+      errcode = 'Z0004',
+      message = format('user ''%s'' already has a membership in project ''%s'' within tenant ''%s''', v_user, v_project, new.tenant);
```
tzvatot: 🔴 **Error code mismatch: trigger raises Z0004 but DAO handles Z0005**

This trigger raises `errcode = 'Z0004'` (`errNotUniqueCode`), but the new DAO changes in `dao_errors.go` define a separate `errDuplicateMembershipCode = "Z0005"` with handlers in `generic_dao_create.go` and `generic_dao_update.go`.

Since the trigger raises Z0004, the *existing* `errNotUniqueCode` handler catches it and returns `ErrNotUnique` (the raw trigger message). The new Z0005 handler is dead code and `ErrAlreadyExists` is never returned for duplicate memberships.

**Fix**: Either change the trigger to raise `Z0005` to match the DAO handler, or remove the new error code and rely on the existing Z0004/`ErrNotUnique` path (which is what `tenant_domains` uses).

internal/database/database_tool.go
```diff
@@ -506,7 +506,8 @@ func (t *tool) listObjectTables(ctx context.Context, pool *pgxpool.Pool) (result
 				'notifications',
 				'schema_migrations',
 				'tenant_domains'
-			)
+			) and
+			c.relname not in ('notifications', 'schema_migrations', 'project_membership_subjects')
```
tzvatot: 🟡 **Redundant NOT IN clause**

This adds a second `NOT IN` that duplicates `notifications` and `schema_migrations` from the clause above. Just add `'project_membership_subjects'` to the existing list:

```sql
c.relname not in (
    'notifications',
    'schema_migrations',
    'tenant_domains',
    'project_membership_subjects'
)
```

tzvatot (changes requested):
## Review Summary

This PR establishes the database foundation for ProjectMembership - overall follows the established patterns well (materialized helper table + trigger from migration 66). However, there is a critical error code mismatch that will cause the wrong error type at runtime.

| Category | Count |
|----------|-------|
| :red_circle: Critical | 1 |
| :yellow_circle: Important | 1 |
| :bulb: Suggestion | 1 |

### :bulb: AI attribution trailer

The HEAD commit uses `Co-Authored-By` for AI attribution. Per Red Hat convention, use `Assisted-by: Claude Code <noreply@anthropic.com>` instead.

### What looks good

- Migration pattern matches the established `tenant_domains` pattern from migration 66
- Thorough test coverage: table creation, insert/query, indexes, FK enforcement, cross-tenant isolation, column metadata, defaults, and duplicate rejection
- Proto change is minimal and correctly scoped
- The backfill line (`update project_memberships set data = data`) is a smart way to trigger materialization for existing rows

internal/database/migrations/67_create_project_memberships_tables.up.sql
```diff
@@ -0,0 +1,104 @@
...
+-- Project memberships must belong to a specific organization (tenant).
+alter table project_memberships add constraint project_memberships_tenant_fk foreign key (tenant) references tenants (id);
+
+-- Helper table to enforce uniqueness of (tenant, project, user) tuples.
+-- A user can only have one membership per project within a tenant.
+create table project_membership_subjects (
+  tenant text not null,
+  project text not null,
+  username text not null,
+  membership_id text not null references project_memberships(id) on delete cascade,
```
jhernand: In other places we don't use the `..._id` suffix. Can we rename this to just `membership` to make it more consistent?

internal/database/migrations/67_create_project_memberships_tables.up.sql
```diff
@@ -0,0 +1,104 @@
...
+
+-- Project memberships must belong to a specific organization (tenant).
+alter table project_memberships add constraint project_memberships_tenant_fk foreign key (tenant) references tenants (id);
+
+-- Helper table to enforce uniqueness of (tenant, project, user) tuples.
+-- A user can only have one membership per project within a tenant.
+create table project_membership_subjects (
+  tenant text not null,
+  project text not null,
+  username text not null,
```
jhernand: I think this should be just `user`, as it is a reference to the user. It may be the user name, or the user identifier, but that shouldn't affect the name of this column. That way we can decide what is the best unique identifier of a user without changing the name of this column.

internal/database/dao/dao_errors.go
```diff
@@ -178,4 +178,7 @@ const (
 	// relationship between organizations and domains returns this code when a domain is already assigned to
 	// another tenant.
 	errNotUniqueCode = "Z0004"
+	// errDuplicateMembershipCode is the SQLSTATE error code returned by the 'materialize_project_membership_subjects'
+	// trigger when an insert or update attempts to create a duplicate (tenant, project, user) tuple.
+	errDuplicateMembershipCode = "Z0005"
```
jhernand: I think we can reuse `errNotUniqueCode` for this.

internal/database/migrations/67_create_project_memberships_tables.up.sql
jhernand: Can we remove the `within tenant ...` part? Regular users will only see a tenant, so we don't need to tell them what tenant it is.

What would be really helpful is to include in the message the name of the existing membership, something like `user 'my-user' is already a member of project 'my-project' via membership 'my-membership'`. Is that doable?

internal/database/dao/generic_dao_create.go
```diff
@@ -218,6 +218,10 @@ func (r *CreateRequest[O]) translateError(ctx context.Context, id, name, tenant
 		return &ErrReference{
 			Reason: pgErr.Message,
 		}
+	case errDuplicateMembershipCode:
+		return &ErrAlreadyExists{
+			ID: id,
+		}
```
jhernand: Aren't we loosing the nice error message that we prepared in the trigger? Can we check if if the database has populated the `Message` field of the error and copy it to a new `Reason` field of `ErrAlreadyExists`? We can then change the `String` method of `ErrAlreadyExists` to return that, and the gRPC server will already translate that into the appropriate gRPC error.

Also, can you add a unit test to verify this behavior, including that it generates the expected error message?

jhernand (changes requested):

internal/database/migrations/67_create_project_memberships_tables.up.sql
CrystalChun: Thanks for catching this! Updated it to use the existing error code

internal/database/database_tool.go
CrystalChun: Modified as suggested. Thanks for catching this!

internal/database/migrations/67_create_project_memberships_tables.up.sql
CrystalChun: Yes thank you! Removed this suffix

internal/database/migrations/67_create_project_memberships_tables.up.sql
CrystalChun: That makes sense! Thanks Juan, updated it to be `user`

internal/database/dao/dao_errors.go
CrystalChun: Removed this and reused the not unique code. Thank you!

internal/database/migrations/67_create_project_memberships_tables.up.sql
CrystalChun: That makes sense for the tenant, removed that part and modified the message to include project and membership.
Thank you!

internal/database/dao/generic_dao_create.go
CrystalChun: Thank you for the suggestion! I've attempted to do this, but I'm not sure it's fully correct. How does it look?

internal/database/dao/generic_dao_create.go
```diff
@@ -233,6 +233,13 @@ func (r *CreateRequest[O]) translateError(ctx context.Context, id, name, tenant
 			return &ErrReference{}
 		}
 	case errNotUniqueCode:
+		// Project membership uniqueness violations should return ErrAlreadyExists with the custom message
+		if strings.Contains(pgErr.Message, "is already a member of project") {
```
jhernand: I think we shouldn't check here for a substring of the message: it is brittle. As this will only happen when we explicilty set the Z.... error code I think we can assume that the message will also be something that we set explicitly. So we can just check if `pgErr.Message` is set, and if it is then set `Reason` on the error. Otherwise we set `Id`.

jhernand (approved):

CrystalChun:
Juan has already approved, merging.

PR #791, opened 2026-06-26
OSAC-1812: Add keycloak_user_id field to User status

CrystalChun:
# Description
Add keycloak_user_id to UserStatus to store the identity provider's user UUID. This field is populated during JIT provisioning and used by controllers to identify users when making IDP API calls.

# Testing

- [x] buf lint passes
- [x] go build passes
- [x] unit tests pass

/cc @jhernand


proto/private/osac/private/v1/user_type.proto
```diff
@@ -76,6 +76,10 @@ message UserStatus {
 
   // Additional conditions.
   repeated UserCondition conditions = 2;
+
+  // The Keycloak user ID for this user in the identity provider.
+  // This is set when the user is created or imported and is used for role assignments.
+  string keycloak_user_id = 3;
```
DakCrowder, line 82: Would it make more sense to prefix this key with `idp` rather than `keycloak` to be more agnostic as we onboard more identity providers?  Or is it crucial we have an understanding this is an id specifically from keycloak?

CrystalChun: It's an ID specifically for keycloak, and it was decided we should only stick with keycloak. I'll need to refactor the client to make it more apparent 😅 

@jhernand wdyt?

jhernand: I think we need to acknowledge that we depend completely on Keycloak. We will never support other identity providers directly; that support will be via Keycloak using it as an OIDC broker.

jhernand (approved):

PR #794, opened 2026-06-29
OSAC-1679: Add E2E full install presubmit job

eliorerz:
## Summary
- Add `e2e-vmaas-full-install.yml` workflow that triggers on PRs to `main` and `workflow_dispatch`
- Calls the reusable workflow in `osac-test-infra` to run a full OSAC install + E2E tests
- Builds the fulfillment-service image from the PR branch and loads it onto the test cluster
- Uses `component-image-key: service.images.service` for Helm override

## Test plan
- [ ] Verify workflow triggers on PR to main
- [ ] Verify `workflow_dispatch` works with optional inputs (test-suite, test-filter, test-infra-ref)
- [ ] Confirm the reusable workflow builds the PR's code (not main branch)

Jira: https://redhat.atlassian.net/browse/OSAC-1679

## Summary by CodeRabbit

* **New Features**
  * Added an automated end-to-end full-install test workflow for pull requests targeting the main branch.
  * Supports manual runs with options to select a test suite, apply a `pytest -k` filter, and choose a test reference.
  * Prevents overlapping runs for the same pull request by canceling older in-progress executions.

### DIFF
--- a/.github/workflows/e2e-vmaas-full-install.yml
+++ b/.github/workflows/e2e-vmaas-full-install.yml
@@ -1,0 +1,41 @@
+---
+name: E2E VMaaS Full Install
+
+permissions:
+  contents: read
+
+on:
+  pull_request:
+    branches: [main]
+  workflow_dispatch:
+    inputs:
+      test-suite:
+        description: "Test suite (vmaas, caas, or empty for all)"
+        required: false
+        default: "vmaas"
+      test-filter:
+        description: "pytest -k filter"
+        required: false
+        default: ""
+      test-infra-ref:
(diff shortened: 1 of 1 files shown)


jhernand (changes requested):
I'd suggest you consider not replacing the image, but replacing the helm chart. I mean, build the helm fulfillment-service helm chart from the branch, and then use it instead of whatever is included with the installer. Bulding that chart should in turn build the image. I think it is better to assume that you can change the image without also changing the helm chart: the contract of the fulfillment service is its Helm chart, not its image.

PR #805, opened 2026-06-30
OSAC-1708: Update RoleBinding controller to extract IDP user ID from User status

CrystalChun:
# Description

Change RoleBinding controller to fetch the OSAC User object and extract the IDP user ID from status.keycloak_user_id before calling IDP client methods. 

# Testing

- [x] go unit-test passes
- [x] go build passes

/cc @jhernand 

Requires https://github.com/osac-project/fulfillment-service/pull/791


jhernand (approved):

CrystalChun:
Was already approved, merging

PR #814, opened 2026-07-01
OSAC-1477: Implement ExternalIPAttachment servers with cross-resource validation

danmanor:
## Summary

- Add public and private ExternalIPAttachment gRPC servers with full CRUD + Signal operations
- Support three target types: ComputeInstance, Cluster, and BareMetalInstance via oneof target field
- Cross-resource validation: ExternalIP existence/ALLOCATED state/attached flag, target resource existence, ExternalIP uniqueness, target_endpoint required for cluster targets
- Immutability enforcement for all spec fields (external_ip, target, target_endpoint)
- ExternalIP attached flag management on create/delete
- Register both servers in gRPC startup

## Test plan

- [x] Unit tests for private server: create with all 3 target types, list, get, update metadata, delete, signal
- [x] Unit tests for validation: nil object, nil spec, missing external_ip, no target, missing target_endpoint for cluster, target_endpoint set for non-cluster
- [x] Unit tests for ExternalIP reference: not found, not ALLOCATED, already attached
- [x] Unit tests for target reference: ComputeInstance/Cluster/BareMetalInstance not found
- [x] Unit tests for uniqueness: ExternalIP already has attachment
- [x] Unit tests for immutable fields: external_ip, compute_instance changes rejected
- [x] Unit tests for attached flag: set to true on create, false on delete
- [x] Unit tests for public server: create, list, get, update, delete, immutable field propagation
- [x] All 1253 server package tests pass
- [x] Build passes, gofmt clean

Jira: https://redhat.atlassian.net/browse/OSAC-1477

## Summary by CodeRabbit

* **New Features**
  * Added gRPC API support for external IP attachments via both public and private access paths.
  * Supports listing, getting, creating, updating metadata/labels (with field mask), deleting, and private signaling for attachments.
* **Bug Fixes**
  * Added stronger validation for attachment creation and updates, including correct target rules, immutable fields, and uniqueness checks.
  * Attachment lifecycle now keeps external IP attachment status synchronized (attached on create, detached on delete).
* **Tests**
  * Added end-to-end test coverage for public and private attachment server behavior.

### DIFF
--- a/internal/servers/external_ip_attachments_server.go
+++ b/internal/servers/external_ip_attachments_server.go
@@ -1,0 +1,279 @@
+/*
+Copyright (c) 2026 Red Hat Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with the
+License. You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific
+language governing permissions and limitations under the License.
+*/
+
+package servers
+
+import (
+	"context"
+	"errors"
+	"log/slog"
+
+	"github.com/prometheus/client_golang/prometheus"
+	grpccodes "google.golang.org/grpc/codes"
+	grpcstatus "google.golang.org/grpc/status"
+
+	privatev1 "github.com/osac-project/fulfillment-service/internal/api/osac/private/v1"
+	publicv1 "github.com/osac-project/fulfillment-service/internal/api/osac/public/v1"
+	"github.com/osac-project/fulfillment-service/internal/auth"
+	"github.com/osac-project/fulfillment-service/internal/events"
+)
+
+type ExternalIPAttachmentsServerBuilder struct {
+	logger            *slog.Logger
+	notifier          events.Notifier
+	attributionLogic  auth.AttributionLogic
+	tenancyLogic      auth.TenancyLogic
+	metricsRegisterer prometheus.Registerer
+}
+
+var _ publicv1.ExternalIPAttachmentsServer = (*ExternalIPAttachmentsServer)(nil)
+
+type ExternalIPAttachmentsServer struct {
+	publicv1.UnimplementedExternalIPAttachmentsServer
+
+	logger    *slog.Logger
+	delegate  privatev1.ExternalIPAttachmentsServer
+	inMapper  *GenericMapper[*publicv1.ExternalIPAttachment, *privatev1.ExternalIPAttachment]
+	outMapper *GenericMapper[*privatev1.ExternalIPAttachment, *publicv1.ExternalIPAttachment]
+}
+
+func NewExternalIPAttachmentsServer() *ExternalIPAttachmentsServerBuilder {
+	return &ExternalIPAttachmentsServerBuilder{}
+}
+
+func (b *ExternalIPAttachmentsServerBuilder) SetLogger(value *slog.Logger) *ExternalIPAttachmentsServerBuilder {
+	b.logger = value
+	return b
+}
+
+func (b *ExternalIPAttachmentsServerBuilder) SetNotifier(value events.Notifier) *ExternalIPAttachmentsServerBuilder {
+	b.notifier = value
+	return b
+}
+
+func (b *ExternalIPAttachmentsServerBuilder) SetAttributionLogic(value auth.AttributionLogic) *ExternalIPAttachmentsServerBuilder {
+	b.attributionLogic = value
+	return b
+}
+
+func (b *ExternalIPAttachmentsServerBuilder) SetTenancyLogic(value auth.TenancyLogic) *ExternalIPAttachmentsServerBuilder {
+	b.tenancyLogic = value
+	return b
+}
+
+func (b *ExternalIPAttachmentsServerBuilder) SetMetricsRegisterer(value prometheus.Registerer) *ExternalIPAttachmentsServerBuilder {
+	b.metricsRegisterer = value
+	return b
+}
+
+func (b *ExternalIPAttachmentsServerBuilder) Build() (result *ExternalIPAttachmentsServer, err error) {
+	if b.logger == nil {
+		err = errors.New("logger is mandatory")
+		return
+	}
+	if b.tenancyLogic == nil {
+		err = errors.New("tenancy logic is mandatory")
+		return
+	}
+	if b.attributionLogic == nil {
+		err = errors.New("attribution logic is mandatory")
+		return
+	}
+
+	inMapper, err := NewGenericMapper[*publicv1.ExternalIPAttachment, *privatev1.ExternalIPAttachment]().
+		SetLogger(b.logger).
+		SetStrict(true).
+		Build()
+	if err != nil {
+		return
+	}
+	outMapper, err := NewGenericMapper[*privatev1.ExternalIPAttachment, *publicv1.ExternalIPAttachment]().
+		SetLogger(b.logger).
+		SetStrict(false).
+		Build()
+	if err != nil {
+		return
+	}
+
+	delegate, err := NewPrivateExternalIPAttachmentsServer().
+		SetLogger(b.logger).
+		SetNotifier(b.notifier).
+		SetAttributionLogic(b.attributionLogic).
+		SetTenancyLogic(b.tenancyLogic).
+		SetMetricsRegisterer(b.metricsRegisterer).
+		Build()
+	if err != nil {
+		return
+	}
+
+	result = &ExternalIPAttachmentsServer{
+		logger:    b.logger,
+		delegate:  delegate,
+		inMapper:  inMapper,
+		outMapper: outMapper,
+	}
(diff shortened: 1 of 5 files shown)


danmanor:
CI failure is a flaky test in `internal/auth/auth_jwt_validator_test.go:557` ("Removes expired tokens from the cache" — timing-sensitive token expiry). Unrelated to this PR's changes. Re-running unit tests.

internal/servers/external_ip_attachments_server.go
```diff
@@ -0,0 +1,279 @@
...
+	_, err := s.delegate.Delete(ctx, privateRequest)
+	if err != nil {
+		return nil, err
+	}
+
+	response := &publicv1.ExternalIPAttachmentsDeleteResponse{}
+	return response, nil
+}
+
+func (s *ExternalIPAttachmentsServer) attachmentFromPrivate(ctx context.Context, privateObj *privatev1.ExternalIPAttachment) (*publicv1.ExternalIPAttachment, error) {
```
jhernand: Why is this needed? Isn't the generic mapper able to do this alreaydy?

internal/servers/external_ip_attachments_server.go
```diff
@@ -0,0 +1,279 @@
...
+			ctx,
+			"Failed to map private external IP attachment to public",
+			slog.Any("error", err),
+		)
+		return nil, err
+	}
+	return publicObj, nil
+}
+
+func (s *ExternalIPAttachmentsServer) attachmentToPrivate(ctx context.Context, publicObj *publicv1.ExternalIPAttachment) (*privatev1.ExternalIPAttachment, error) {
```
jhernand: Same, the generic mapper should be able to do this.

internal/servers/private_external_ip_attachments_server.go
```diff
@@ -0,0 +1,543 @@
...
+		var notFoundErr *dao.ErrNotFound
+		if errors.As(err, &notFoundErr) {
+			return grpcstatus.Errorf(grpccodes.InvalidArgument,
+				"ExternalIP '%s' does not exist", externalIPID)
+		}
+		s.logger.ErrorContext(ctx, "Failed to query ExternalIP",
+			slog.String("external_ip_id", externalIPID),
+			slog.Any("error", err))
+		return grpcstatus.Errorf(grpccodes.Internal, "failed to validate external_ip")
+	}
```
jhernand, line 392: What happens if the external IP is deleted or modified after this check, but before this transaction completes? Consider using the DAO `Lock` method:

```go
getResponse, err := s.externalIPDao.Get().
        SetId(externalIPID).
        SetLock(true).
        Do(ctx)
```

That will use `select ... for update` so the row will be locked for the duration of the transaction.

internal/servers/private_external_ip_attachments_server.go
```diff
@@ -0,0 +1,543 @@
...
+		return grpcstatus.Errorf(grpccodes.InvalidArgument,
+			"exactly one target must be set (compute_instance, cluster, or baremetal_instance)")
+	}
+}
+
+func (s *PrivateExternalIPAttachmentsServer) validateComputeInstanceReference(
+	ctx context.Context, computeInstanceID string) error {
+	_, err := s.computeInstanceDao.Get().
+		SetId(computeInstanceID).
+		Do(ctx)
```
jhernand, line 430: Same, consider using `SetLock(true)`.

internal/servers/private_external_ip_attachments_server.go
```diff
@@ -0,0 +1,543 @@
...
+		return grpcstatus.Errorf(grpccodes.Internal, "failed to validate cluster")
+	}
+	return nil
+}
+
+func (s *PrivateExternalIPAttachmentsServer) validateBareMetalInstanceReference(
+	ctx context.Context, bareMetalInstanceID string) error {
+	_, err := s.bareMetalInstanceDao.Get().
+		SetId(bareMetalInstanceID).
+		Do(ctx)
```
jhernand, line 470: Same, consider using `SetLock(true)`.

internal/servers/private_external_ip_attachments_server.go
```diff
@@ -0,0 +1,543 @@
...
+	case spec.HasBaremetalInstance():
+		return spec.GetBaremetalInstance()
+	default:
+		return ""
+	}
+}
+
+func (s *PrivateExternalIPAttachmentsServer) validateUniqueness(
+	ctx context.Context, externalIPID string, targetID string) error {
+	eipFilter := "this.spec.external_ip == " + strconv.Quote(externalIPID)
```
jhernand: Consider using the `%q` placeholder of `fmt.Sprintf`:

```go
eipFilter := fmt.Sprintf("this.spec.external_ip == %q", externalIPid)
```

It makes it more readable, and easier to add other criteria to the filter.

internal/servers/private_external_ip_attachments_server.go
```diff
@@ -0,0 +1,543 @@
...
+		SetFilter(eipFilter).
+		SetLimit(1).
+		Do(ctx)
+	if err != nil {
+		s.logger.ErrorContext(ctx, "Failed to list attachments for uniqueness check",
+			slog.String("external_ip_id", externalIPID),
+			slog.Any("error", err))
+		return grpcstatus.Errorf(grpccodes.Internal, "failed to validate uniqueness")
+	}
+	if eipResp.GetSize() > 0 {
```
jhernand: In this kind of checks in general you don't care about the returned items, which is what `GetSize` is about: the actual returned number of items. It would be better to use `GetTotal` which is the total of number of items in the database that satisfy the condition. In the future we plan to change the DAO so that `SetLimit(0)` will return no items. Then you will want to use `SetLimit(0)` to avoid fetching the item, as you are only interested in the count. If you do that you will need to change from `GetSize` to `GetTotal`, so better if you do it now.

internal/servers/private_external_ip_attachments_server.go
```diff
@@ -0,0 +1,543 @@
...
+}
+
+func (s *PrivateExternalIPAttachmentsServer) validateUniqueness(
+	ctx context.Context, externalIPID string, targetID string) error {
+	eipFilter := "this.spec.external_ip == " + strconv.Quote(externalIPID)
+	eipResp, err := s.externalIPAttachmentDao.List().
+		SetFilter(eipFilter).
+		SetLimit(1).
+		Do(ctx)
+	if err != nil {
```
jhernand, line 506: What happens if external IP attachments for the same external IP are added or after this check but before this transaction finishes? There is no `SetLock` for `List` operations.

internal/servers/private_external_ip_attachments_server.go
```diff
@@ -0,0 +1,543 @@
...
+	}
+
+	return nil
+}
+
+func (s *PrivateExternalIPAttachmentsServer) updateExternalIPAttachedFlag(
+	ctx context.Context, externalIPID string, attached bool) error {
+	getResponse, err := s.externalIPDao.Get().
+		SetId(externalIPID).
+		Do(ctx)
```
jhernand, line 525: Consider using `SetLock(true)`.

jhernand (changes requested):

internal/servers/external_ip_attachments_server.go
danmanor: Good point — removed the helpers and inlined the mapper calls directly, matching the ExternalIP server pattern. Fixed in 4bae2b5.

internal/servers/external_ip_attachments_server.go
danmanor: Done, same fix.

internal/servers/private_external_ip_attachments_server.go
danmanor: Great catch — added `SetLock(true)` on all validation DAO queries and on `updateExternalIPAttachedFlag`. Fixed in 4bae2b5.

internal/servers/private_external_ip_attachments_server.go
danmanor: Done.

internal/servers/private_external_ip_attachments_server.go
danmanor: Done.

internal/servers/private_external_ip_attachments_server.go
danmanor: Good suggestion, switched to `fmt.Sprintf` with `%q`. Fixed in 4bae2b5.

internal/servers/private_external_ip_attachments_server.go
danmanor: Makes sense — switched to `GetTotal`. Fixed in 4bae2b5.

internal/servers/private_external_ip_attachments_server.go
danmanor: Good question. The database already has a unique index (`external_ip_attachments_one_per_external_ip`) from migration 61, so concurrent creates would fail at the DB level. The application-level check gives a friendlier error message, but the DB constraint is the real guard.

internal/servers/private_external_ip_attachments_server.go
danmanor: Done. Fixed in 4bae2b5.

jhernand (approved):
Thanks @danmanor, looks good now.

PR #815, opened 2026-07-01
OSAC-1451: Add fabric_manager and k8s_manager validation to NetworkClass server

danmanor:
## Summary

- Add `fabric_manager` as a required field on NetworkClass Create (NC-VAL-05)
- Add `fabric_manager` immutability check on NetworkClass Update (NC-VAL-06)
- Add `k8s_manager` immutability check on NetworkClass Update (NC-VAL-07) — immutable once set, but can be set for the first time to support adding VM support to a BM-only region
- Add `fabric_manager` and `k8s_manager` field mask paths to `applyNetworkClassUpdate`
- Update all existing test helpers and DAO builders to include `fabric_manager`
- Add 10 new tests covering manager field validation and field mask handling

## Context

Part of the **NetworkClass Two-Manager Model** epic ([OSAC-1438](https://redhat.atlassian.net/browse/OSAC-1438)). This task implements server-side validation for the two-manager fields added in [OSAC-1449](https://redhat.atlassian.net/browse/OSAC-1449) (proto) and [OSAC-1452](https://redhat.atlassian.net/browse/OSAC-1452) (DB migration).

**Jira:** [OSAC-1451](https://redhat.atlassian.net/browse/OSAC-1451)

## Test plan

- [x] All 10 new manager field tests pass
- [x] All 28 existing NetworkClass server tests pass (no regressions)
- [x] Full unit test suite (71 suites) passes
- [x] `buf lint` clean
- [x] `golangci-lint` — 0 issues

requires - https://github.com/osac-project/osac-aap/pull/388

### DIFF
--- a/internal/servers/network_classes_server_test.go
+++ b/internal/servers/network_classes_server_test.go
@@ -103,6 +103,7 @@ var _ = Describe("Network classes server", func() {
 				Object: privatev1.NetworkClass_builder{
 					Title:                  "Test Network Class",
 					ImplementationStrategy: "ovn-kubernetes",
+					FabricManager:          "netris",
 				}.Build(),
 			}.Build())
 			Expect(err).ToNot(HaveOccurred())
@@ -115,6 +116,7 @@ var _ = Describe("Network classes server", func() {
 				Object: privatev1.NetworkClass_builder{
 					Title:                  "Default Network Class",
 					ImplementationStrategy: "ovn-kubernetes",
+					FabricManager:          "netris",
 					IsDefault:              new(true),
 				}.Build(),
 			}.Build())
@@ -264,6 +266,7 @@ var _ = Describe("Network classes server", func() {
 					Id:                     callerProvidedId,
 					Title:                  "Test Network Class",
 					ImplementationStrategy: "ovn-kubernetes",
+					FabricManager:          "netris",
 				}.Build(),
 			}.Build())
 			Expect(err).ToNot(HaveOccurred())
@@ -479,6 +482,7 @@ var _ = Describe("Network classes server", func() {
 					SetObject(privatev1.NetworkClass_builder{
 						Title:                  "NC-A",
 						ImplementationStrategy: "ovn-kubernetes",
+						FabricManager:          "netris",
 						IsDefault:              new(true),
 						Metadata: privatev1.Metadata_builder{
 							Tenant: auth.SharedTenant,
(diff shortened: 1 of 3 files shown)


AlonaKaplan (commented):
## Review

This PR adds server-side validation for the two-manager model: `fabric_manager` required + immutable, `k8s_manager` immutable once set. Includes field mask handling and 10 new tests. Code looks clean and follows the existing validation pattern in `validateNetworkClass`.

### Finding

**`internal/servers/network_classes_server.go:215` — Public API Create will always fail because `fabric_manager` is not in the public proto**

The public `NetworkClassesServer.Create` maps public NetworkClass to private via `inMapper`, then delegates to the private server. Since `fabric_manager` doesn't exist in the public proto (`proto/public/osac/public/v1/network_class_type.proto`), the mapped object will always have an empty `fabric_manager`, failing NC-VAL-05. The public service exposes Create/Update/Delete RPCs — any public Create call now returns `InvalidArgument: field 'fabric_manager' is required`. Either the public Create RPC is intentionally unused (and should be removed or documented as admin-only), or the validation needs to account for the public API path.

AlonaKaplan (approved):
Thanks!

AlonaKaplan (changes requested):
Approved by mistake. Please review the comment first:)

danmanor:
Good catch Alona. You're right that public Create will always fail because `fabric_manager` doesn't exist in the public proto.

The thing is, this is actually a bigger issue than just Create. Looking at the unified networking PRD (requirement R6), it explicitly says "NetworkClass is not exposed in the tenant API", and the EP says "tenants never interact with it" and "provider-only, tenants never see it". So the entire public NetworkClasses service (Create, Get, List, Update, Delete) shouldn't exist at all, not just the mutation RPCs.

This requirement was missed when the EP was decomposed into Jira tasks. I created [OSAC-1980](https://redhat.atlassian.net/browse/OSAC-1980) under the Migration and Deprecation epic ([OSAC-1441](https://redhat.atlassian.net/browse/OSAC-1441)) to track removing the full public service, proto types, and CLI subcommands.

For this PR specifically, the validation logic is correct for the private API path where NetworkClasses are actually created and managed by admins. The public path being broken is a pre-existing gap, not something introduced here. This PR just makes it fail loudly instead of silently creating NetworkClasses without a `fabric_manager`.

AlonaKaplan (approved):

PR #818, opened 2026-07-01
OSAC-1832: Populate node sets from template in catalog item cluster creation

tzvatot:
## [OSAC-1832](https://redhat.atlassian.net/browse/OSAC-1832): Populate node sets from template in catalog item cluster creation

**Jira:** https://redhat.atlassian.net/browse/OSAC-1832
**Type:** Bug fix

### Summary

When a cluster is created using `--catalog-item`, the fulfillment-service creates a cluster without populating `spec.node_sets` from the referenced template. The cluster controller then creates a ClusterOrder CR with empty `nodeRequests`, causing the AAP playbook to fall back to wrong defaults (wrong resource class and node count).

This fix makes the catalog item code path look up the template and apply the same template-driven operations as the direct template path: spec defaults, host type resolution, node set merging, and template parameter validation.

### Changes

- **`internal/servers/private_clusters_server.go`**: Extended `validateAndTransformCatalogItem` to look up the template after applying field definitions, then apply spec defaults, look up and index host types, validate and merge node sets, and validate/default template parameters - matching the behavior of `validateAndTransformCluster`.

- **`internal/servers/private_clusters_server_test.go`**: Updated the happy-path catalog item test to verify `node_sets` are populated from the template (both entries, correct host type IDs and sizes). Added a new test for spec defaults applied via the catalog item path using a template with `SpecDefaults`.

### Testing

- **Unit tests:** 2 test cases added/updated covering node set population and spec defaults via catalog item
- **Integration tests:** N/A - fix is in the server layer's request transformation, fully testable through unit tests
- **Coverage:** `validateAndTransformCatalogItem` at 80%, `mergeNodeSetsWithTemplate` at 100%, all 1137 server tests pass

### Acceptance Criteria

- [x] `spec.node_sets` populated from template's `node_sets` when creating a cluster via catalog item
- [x] ClusterOrder CR contains correct `spec.nodeRequests` derived from the template's node sets
- [x] Template spec defaults (pull_secret, ssh_public_key, release_image, network) applied when using catalog item path

### DIFF
--- a/internal/servers/private_clusters_server.go
+++ b/internal/servers/private_clusters_server.go
@@ -758,19 +758,16 @@ func (s *PrivateClustersServer) validateAndTransformCatalogItem(ctx context.Cont
 	}
 
 	templateRef := catalogItem.GetTemplate()
-	if templateRef != "" {
-		cluster.GetSpec().SetTemplate(templateRef)
+	if templateRef == "" {
+		return grpcstatus.Errorf(grpccodes.InvalidArgument, "catalog item '%s' has no template", catalogItemRef)
 	}
+	cluster.GetSpec().SetTemplate(templateRef)
 
 	if err := applyFieldDefinitions(cluster.GetSpec(), catalogItem.GetFieldDefinitions()); err != nil {
 		return err
 	}
 
 	// Look up the template to apply spec defaults, node sets, and parameter validation:
-	templateRef = cluster.GetSpec().GetTemplate()
-	if templateRef == "" {
-		return grpcstatus.Errorf(grpccodes.InvalidArgument, "catalog item has no template")
-	}
 	template, err := s.lookupTemplate(ctx, templateRef)
 	if err != nil {
 		return err
@@ -809,7 +806,6 @@ func (s *PrivateClustersServer) validateAndTransformCatalogItem(ctx context.Cont
 	)
 	cluster.GetSpec().SetTemplateParameters(actualClusterParameters)
 
-	cluster.GetSpec().SetTemplate(template.GetId())
 	for _, clusterNodeSet := range cluster.GetSpec().GetNodeSets() {
 		hostTypeRef := clusterNodeSet.GetHostType()
 		hostType := hostTypes[hostTypeRef]
--- a/internal/servers/private_clusters_server.go
+++ b/internal/servers/private_clusters_server.go
@@ -766,8 +766,54 @@ func (s *PrivateClustersServer) validateAndTransformCatalogItem(ctx context.Cont
 		return err
 	}
 
+	// Look up the template to apply spec defaults, node sets, and parameter validation:
+	templateRef = cluster.GetSpec().GetTemplate()
+	if templateRef == "" {
+		return grpcstatus.Errorf(grpccodes.InvalidArgument, "catalog item has no template")
+	}
+	template, err := s.lookupTemplate(ctx, templateRef)
+	if err != nil {
+		return err
+	}
+	if template.GetMetadata().HasDeletionTimestamp() {
+		return grpcstatus.Errorf(grpccodes.InvalidArgument, "template '%s' has been deleted", templateRef)
+	}
+
+	utils.ApplyClusterSpecDefaults(cluster.GetSpec(), template.GetSpecDefaults())
+
 	if err := utils.ValidateClusterSpecFields(cluster.GetSpec()); err != nil {
 		return err
+	}
+
+	hostTypes, err := s.lookupAndIndexHostTypes(ctx, template)
+	if err != nil {
+		return err
+	}
+
+	templateNodeSets := template.GetNodeSets()
+	clusterNodeSets := cluster.GetSpec().GetNodeSets()
+	if err := s.validateNodeSets(clusterNodeSets, templateNodeSets, hostTypes, templateRef); err != nil {
+		return err
+	}
+
+	mergeNodeSetsWithTemplate(cluster, templateNodeSets, clusterNodeSets)
+
+	clusterParameters := cluster.GetSpec().GetTemplateParameters()
+	if err := utils.ValidateClusterTemplateParameters(template, clusterParameters); err != nil {
+		return err
+	}
+
+	actualClusterParameters := utils.ProcessTemplateParametersWithDefaults(
+		utils.ClusterTemplateAdapter{ClusterTemplate: template},
+		clusterParameters,
+	)
+	cluster.GetSpec().SetTemplateParameters(actualClusterParameters)
+
+	cluster.GetSpec().SetTemplate(template.GetId())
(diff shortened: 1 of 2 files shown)


internal/servers/private_clusters_server.go
```diff
@@ -766,10 +766,56 @@ func (s *PrivateClustersServer) validateAndTransformCatalogItem(ctx context.Cont
 		return err
 	}
 
+	// Look up the template to apply spec defaults, node sets, and parameter validation:
```
sk-ilya, line 770: I think it contradicts the EP
https://github.com/osac-project/enhancement-proposals/blob/main/enhancements/catalog-items/README.md#api-behavior so perhaps, if you want to change it, the EP needs to be updated first

tzvatot: The EP has a gap around how node sets flow from the template to the cluster spec.

 The EP says fields is "the complete contract" and that fields not listed are "neither editable nor pre-defined." The field definition mechanism works well for scalar values like `template_parameters.cpu_count` - the admin defines a path, a default, and whether it's editable.

But node sets are different. A template defines which node sets exist (e.g., control-plane with 3 nodes on host-type-A, workers with 5 nodes on host-type-B). The EP's path examples like `node_sets.workers.size` assume a node set called "workers" already exists in the cluster spec, but nothing in the EP describes how it gets there. Field definitions can set properties on node sets, but they can't create the node sets themselves.

This PR fills that gap by copying node sets from the template into the cluster spec before field definitions are applied. I agree the EP should be updated to describe this. Does that make sense?

sk-ilya: I haven't checked yet, is it similar to how compute instance catalog items work, or VmaaS needs to be updated as well? CC @ygalblum

ygalblum: I don't understand why node sets is handled differently. Can't a `CatalogItem` define them? Can't the user?

In my view the flow should be:
1. Start from the Templates default
2. Apply the Catalog Item defaults
3. Apply the user provided values if editable and passes validation, otherwise, fail the request
This should be the case for all types (`ComputeInstance`, `BareMetalInstance`, `Cluster` or any other type OSAC may have) 

We did something similar in DCM: https://github.com/dcm-project/catalog-manager/blob/cb7127581dc6062c81a8bff3db264fb879dd0655/internal/service/spec_builder.go#L33. Here we used the same structure for all service types so the code is also shared.

tzvatot: I agree the layered approach (template → catalog item defaults → user values) is the right model.

 The reason node sets need separate handling in this PR is a proto design choice. `ClusterTemplate` splits its spec into three separate surfaces:

  ClusterTemplate
  ├── node_sets          → ClusterSpec.node_sets
  ├── spec_defaults      → ClusterSpec.{pull_secret, ssh_public_key, release_image, network}
  └── parameters         → ClusterSpec.template_parameters

https://github.com/osac-project/fulfillment-service/blob/main/proto/private/osac/private/v1/cluster_template_type.proto

`spec_defaults` is a dedicated proto message (`ClusterTemplateSpecDefaults`) that mirrors a subset of `ClusterSpec` scalar fields. `ApplyClusterSpecDefaults` copies those into the cluster spec. But `node_sets` sits directly on `ClusterTemplate`, outside `spec_defaults`, so it doesn't flow through that path and needs its own merge step.

In DCM, the `ServiceType` has one unified `map[string]any` spec - everything including the equivalent of `node_sets` flows through a single deep-copy. OSAC's split into three surfaces is what creates the special-casing.

To unify this like DCM, we'd need to fold `node_sets` into `spec_defaults` (making it the full cluster spec base) or adopt a generic "start from full template spec" approach with map[string]any. Both touch the proto definition and affect all resource types.


I'd suggest landing this as the bug fix for OSAC-1832 (clusters created via catalog items are currently broken without it) and opening a follow-up to unify the spec-building flow. WDYT?

ygalblum: > To unify this like DCM, we'd need to fold `node_sets` into `spec_defaults`

Not sure why this wasn't the implementation from the get go. If we need a quick fix, I guess it's OK. But, we keep accumulating technical debt.

internal/servers/private_clusters_server.go
```diff
@@ -766,10 +766,56 @@ func (s *PrivateClustersServer) validateAndTransformCatalogItem(ctx context.Cont
...
+	templateRef = cluster.GetSpec().GetTemplate()
```
ygalblum: Didn't we already get the `templateRef` from the catalog item? Why again? Even the error message says that the catalog item didn't have it

internal/servers/private_clusters_server.go
```diff
@@ -766,10 +766,56 @@ func (s *PrivateClustersServer) validateAndTransformCatalogItem(ctx context.Cont
...
+		return err
+	}
+
+	actualClusterParameters := utils.ProcessTemplateParametersWithDefaults(
+		utils.ClusterTemplateAdapter{ClusterTemplate: template},
+		clusterParameters,
+	)
+	cluster.GetSpec().SetTemplateParameters(actualClusterParameters)
+
+	cluster.GetSpec().SetTemplate(template.GetId())
```
ygalblum: Why is this needed? 
IIUC we got the `templateRef` from the `cluster`, then got the `template` using the `templateRef` and now we're assigning the `id` of the template back to the `cluster`.

internal/servers/private_clusters_server.go
tzvatot: This is accumulated structure rather than a deliberate design choice.

The original `ClusterTemplate` (PoC, April 2025) had `node_sets` and `parameters` as direct fields with no defaults mechanism at all. 
When I added `spec_defaults` later (MGMT-24213), I followed the existing `ComputeInstanceTemplate` pattern which uses a separate message for scalar defaults. `node_sets` already had its own merge logic that predated `spec_defaults`, so I didn't fold them together.

The gap only surfaced when catalog items introduced field definitions as "the complete contract" - that path
skipped the template's node_sets merge entirely.

I can open a follow-up ticket to unify the spec-building flow so templates carry one unified spec base (folding node_sets into spec_defaults or adopting a generic approach), and merge this one. WDYT?

internal/servers/private_clusters_server.go
tzvatot: Fixed in f931ad18 - moved the template-ref validation to right after extracting it from the catalog item, removing the redundant re-read from `cluster.GetSpec()`. The error message now also includes the catalog item reference for clarity.

internal/servers/private_clusters_server.go
tzvatot: Good catch. I removed it.

ygalblum (approved):
/approve
/lgtm

PR #819, opened 2026-07-01
OSAC-1579: Fix handling of hierarchical projects in Keycloak

jhernand:
## Summary

- Convert dot-separated project names to slash-separated paths before creating Keycloak groups, producing a proper hierarchical group tree instead of a flat set.
- Replace the path-based group lookup (which fails for nested groups) with a new `getGroupIDByName` method that lists children of the parent group and matches by name.
- Remove self-reference and circular dependency checks from the project controller, as these invariants are already enforced by the projects server at creation time.

## Test plan

- [ ] Verify unit tests pass: `ginkgo run -r internal`
- [ ] Verify hierarchical project groups are created correctly in Keycloak with nested paths
- [ ] Verify conflict handling works when a group already exists in a nested hierarchy

Related: https://redhat.atlassian.net/browse/OSAC-1579

## Summary by CodeRabbit

* **New Features**
  * Project and authorization group handling now works with full project paths, including nested paths.
  * Group names are created using slash-based hierarchy paths for more consistent organization.

* **Bug Fixes**
  * Improved handling when a group already exists during creation.
  * Added clearer validation for invalid project paths.
  * Project creation now rejects self-referencing parent relationships.

### DIFF
--- a/internal/controllers/computeinstance/instance_types_client_mock.go
+++ b/internal/controllers/computeinstance/instance_types_client_mock.go
@@ -22,6 +22,7 @@ import (
 type MockInstanceTypesClient struct {
 	ctrl     *gomock.Controller
 	recorder *MockInstanceTypesClientMockRecorder
+	isgomock struct{}
 }
 
 // MockInstanceTypesClientMockRecorder is the mock recorder for MockInstanceTypesClient.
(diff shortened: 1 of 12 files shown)


CrystalChun (approved):
/lgtm

PR #822, opened 2026-07-01
OSAC-1639: resolve tenant deletion failures caused by project FK violations

danielerez:
Tenant deletion was failing with cascading errors because no reconciler was responsible for cleaning up child projects before archiving the tenant. The DAO archive performs a plain DELETE (no CASCADE), so any row in the projects table referencing the tenant caused an FK violation.

Root cause: two reconcilers (tenant and onboarding) share the same "fulfillment-controller" finalizer and both attempted to remove it without first ensuring child projects were archived. Additionally, root projects (created by a DB trigger with an empty name) had several issues that prevented their cleanup.

Changes:

1. Tenant reconciler — cascade delete projects before finalizer removal:
   - Added deleteProjects() that lists all tenant projects, soft-deletes any without a deletion timestamp, then re-queries to return the accurate remaining count (projects without finalizers are archived immediately by the DAO Delete call)
   - delete() now blocks until all projects reach zero before proceeding to IDP cleanup and finalizer removal

2. Onboarding reconciler — guard against premature finalizer removal:
   - Added the same project count check before removeFinalizer() to prevent the onboarding reconciler from racing ahead and triggering the FK-violating archive

3. Project reconciler — fix root project lifecycle:
   - Self-reference fix: added "this.id != <self>" exclusion to the child project query in delete(), preventing root projects (empty name) from matching themselves and blocking their own deletion forever
   - Skip DeleteProjectGroups for root projects (empty name) since they have no Keycloak groups
   - Fast-track root project activation in validateAndActivate() to avoid failed CreateProjectGroups("") calls

4. Keycloak client — idempotent DeleteOrganization:
   - Treat "not found" from GetOrganization as success, so re-reconciliation after partial completion doesn't fail permanently

## Summary by CodeRabbit

* **Bug Fixes**
  * Tenant decommissioning now keeps the tenant finalizer until all associated projects are fully deleted.
  * Project deletion now avoids self-matching when checking for remaining child projects.
  * Keycloak group cleanup errors now prevent finalizer removal, reducing partial teardown.
  * Deleting a root project now signals its owning tenant to reconcile again.
  * Keycloak tenant/org deletion is now idempotent when the organization is already missing.


jhernand (changes requested):
I think that none of our controllers should automatically delete projects when a tenant is deleted, including the default project of the tenant. The administrator should be in charge of that.

To complete the deletion of the tenant when all its projects are deleted what we can do is from the project controller, when a project has been deleted and we remove its finalizer, check if it is the default project of a tenant and signal that tenant so that it will be reconciled again.

internal/controllers/project/project_reconciler_function.go
```diff
@@ -161,6 +161,13 @@ func (t *task) update(ctx context.Context) error {
 func (t *task) validateAndActivate(ctx context.Context) error {
 	t.project.GetStatus().SetState(privatev1.ProjectState_PROJECT_STATE_PENDING)
 
+	// The root project (empty name) is auto-created by a database trigger for each tenant.
+	// It requires no Keycloak groups or parent validation — activate it immediately.
+	if t.project.GetMetadata().GetName() == "" {
+		t.project.GetStatus().SetState(privatev1.ProjectState_PROJECT_STATE_ACTIVE)
+		return nil
+	}
```
jhernand: Why do you think that it doesn't require Keycloak groups? I think it does, like any other project.

internal/controllers/project/project_reconciler_function.go
```diff
@@ -286,10 +293,10 @@ func (t *task) findProjectByName(ctx context.Context, name string) (*privatev1.P
 
 // delete performs the deletion cleanup for a project.
 func (t *task) delete(ctx context.Context) error {
-	// Check for child projects before deletion
+	// Check for child projects before deletion (exclude self to avoid blocking when name is empty)
 	listFilter := fmt.Sprintf(
-		"this.metadata.tenant == %q && this.metadata.project == %q",
-		t.project.GetMetadata().GetTenant(), t.project.GetMetadata().GetName(),
+		"this.metadata.tenant == %q && this.metadata.project == %q && this.id != %q",
```
jhernand: We are trying to make object names unique, and for projects they already are. Can you use here `this.metadata.name` instead of `this.id`?

internal/controllers/project/project_reconciler_function.go
```diff
@@ -315,17 +322,19 @@ func (t *task) delete(ctx context.Context) error {
...
-		t.project.GetMetadata().GetTenant(),
-		t.project.GetMetadata().GetName())
-	if err != nil {
-		t.r.logger.WarnContext(ctx, "Failed to delete Keycloak groups",
-			slog.String("project_id", t.project.GetId()),
-			slog.String("tenant", t.project.GetMetadata().GetTenant()),
-			slog.String("project_name", t.project.GetMetadata().GetName()),
-			slog.Any("error", err),
-		)
+	// Clean up Keycloak groups (root projects have empty names and no Keycloak groups)
```
jhernand: I think default projects should have Keycloak groups, like any other project.

internal/controllers/tenant/tenant_reconciler_function.go
```diff
@@ -275,6 +279,21 @@ func (t *task) isBuiltin() bool {
 
 // delete performs the deletion cleanup for a tenant.
 func (t *task) delete(ctx context.Context) error {
+	// Cascade delete to child projects — soft-delete any that don't have a
+	// deletion timestamp yet, then wait for them to be fully archived by the
+	// project reconciler before proceeding.
```
jhernand: In general you will not be able to do this because there will likely be objects in those projects. I think this should be a task for the user, not for the system.

jhernand (changes requested):

internal/controllers/tenant/tenant_reconciler_function.go
danielerez: Ok, makes sense. Removed the auto deletion of project.

internal/controllers/project/project_reconciler_function.go
danielerez: Oh, just assumed it's not required. Removed from this change.

internal/controllers/project/project_reconciler_function.go
danielerez: Sure, changed.

internal/controllers/project/project_reconciler_function.go
danielerez: done

danielerez:
> I think that none of our controllers should automatically delete projects when a tenant is deleted, including the default project of the tenant. The administrator should be in charge of that.
> 
> To complete the deletion of the tenant when all its projects are deleted what we can do is from the project controller, when a project has been deleted and we remove its finalizer, check if it is the default project of a tenant and signal that tenant so that it will be reconciled again.

 Sure, changed the approach accordingly:
 * Removed auto deletion of project. The tenant can be deleted once the admin delete all associated projects.
 * The project controller now signals the parent tenant after the root project's finalizer is removed.

jhernand (approved):

PR #823, opened 2026-07-01
OSAC-1812: Populate keycloak_user_id in user status

CrystalChun:
# Description

New users controller populates the `User.status.keycloak_user_id` which will be used to identify users in IDP operations.

# Testing

1. Deployed integration environment, logged into CLI as cloud provider admin
2. Created a tenant and an IdP in the tenant
3. Logged in as user from IdP
4. Logged in as cloud provider admin and ran `./osac get users -oyaml` and observed the Keycloak ID get populated for the user
5. Verified the ID matches the ID in the keycloak console

/cc @jhernand 

## Summary of changes

* **New Features**
  * Added a new user reconciliation flow that looks up Keycloak users by username and writes Keycloak user IDs into user status, using update masking to avoid unnecessary updates.
  * Extended identity provider support with username-based user lookup.

* **Bug Fixes**
  * Improved user provisioning to treat concurrent “already exists” cases as success.
  * Startup user services are initialized earlier to improve the boot sequence reliability.

* **Tests**
  * Expanded unit and integration coverage for user provisioning and reconciliation, including edge cases and validation.


jhernand (changes requested):
Why is this needed? Won't the controller see the new user, find the keycloak user id and put it in the status?

My concern is that we are breaking the rule of spec/status: status is only for the controllers, but here we are setting it in something that isn't a controller. If we really need this, and the controller can't handle this, then I'd suggest to move the  `keycloak_user_id` field to the spec of the user type (only in the private API), or have it in both places. But I think it is better if the controller handles this.

CrystalChun:
> Why is this needed? Won't the controller see the new user, find the keycloak user id and put it in the status?
> 
> My concern is that we are breaking the rule of spec/status: status is only for the controllers, but here we are setting it in something that isn't a controller. If we really need this, and the controller can't handle this, then I'd suggest to move the `keycloak_user_id` field to the spec of the user type (only in the private API), or have it in both places. But I think it is better if the controller handles this.

Thank you, Juan! That makes sense that the controllers would set the status. 

I've removed the status setting from the provisioner and created a new Users controller to set this. How does it look now?

jhernand (approved):

jhernand:
> Thank you, Juan! That makes sense that the controllers would set the status.
> 
> I've removed the status setting from the provisioner and created a new Users controller to set this. How does it look now?

Looks good. But I just realized there may be a problem. As we are creating the user directly via the DAO, there will be no event triggered when that happens, because events are generated by the servers. As a result the just created user won't be reconciled immediately. It will only be reconciled later, when something changes. Maybe we should change the provisioner so that it creates the user via the server instead of via the DAO?

CrystalChun:
> Looks good. But I just realized there may be a problem. As we are creating the user directly via the DAO, there will be no event triggered when that happens, because events are generated by the servers. As a result the just created user won't be reconciled immediately. It will only be reconciled later, when something changes. Maybe we should change the provisioner so that it creates the user via the server instead of via the DAO?

You're right, I just ran into this while testing it. ~I modified it to create it via the gRPC user client instead, does it look ok?~

EDIT: sorry going with a different method: using DAO's event notifier instead

CrystalChun:
Ok updated with a DAO notifier, how does it look @jhernand?

jhernand:
> You're right, I just ran into this while testing it. ~I modified it to create it via the gRPC user client instead, does it look ok?~
> 
> EDIT: sorry going with a different method: using DAO's event notifier instead

You would be replicating the logic to generate events. Can we use the server instead? I mean, add something like this to the builder of the provisioner:

```go
func (b *UserProvisionerBuilder) SetUsersServer(value privatev1.UsersServer) *UserProvisionerBuilder {
        ...
}
```

Then you can call the server in a very similar way to how you call the DAO. The server will take care of generating the event. The only drawback of this is that it will require some reordering in the `start grpc-server` command to make sure that the private server is created before the provisioner, but that isn't a big deal, I think.

CrystalChun:
> You would be replicating the logic to generate events. Can we use the server instead? I mean, add something like this to the builder of the provisioner:
> 
> ```go
> func (b *UserProvisionerBuilder) SetUsersServer(value privatev1.UsersServer) *UserProvisionerBuilder {
>         ...
> }
> ```
> 
> Then you can call the server in a very similar way to how you call the DAO. The server will take care of generating the event. The only drawback of this is that it will require some reordering in the `start grpc-server` command to make sure that the private server is created before the provisioner, but that isn't a big deal, I think.

Ah gotcha, ok I've attempted to use just the server and added the recommended function to the provisioner. How does it look now?

jhernand (approved):

PR #834, opened 2026-07-02
OSAC-1284: Fix assigning/unassigning users from projects

CrystalChun:
# Description

Relevant changes:
- Change attribution logic to use user id rather than username
- Use the Keycloak ID for the creator of the project to assign them to the project's `system:managers` group when they created the project.

# Testing

- [x] go build passes
- [x] go unit test passes
- [x] deployed integration environment, created tenant, connected idp
    - created project as idp user and verified user was auto added to `system:managers` group of project in keycloak
    - added another tenant user to project as `system:viewer` and verified they appeared on keycloak as a member of the project's `system:viewers` group 

/cc @jhernand 

## Summary by CodeRabbit

* **New Features**
  * Creator attribution now uses stable user IDs instead of usernames when available.
  * Project activation now automatically adds the project creator to the managers group when their account can be resolved.

* **Bug Fixes**
  * Reconciliation now continues safely if the creator can’t be found or lacks a linked user ID.
  * Attribution falls back to the original username when user lookup or ID resolution fails.


internal/servers/generic_server.go
```diff
@@ -399,28 +399,50 @@ func (s *GenericServer[O]) Get(ctx context.Context, request any, response any) e
...
 		Do(ctx)
 	if err != nil {
 		var notFoundErr *dao.ErrNotFound
 		if errors.As(err, &notFoundErr) {
-			return grpcstatus.Errorf(grpccodes.NotFound, "object with identifier '%s' not found", requestId)
-		}
-		var deniedErr *dao.ErrDenied
-		if errors.As(err, &deniedErr) {
-			return grpcstatus.Errorf(grpccodes.PermissionDenied, "%s", deniedErr.Reason)
+			// If not found by ID, try to find by name
```
jhernand: This doesn't belong into the generic server, in my opinion. If you need this logic it should go into the specific server that needs it. I would argue why is it needed at all. Can you elaborate on what is the reason?

jhernand (changes requested):

internal/servers/generic_server.go
CrystalChun: Gotcha, and I added this because I stored the creator's username in the metadata of the project, and thought it would help with user lookup
https://github.com/osac-project/fulfillment-service/pull/834/changes#diff-68513e93c3b3867a6e54c4f18ba3cb340315f55236232bdacc822e079efc2e68R245-R250

However, maybe I shouldn't use the creator's username, and use their id instead?

CrystalChun: I switched it to use a user's id instead of username in the `metadata.creator` field, what do you think of this?
I also moved some of the unrelated changes to a different PR

jhernand (approved):

jhernand (approved):

PR #835, opened 2026-07-03
NO-ISSUE: Move `NormalizeNil` back to the reflection package

jhernand:
## Summary

- Completes the decoupling of the `reflection` package from `config` by removing the remaining
  dependency in the test files.
- Moves `NormalizeNil` back to the `reflection` package (its natural home) and removes the `util`
  package entirely.
- Updates all callers to import from `reflection` instead of `util`.

## Test plan

- [x] Pre-commit hooks pass (golangci-lint, formatting, etc.)
- [x] `ginkgo run -r internal` passes

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of empty values across several app components, reducing the chance of unexpected failures when optional settings are not provided.
  * Fixed filtering behavior so tenant-based results and user-defined filters are applied more consistently.

* **Refactor**
  * Consolidated shared empty-value handling into a single approach used throughout the app.

* **Tests**
  * Updated and expanded coverage for setup, lookup, CRUD, and filtering behavior.


internal/reflection/reflection_helper_test.go
```diff
@@ -58,28 +53,28 @@ var _ = Describe("Reflection helper", func() {
 
 	Describe("Creation", func() {
 		It("Can be created with all the mandatory parameters", func() {
-			helper, err := NewHelper().
+			globalHelper, err := NewHelper().
```
ygalblum, line 56: Not sure I understand the name change. Plus calling it `global` is a bit confusing because it made me think it's global to the suite or something, but it's local to the test.

internal/reflection/reflection_helper_test.go
ygalblum, line 1: Should the deleted test for `NormalizeNil` be added here?

PR #837, opened 2026-07-05
OSAC-1972: add image defaults, validation, and immutability for BareMetalInstance

mennyaboush:
## Summary

Add server logic for the BareMetalInstance `image` field (proto added in #831):

- **Spec defaults**: `ApplyBareMetalInstanceSpecDefaults()` merges template `spec_defaults.image` into the spec when the user doesn't provide one. Supports partial merge (user provides `source_ref`, template provides `source_type`). Pattern follows `mergeImageDefaults()` from ComputeInstance.
- **Validation**: `ValidateBareMetalInstanceImage()` validates `source_type` and `source_ref` are present when an image is provided. Called from `validateSpec()` during Create.
- **Immutability**: Rejects PATCH requests that change `spec.image` after creation. Follows the same pattern as `ssh_public_key`, `user_data`, `template_parameters`.

Safe to merge independently — the image field is optional, so existing Create/Update calls without an image continue to work.

## Files changed

- `internal/utils/baremetal_spec_defaults.go` (new) — spec defaults and validation
- `internal/servers/private_baremetal_instances_server.go` — wire defaults, validation, immutability
- `internal/servers/private_baremetal_instances_server_test.go` — 6 new tests

## Test coverage

| Test | What it verifies |
|------|-----------------|
| Creates object with image and persists it | Image stored and returned on Get |
| Rejects image with missing source_type | Validation catches incomplete image |
| Rejects image with missing source_ref | Validation catches incomplete image |
| Applies image defaults from template spec_defaults | Template default fills in when user omits image |
| User-provided image overrides template default | User value takes precedence |
| Rejects PATCH that changes image | Immutability enforced |

## Test plan

- [x] `go build ./...` passes
- [x] `ginkgo run internal/servers` — 1292 tests pass (6 new)

## Summary by CodeRabbit

* **Bug Fixes**
  * Added stricter validation for bare metal instance `image` during creation, ensuring required image details are provided.
  * Enforced `image` immutability on update when `spec.image` is targeted, rejecting mismatched changes.
  * Improved template behavior so `spec.image` defaults are applied when omitted, with user-provided values overriding or completing missing fields.
  * Enhanced error messaging for cases where required image fields are missing.

### DIFF
--- a/internal/servers/private_baremetal_instances_server.go
+++ b/internal/servers/private_baremetal_instances_server.go
@@ -18,6 +18,7 @@ import (
 	"errors"
 	"log/slog"
 	"maps"
+	"strings"
 
 	"github.com/prometheus/client_golang/prometheus"
 
@@ -208,6 +209,12 @@ func (s *PrivateBareMetalInstancesServer) validateSpec(bmi *privatev1.BareMetalI
 		}
 	}
 
+	if spec.HasImage() {
+		if err := s.validateBareMetalInstanceImage(spec.GetImage()); err != nil {
+			return err
+		}
+	}
+
 	return nil
 }
 
@@ -280,6 +287,8 @@ func (s *PrivateBareMetalInstancesServer) validateAndApplyTemplateParameters(ctx
 	}
 	template := getResponse.GetObject()
 
+	s.applyBareMetalInstanceSpecDefaults(bmi.GetSpec(), template.GetSpecDefaults())
+
 	if len(template.GetParameters()) == 0 && len(providedParams) == 0 {
 		return nil
 	}
@@ -297,8 +306,8 @@ func (s *PrivateBareMetalInstancesServer) validateAndApplyTemplateParameters(ctx
 	return nil
 }
 
-// validateImmutability ensures catalog_item, ssh_public_key, user_data, and template_parameters
-// cannot be changed after creation.
+// validateImmutability ensures catalog_item, ssh_public_key, user_data, template_parameters,
+// and image cannot be changed after creation.
 func (s *PrivateBareMetalInstancesServer) validateImmutability(ctx context.Context,
 	request *privatev1.BareMetalInstancesUpdateRequest) error {
 	mask := request.GetUpdateMask()
@@ -306,13 +315,14 @@ func (s *PrivateBareMetalInstancesServer) validateImmutability(ctx context.Conte
 	updatingSshKey := updateIncludesField(mask, "spec.ssh_public_key")
 	updatingUserData := updateIncludesField(mask, "spec.user_data")
 	updatingTemplateParams := updateIncludesField(mask, "spec.template_parameters")
+	updatingImage := updateIncludesField(mask, "spec.image")
 
 	bmi := request.GetObject()
 	if bmi == nil {
 		return grpcstatus.Errorf(grpccodes.InvalidArgument, "bare metal instance is mandatory")
 	}
 	newSpec := bmi.GetSpec()
-	if newSpec == nil && (updatingCatalogItem || updatingSshKey || updatingUserData || updatingTemplateParams) {
+	if newSpec == nil && (updatingCatalogItem || updatingSshKey || updatingUserData || updatingTemplateParams || updatingImage) {
 		return grpcstatus.Errorf(grpccodes.InvalidArgument, "bare metal instance spec is mandatory")
 	}
 	id := bmi.GetId()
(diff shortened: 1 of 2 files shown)


internal/servers/private_baremetal_instances_server_test.go
```diff
@@ -398,6 +398,205 @@ var _ = Describe("Private bare metal instances server", func() {
...
+					Spec: privatev1.BareMetalInstanceSpec_builder{
+						CatalogItem: catID,
+					}.Build(),
+				}.Build(),
+			}.Build())
+			Expect(err).ToNot(HaveOccurred())
+			spec := createResponse.GetObject().GetSpec()
+			Expect(spec.GetImage().GetSourceType()).To(Equal("registry"))
+			Expect(spec.GetImage().GetSourceRef()).To(Equal("quay.io/default:latest"))
+		})
```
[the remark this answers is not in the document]
mennyaboush: Addressed in commit 3346d0a

internal/utils/baremetal_spec_defaults.go
jhernand, line 1: This doesn't belong into a `utils` package. Make these functions methods of the server.

jhernand (changes requested):

internal/servers/private_baremetal_instances_server_test.go
```diff
@@ -398,6 +398,289 @@ var _ = Describe("Private bare metal instances server", func() {
...
+						}.Build(),
+					}.Build(),
+				}.Build(),
+				UpdateMask: &fieldmaskpb.FieldMask{
+					Paths: []string{"spec.image"},
+				},
+			}.Build())
+			Expect(err).ToNot(HaveOccurred())
+		})
+
```
adriengentil, line 683: Can you add a test with a default image defined in catalog item, and a default image defined in template. The default value in catalog item must win, and be set in the BareMetalInstance.

adriengentil:
> Safe to merge independently — the image field is optional, so existing Create/Update calls without an image continue to work.

once the main flow is implemented, we'll want to reject the creation of BareMetalInstance when there's no image. Do we have a follow up task for that?

tzumainn:
> > Safe to merge independently — the image field is optional, so existing Create/Update calls without an image continue to work.
> 
> once the main flow is implemented, we'll want to reject the creation of BareMetalInstance when there's no image. Do we have a follow up task for that?

Is this accurate? I've heard use cases where hosts are hooked up to a provisioning network and a provisioning server, and provisioned through that without specifying an image.

internal/servers/private_baremetal_instances_server_test.go
mennyaboush: The test was added

internal/servers/private_baremetal_instances_server.go
```diff
@@ -362,5 +372,52 @@ func (s *PrivateBareMetalInstancesServer) validateImmutability(ctx context.Conte
...
+	}
+	img := spec.GetImage()
+	defImg := defaults.GetImage()
+	if img.GetSourceType() == "" && defImg.GetSourceType() != "" {
+		img.SetSourceType(defImg.GetSourceType())
+	}
+	if img.GetSourceRef() == "" && defImg.GetSourceRef() != "" {
+		img.SetSourceRef(defImg.GetSourceRef())
+	}
+}
```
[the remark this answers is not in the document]
mennyaboush: Adding a safe assertion here would be inconsistent with the rest of the code.

internal/servers/private_baremetal_instances_server.go
```diff
@@ -362,5 +372,52 @@ func (s *PrivateBareMetalInstancesServer) validateImmutability(ctx context.Conte
...
 
+	if updatingImage && !proto.Equal(existingSpec.GetImage(), newSpec.GetImage()) {
+		return grpcstatus.Errorf(grpccodes.InvalidArgument,
+			"cannot change spec.image: image is immutable after creation")
+	}
+
+	return nil
+}
+
+func applyBareMetalInstanceSpecDefaults(spec *privatev1.BareMetalInstanceSpec, defaults *privatev1.BareMetalInstanceTemplateSpecDefaults) {
```
jhernand: This package has dozens of files. Having global functions like this means they will clash if re-defined in other files, and may be re-used without you noticing. It may be unlikely in this particular case, but anyhow I'd suggest you make these methods of the server type.

jhernand (changes requested):

jhernand (approved):

jhernand (approved):

PR #842, opened 2026-07-06
OSAC-1635: report tenant CR creation failure in API object status

danielerez:
When the onboarding controller fails to create a Tenant CR or namespace on a hub cluster, set status.state to FAILED with a descriptive message instead of silently retrying. This makes hub sync failures visible to API consumers.

Note: Integration tests (tenant lifecycle) fail because clearFailed can reset a previously-synced  tenant to PENDING, causing the IDP controller to re-attempt creation and hit a 409 conflict. This is fixed by #843 which should be merged first.

## Summary by CodeRabbit

* **Bug Fixes**
  * Hub synchronization failures no longer interrupt reconciliation; instead, the tenant status is set to **Failed** with a message identifying the failing hub.
  * When synchronization succeeds after a prior failure, the tenant status is automatically cleared—returning to **Pending** (no IDP sync) or **Synced** (IDP already synced).
  * Improved consistency of onboarding tenant reconciliation status handling.

### DIFF
--- a/internal/controllers/onboarding/tenant_reconciler_function.go
+++ b/internal/controllers/onboarding/tenant_reconciler_function.go
@@ -118,15 +118,15 @@ func (r *function) run(ctx context.Context, tenant *privatev1.Tenant) error {
 	} else {
 		err = t.update(ctx)
 	}
-	if err != nil {
-		return err
-	}
 	updateMask := r.maskCalculator.Calculate(oldTenant, tenant)
 	if len(updateMask.GetPaths()) > 0 {
-		_, err = r.tenantsClient.Update(ctx, privatev1.TenantsUpdateRequest_builder{
+		_, updateErr := r.tenantsClient.Update(ctx, privatev1.TenantsUpdateRequest_builder{
 			Object:     tenant,
 			UpdateMask: updateMask,
 		}.Build())
+		if err == nil {
+			err = updateErr
+		}
 	}
 	return err
 }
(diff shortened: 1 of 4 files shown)


internal/controllers/onboarding/tenant_reconciler_function_test.go
```diff
@@ -545,7 +551,64 @@ var _ = Describe("run", func() {
...
+				f := newFunction(mockHubCache, mockHubs, mockTenants)
+				err := f.run(ctx, tenant)
+
+				Expect(err).ToNot(HaveOccurred())
+				Expect(tenant.GetStatus().GetState()).To(
+					Equal(privatev1.TenantState_TENANT_STATE_PENDING),
+				)
+				Expect(tenant.GetStatus().HasMessage()).To(BeFalse())
 			})
 		})
```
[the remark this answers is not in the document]
danielerez: Good catch, done.

internal/controllers/onboarding/tenant_reconciler_function.go
```diff
@@ -150,13 +150,31 @@ func (t *task) update(ctx context.Context) error {
 		}
 
 		if err := t.createOrUpdateOnHub(ctx, hub.GetId(), hubEntry); err != nil {
-			return err
+			t.setFailed(fmt.Sprintf(
+				"Failed to sync tenant to hub %s: %v", hub.GetId(), err,
```
jhernand: I believe this will not be visible to the user in the public API, but nevertheless I'd suggest you consider if putting in the status the error message directly is a good practice. It isn't in general because it may leak sensitive information about the system, or be useless for regular users. Can we instead write this to the log and put in the status a more generic message?

danielerez: Yeah, good point. I'll just move the detailed error log and use a generic message in status.

jhernand (approved):

CrystalChun (approved):
/lgtm

danielerez:
> New changes are detected. LGTM label has been removed.

Rebased and fixed integration tests.

CrystalChun (approved):
/lgtm

PR #844, opened 2026-07-06
OSAC-1701: Add integration tests for Identity Provider lifecycle

obochan-rh:
## Summary

- Add 8 integration tests covering the full IdP CRUD lifecycle: creation + reconciliation to READY, deletion + Keycloak cleanup, RBAC enforcement, tenant isolation, early delete race condition, invalid tenant rejection, concurrent multi-tenant creation, and duplicate name rejection
- Tests verify end-to-end flow from gRPC API through controller reconciliation to Keycloak, using the existing `it/` test infrastructure (Kind cluster + Helm deploy)
- Discovered and worked around a pre-existing public→private proto schema mismatch (`health_check_trigger` field) in the RBAC test

## Jira

https://redhat.atlassian.net/browse/OSAC-1701

## Test plan

- [x] `go vet ./it/` passes
- [x] Unit tests pass (1286 servers specs + 36 IdP controller specs)
- [x] All 8 integration tests pass against Kind cluster with Keycloak
- [x] No regressions in other test suites

### DIFF
--- a/it/it_identity_providers_test.go
+++ b/it/it_identity_providers_test.go
@@ -236,12 +236,20 @@ var _ = Describe("Identity provider lifecycle", func() {
 				}.Build(),
 			}.Build(),
 		}.Build())
-		Expect(err).ToNot(HaveOccurred())
-		DeferCleanup(func() {
-			_, _ = client.Delete(ctx, privatev1.IdentityProvidersDeleteRequest_builder{
-				Id: createResponse.GetObject().GetId(),
-			}.Build())
-		})
+		if err == nil && createResponse != nil {
+			DeferCleanup(func() {
+				_, _ = client.Delete(ctx, privatev1.IdentityProvidersDeleteRequest_builder{
+					Id: createResponse.GetObject().GetId(),
+				}.Build())
+			})
+		} else if err != nil {
+			adminStatus, adminOk := grpcstatus.FromError(err)
+			Expect(adminOk).To(BeTrue())
+			Expect(adminStatus.Code()).ToNot(SatisfyAny(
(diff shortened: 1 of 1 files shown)


it/it_identity_providers_test.go
```diff
@@ -0,0 +1,554 @@
...
+				g.Expect(object.GetStatus().GetMessage()).To(ContainSubstring(expectedAlias))
+			},
+			2*time.Minute,
+			time.Second,
+		).Should(Succeed())
+
+		code, _, err := tool.KeycloakAdminRequest(ctx, http.MethodGet,
+			fmt.Sprintf("/identity-provider/instances/%s", expectedAlias), nil)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(code).To(Equal(http.StatusOK))
```
jhernand, line 119: How can this succeed if the provided URLs don't exist? That means we need to harden the controller that does this job: the identity provider shouldn't be ready unless we have verified that it is working.

obochan-rh: Good observation. The current controller design only registers the IdP configuration in Keycloak (creates the alias with the provided URLs) — it doesn't validate that the OIDC endpoints are actually reachable before marking it READY.

This test validates the current behavior accurately: the controller successfully syncs the config to Keycloak and marks the phase as READY.

I agree this is a hardening opportunity — adding endpoint health checks (e.g., hitting the `.well-known/openid-configuration` endpoint) before transitioning to READY would make the system more robust. Should I file a follow-up Jira task for that?

jhernand: Yes, open a Jira ticket.

jhernand (approved):

PR #860, opened 2026-07-07
NO-ISSUE: Remove Ldap option from IdP proto

CrystalChun:
LDAP is not supported so removing this from the IDP type


jhernand (approved):

CrystalChun:
Was previously approved by Juan.
Merging since the force push only removed/updated unit tests that were stale.

PR #864, opened 2026-07-08
OSAC-2285: add hub-access, bundled-postgres, db-migrate, controller-credentials

omer-vishlitzky:
## Summary

- Add hub-access SA + RoleBinding (binds operator's exported ClusterRole)
- Add hosted-clusters ClusterRole/ClusterRoleBinding
- Add bundled-postgres template (Secret + Certificate)
- Add db-migrate pre-upgrade hook
- Add controller-credentials Secret template

Part of [OSAC-1928](https://redhat.atlassian.net/browse/OSAC-1928) umbrella cleanup. Tested atomically via [osac-installer PR #404](https://github.com/osac-project/osac-installer/pull/404).

## Test plan

- [ ] `helm template` renders correctly
- [ ] e2e passes via installer PR #404

## Summary by CodeRabbit

* **New Features**
  * Added standardized application labels to deployed Kubernetes resources for improved identification and management.
  * Added support for configuring OAuth controller credentials through a dedicated Secret, with options to create the Secret and provide client ID and client secret values.

* **Improvements**
  * Simplified authentication credential configuration for Helm deployments.


charts/service/templates/hooks/db-migrate.yaml
jhernand, line 1: This isn't needed: the service automatically migrates the database during startup.

charts/service/values.yaml
```diff
@@ -183,3 +164,18 @@ idp:
...
+hubAccess:
+  enabled: false
+
+bundledPostgres:
+  enabled: false
+  database:
+    name: service
+    user: service
+
+dbMigrate:
```
jhernand: This isn't needed: the service migrates the database automatically when it starts.

charts/service/values.yaml
```diff
@@ -51,34 +51,15 @@ auth:
...
-  controllerCredentials:
+  # Create a Secret named fulfillment-controller-credentials with OAuth client
+  # credentials for the fulfillment controller. When create is true, the chart
+  # renders the Secret with client-id and client-secret keys. The controller
+  # deployment mounts this Secret via auth.controllerCredentials (a list of
+  # ConfigMap/Secret sources passed from the umbrella chart).
+  controllerCredentialsSecret:
+    create: false
+    clientId: ""
+    clientSecret: ""
```
jhernand: One of the purposes of the previous implementation was to avoid putting in the values of the chart secret values. I think that we should keep that.

charts/service/templates/bundled-postgres.yaml
jhernand, line 1: I very much prefer to keep the database creation external to the fulfillment service. If you still want to have this bundled database you can keep it in osac-installer, or in the umbrella chart.

charts/service/templates/hooks/create-hub.yaml
jhernand, line 1: If I understand correctly this assumes that the fulfillment service runs in the same cluster that is used as a hub. That is conceptually incorrect.

If we wish to make this conceptually incorrect configuration the default, then I'd prefer if we implement it in a less fragile way, without shell scripts. Please open a ticket and assign it to the core team.

charts/service/templates/controller-credentials.yaml
```diff
@@ -0,0 +1,13 @@
...
+kind: Secret
+metadata:
+  name: fulfillment-controller-credentials
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "fulfillment-service.labels" . | nindent 4 }}
+type: Opaque
+stringData:
+  client-id: {{ .Values.auth.controllerCredentialsSecret.clientId | quote }}
+  client-secret: {{ .Values.auth.controllerCredentialsSecret.clientSecret | quote }}
```
jhernand: We introduced the complicated way to provide credentials to avoid this anti-pattern: have secrets in the "values.yaml" file of the chart.

jhernand (changes requested):
I understand the desire to move this out of osac-installer, to simplify it. But I think that these tasks don't belong into the fulfillment service.

charts/service/templates/hub-access.yaml
jhernand, line 1: The job that creates the default hub was removed, but this is still assuming that the cluster where the fulfillment service runs is also a hub. I think this should be removed as well.

charts/service/templates/_helpers.tpl
```diff
@@ -38,3 +38,12 @@ Return the hostname for the fulfillment internal API. Fails if 'internalHostname
...
 {{- end -}}
+
+{{/*
+Common labels
+*/}}
+{{- define "fulfillment-service.labels" -}}
+app.kubernetes.io/part-of: fulfillment-service
+app.kubernetes.io/instance: {{ .Release.Name }}
+app.kubernetes.io/managed-by: {{ .Release.Service }}
+{{- end -}}
```
jhernand: If you remove the `hub-access.yaml` file then this will not be used.

charts/service/values.yaml
```diff
@@ -51,35 +51,6 @@ auth:
...
-  #         items:
-  #         - key: id
-  #           param: client-id
-  #         - key: secret
-  #           param: client-secret
-  #
-  # The keys of the ConfigMap and Secret don't need to match the parameter names. In the above example the client
-  # identifier is in an `id` key but the credential parameter name is `client-id`.
-  controllerCredentials:
-
```
jhernand, line 82: If we remove this how will the controller get its credentials?

jhernand (changes requested):

charts/service/values.yaml
omer-vishlitzky: osac-prereqs in https://github.com/osac-project/osac-installer/pull/404 chart's create-controller-credentials hook, which reads the keycloak client secret and creates fulfillment-controller-credentials in the OSAC namespace.
The old auth.controllerCredentials list format in the deployment template still works because the umbrella chart passes the right volume source references.

jhernand: I don't understand that. Do you mean that osac-installer patches the deployment created by the chart of the fulfillment service to inject the volume containing the secret directly? If not, how does the fulfillment service get that secret or configmap mounted?

If that injection is what you are doing then we need to change it. The contract of the fulfillment service is this chart. If you inject volumes then it may break without notice.

PR #874, opened 2026-07-09
OSAC-2200: Update docs surrounding Keycloak

CrystalChun:
Updates documentation, comments, and log messages to remove the implication that another identity broker can be used since we now only support Keycloak as the identity broker.

Updates `AUTH.md` to reflect the current implementation.

## Summary by CodeRabbit

* **Documentation**
  * Updated the authorization and tenancy guide to be Keycloak-specific.
  * Clarified that users map through a single organization membership and that tenant access comes from the `organization` claim.
  * Added guidance on organization-based setup and single-organization limitations.

* **Bug Fixes**
  * Improved validation messaging to clearly state when a Keycloak client is required.

* **Tests**
  * Adjusted tests to match the updated Keycloak-specific validation wording.

### DIFF
--- a/docs/AUTH.md
+++ b/docs/AUTH.md
@@ -7,8 +7,8 @@ and configure the necessary mappings and authorization rules.
 > The Keycloak setup requires:
 >
 > - A valid OAuth issuer URL
-> - JWT tokens containing a username claim (`preferred_username` or `username`) and optionally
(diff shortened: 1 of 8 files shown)


jhernand (approved):

CrystalChun:
Previously approved, merging.

PR #876, opened 2026-07-10
OSAC-1852: Introduce osac whoami cli cmd

CrystalChun:
# Description

Introduces the `osac whoami` command to display the current logged in user via the CLI. 

Combines the logic for displaying the user's token from `osac get payload --token` with `osac whoami --show-token-claims` 

# Testing 

While logged out, ran `osac whoami`
```sh
$ ./osac whoami
Not logged in. Use 'osac login' to authenticate.
```

Logged in as keycloak user
```sh
$ ./osac whoami
Logged in as: ben
Tenant: osac-demo
Roles: tenant-admin
```

## Summary by CodeRabbit

* **New Features**
  * Added a `whoami` command to show the currently authenticated user.
  * Added flags to optionally show the raw access token and decoded token claims.
  * Added options to render token timestamp claims in RFC 3339 and optionally force UTC.
* **Bug Fixes**
  * Improved consistent error messaging for unauthenticated sessions and token retrieval failures (“Not logged in… use `osac login`”).
* **Refactor**
  * Centralized JWT claim decoding and timestamp rendering for CLI token display.
* **Tests**
  * Added coverage for `whoami` and token claim display utilities.

### DIFF
--- a/internal/cmd/cli/tokenutil/token_display.go
+++ b/internal/cmd/cli/tokenutil/token_display.go
@@ -1,0 +1,100 @@
+/*
+Copyright (c) 2026 Red Hat Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with the
+License. You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific
+language governing permissions and limitations under the License.
+*/
+
+package tokenutil
+
+import (
+	"context"
+	"encoding/json"
+	"log/slog"
+	"time"
+
+	"github.com/golang-jwt/jwt/v5"
+
+	"github.com/osac-project/fulfillment-service/internal/logging"
+	"github.com/osac-project/fulfillment-service/internal/terminal"
+)
+
+// ParseTokenClaims parses a JWT token string and returns the claims without verification.
+// This is useful for displaying token information to the user.
+func ParseTokenClaims(tokenString string) (jwt.MapClaims, error) {
+	parser := jwt.NewParser(jwt.WithoutClaimsValidation())
+	parsedToken, _, err := parser.ParseUnverified(tokenString, jwt.MapClaims{})
+	if err != nil {
+		return nil, err
+	}
+
+	claims, ok := parsedToken.Claims.(jwt.MapClaims)
+	if !ok {
+		return nil, nil
+	}
+
+	return claims, nil
+}
+
+// DisplayTokenClaims displays decoded token claims to the console as JSON.
+// If rfc3339 is true, time claims (iat, exp, nbf, auth_time) are converted to RFC3339 format.
+// If utc is true, time claims are displayed in UTC timezone.
+func DisplayTokenClaims(ctx context.Context, console *terminal.Console, claims jwt.MapClaims, rfc3339, utc bool) {
+	if rfc3339 {
+		logger := logging.LoggerFromContext(ctx)
+		claims = replaceTimeClaims(ctx, logger, claims, utc)
+	}
+	console.RenderJson(ctx, claims)
+}
+
+// replaceTimeClaims converts time claims to RFC3339 format for better readability.
+func replaceTimeClaims(ctx context.Context, logger *slog.Logger, claims jwt.MapClaims, utc bool) jwt.MapClaims {
+	result := jwt.MapClaims{}
+	for name, value := range claims {
+		switch name {
+		case "iat", "exp", "nbf", "auth_time":
+			result[name] = replaceTimeClaim(ctx, logger, name, value, utc)
+		default:
+			result[name] = value
+		}
+	}
+	return result
+}
+
+// replaceTimeClaim converts a single time claim from Unix timestamp to RFC3339 format.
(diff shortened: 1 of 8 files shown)


internal/cmd/cli/whoami/whoami_cmd_test.go
```diff
@@ -0,0 +1,63 @@
...
+*/
+
+package whoami
+
+import (
+	. "github.com/onsi/ginkgo/v2/dsl/core"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("Whoami command", func() {
```
ygalblum: Should't the test also check the actual command? For example, there's a test about supporting the `--show-token` flag. But, nothing verifies that setting it to true actually shows the token.

internal/cmd/cli/tokenutil/token_display.go
```diff
@@ -0,0 +1,127 @@
...
+		return jwt.MapClaims{}, nil
+	}
+
+	return claims, nil
+}
+
+// DisplayTokenAndClaims displays both the raw token and its decoded claims to the console.
+// If rfc3339 is true, time claims (iat, exp, nbf, auth_time) are converted to RFC3339 format.
+// If utc is true, time claims are displayed in UTC timezone.
+func DisplayTokenAndClaims(ctx context.Context, console *terminal.Console, tokenString string, rfc3339, utc bool) error {
```
ygalblum: I don't see any usage of this function. Did I miss it? If not, should we add unused code?

internal/cmd/cli/tokenutil/token_display.go
```diff
@@ -0,0 +1,127 @@
...
+func DisplayTokenClaims(ctx context.Context, console *terminal.Console, claims jwt.MapClaims, rfc3339, utc bool) {
+	if rfc3339 {
+		logger := logging.LoggerFromContext(ctx)
+		claims = ReplaceTimeClaims(ctx, logger, claims, utc)
+	}
+	console.RenderJson(ctx, claims)
+}
+
+// ReplaceTimeClaims converts time claims to RFC3339 format for better readability.
+func ReplaceTimeClaims(ctx context.Context, logger *slog.Logger, claims jwt.MapClaims, utc bool) jwt.MapClaims {
```
ygalblum: AFAICT this function is only used internally. Should it be exported?

CrystalChun: Good catch! I've made it a private function. Thank you!

internal/cmd/cli/tokenutil/token_display.go
CrystalChun: Thanks for catching this! It was used in a previous edit, but is no longer used now. I have removed this function.

internal/cmd/cli/whoami/whoami_cmd.go
```diff
@@ -0,0 +1,217 @@
...
+	tenant := config.TenantFromContext(ctx)
+	if tenant != "" {
+		c.console.Infof(ctx, "Tenant: %s\n", tenant)
+	} else if orgMap, ok := claims["organization"].(map[string]interface{}); ok && len(orgMap) > 0 {
+		// Organization in JWT is a map where keys are org names
+		// Display the first organization as the tenant
+		for orgName := range orgMap {
+			c.console.Infof(ctx, "Tenant: %s\n", orgName)
+			break
+		}
```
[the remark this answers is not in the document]
CrystalChun: I'm not sure how to handle this, technically a user should only belong to one organization but checking for that here seems inappropriate. Should we only display the first one regardless?

internal/cmd/cli/whoami/whoami_cmd_test.go
CrystalChun: Yes, that's a good point. I've added additional tests. How does it look?

internal/cmd/cli/whoami/whoami_cmd_test.go
```diff
@@ -0,0 +1,495 @@
...
+
+	createTestToken := func(username, tenant string, roles []string) string {
+		claims := jwt.MapClaims{
+			"username": username,
+			"iat":      time.Now().Unix(),
+			"exp":      time.Now().Add(1 * time.Hour).Unix(),
+		}
+		if tenant != "" {
+			claims["organization"] = map[string]interface{}{
+				tenant: map[string]interface{}{},
```
ygalblum: Small nit. I know we're using tenant and organization interchangeably. But, maybe it's better to name the parameter that is used to set the `organization`, `organization`.

ygalblum (commented):
small nit.

internal/cmd/cli/whoami/whoami_cmd_test.go
CrystalChun: Gotcha, modified it! Thanks for your review Ygal!! 🙏

internal/cmd/cli/whoami/whoami_cmd.go
```diff
@@ -0,0 +1,217 @@
...
+		c.console.Errorf(ctx, "Not logged in. Use 'osac login' to authenticate.\n")
+		return exit.Error(1)
+	}
+
+	// Get the current token
+	token, err := tokenSource.Token(ctx)
+	if err != nil {
+		c.console.Errorf(ctx, "Not logged in. Use 'osac login' to authenticate.\n")
+		return exit.Error(1)
+	}
```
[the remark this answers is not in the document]
CrystalChun: If token source is empty, would it still be valid to say that the user should log in?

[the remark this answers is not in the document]
CrystalChun: added

ygalblum (approved):
/approve
/lgtm

PR #878, opened 2026-07-10
OSAC-2187: Document that published OpenAPI docs track main only

minmzzhang:
## Summary

Document that the Pages-hosted OpenAPI docs track `main` only and are not release-versioned.

- Add `docs/OPENAPI.md` explaining the publishing model, rationale for excluding tag triggers, and how consumers can generate docs for a specific version locally.
- Add a brief workflow comment pointing to the docs file.

## Test plan

- [ ] Confirm workflow YAML still parses correctly
- [ ] Confirm the job runs correctly on a `main` push (no functional change to the trigger)
- [ ] Verify `docs/OPENAPI.md` renders correctly on GitHub

Fixes [OSAC-2187](https://redhat.atlassian.net/browse/OSAC-2187).

## Summary by CodeRabbit

* **Documentation**
  * Clarified that hosted OpenAPI documentation reflects the latest successful deployment from `main` rather than release-specific versions.
  * Documented why tag-triggered publishing is excluded.
  * Added instructions for generating version-specific OpenAPI documentation locally from a release tag.

### DIFF
--- a/.github/workflows/publish-openapi.yaml
+++ b/.github/workflows/publish-openapi.yaml
@@ -12,6 +12,8 @@
 #
 
 # This workflow generates the OpenAPI specification and uploads it to the GitHub pages of the project.
+# The published docs reflect the latest successfully deployed output from 'main' and are NOT release-versioned.
+# See docs/OPENAPI.md for details on why tag-triggered publishing is intentionally excluded.
 
 name: Publish OpenAPI
 
--- a/docs/OPENAPI.md
+++ b/docs/OPENAPI.md
@@ -1,0 +1,35 @@
+# OpenAPI Documentation Publishing
+
+The hosted OpenAPI docs (GitHub Pages) reflect the latest successfully
+deployed output generated from the tip of `main`. They are **not**
+release-versioned — each deployment replaces the entire site artifact, so no
+per-version snapshot is retained across runs.
+
+## Why no tag trigger?
+
+Unlike sibling workflows (`publish-charts`, `publish-proto`, `publish-image`),
+`publish-openapi.yaml` intentionally does **not** trigger on `v*` tag pushes:
+
+1. **Redundant in the common case** — releases are cut from the tip of `main`,
+   so by the time a tag is pushed the preceding `main` push has already
+   published that exact content.
+
+2. **Harmful in edge cases** — if a tag is ever cut from a commit that is not
+   the current tip of `main` (release lag, hotfix), a tag-triggered run would
+   republish older docs and overwrite newer content already live on the site
+   until the next `main` push corrects it.
+
+3. **Versioned publishing is not feasible** with the current GitHub Pages
+   Actions-based deployment model, which replaces the whole site artifact on
+   every run with no built-in mechanism to retain prior versions.
+
+## Implications for consumers
+
+Do not assume the hosted OpenAPI spec matches any specific tagged release.
+To inspect the API surface for a particular version, check out that tag
+locally and run:
+
+```bash
+go build ./cmd/osac-dev
+./osac-dev generate openapi --project-dir . --output-dir pages/openapi
+```


.github/workflows/publish-openapi.yaml
```diff
@@ -12,13 +12,22 @@
 #
 
 # This workflow generates the OpenAPI specification and uploads it to the GitHub pages of the project.
+#
+# NOTE: the published site is intentionally NOT per-release versioned. It always reflects whatever was most
+# recently generated from 'main' (or from a 'v*' release tag, which triggers a refresh at the exact commit that
+# was released). GitHub Pages deployments replace the entire site artifact on every run, so there is no
+# per-version snapshot retained across deployments -- do not assume the hosted docs match any specific past
+# release once a newer 'main' commit or tag has been published. The 'v*' tag trigger below only guarantees the
+# docs are refreshed at release time; it does not pin or preserve that release's docs going forward.
```
sk-ilya: So the docs are updated on every push to `main`, then regenerated at release time from a commit that may be older than the current tip of `main`, and updated again on the next push to `main`? Assuming releases are created from `main` anyway, I'm struggling to understand why we need to regenerate them on tag creation as well

minmzzhang: You're right, and I agree. I dropped the tag trigger in the latest commit.

You called it exactly: since this workflow doesn't produce a versioned artifact (it always overwrites the whole Pages site), a tag-triggered run adds no value when releases are cut from the tip of `main` (it's just a redundant re-publish of content the preceding `main` push already published). Worse, if a tag is ever cut from a commit that isn't the current tip of `main`, the tag-triggered run would republish *older* docs and overwrite newer content already live on the site until the next `main` push corrects it: a regression, not a safety net.

The ticket (OSAC-2187) actually offers two coherent options: (a) tag trigger *bundled with real versioned publishing*, or (b) documented main-only behavior. Adding the tag trigger alone was neither. I looked into (a) properly (a versioned-path + history-preserving publish step) but GitHub Pages' Actions-based deployment replaces the whole site artifact on every run with no built-in way to retain prior versions, so I went with (b): main-only trigger, plus a comment explaining why tags are intentionally not a trigger here.

sk-ilya: So we don't need this change? Only a comment is left...

minmzzhang: We do need this change, as it was asked to add comments in the original jira...

minmzzhang: Additionally, added a second commit that puts a small, visible note directly on the Swagger UI landing page itself ("tracks main, not version-pinned", with a tooltip for the full explanation). That's the part that actually reaches the people who'd otherwise wrongly assume version alignment -- the workflow comment alone wasn't going to do that.

internal/cmd/osac-dev/generate/openapi/swagger_ui.html
```diff
@@ -38,6 +43,7 @@
 <body>
   <div class="api-selector">
     <h1>OSAC Fulfillment Service</h1>
+    <span class="version-note" title="This page always reflects the latest 'main' at the time it was published. It is not pinned to any specific release version.">tracks main, not version-pinned</span>
```
sk-ilya: The static html itself has nothing to do with where it's being published... it might be served from an osac deployment as well, for example. Let's not couple the two

.github/workflows/publish-openapi.yaml
sk-ilya: @eliorerz I saw you reported that ticket - do we really need this?

eliorerz: I think we probably still do. From what I've checked and the underlying thing hasn't changed on main. there's still no tag trigger and nothing telling anyone the docs aren't version-pinned, so the original concern seems to still exist. 
That said, I'm not 100% sure the docs-only fix is the most valuable way to spend effort here versus just leaving it as known behavior it was mostly meant to avoid someone assuming the hosted spec matches a specific release.
WDYT?

sk-ilya: IMO it's better to not make these changes. Reg. "push on tag", see https://github.com/osac-project/fulfillment-service/pull/878#discussion_r3566127748 this is what @minmzzhang originally implemented

minmzzhang:
@sk-ilya and @eliorerz , I think this is still worth merging. The change is small (a workflow comment + `docs/OPENAPI.md`) but it closes [OSAC-2187](https://redhat.atlassian.net/browse/OSAC-2187) properly:

- The ticket asked us to either add versioned publishing or document why we don't. We investigated versioned publishing and it's not feasible with Pages' replace-whole-site model, so documenting main-only is the right resolution.
- Without this, there's nothing telling anyone the hosted docs aren't version-pinned, the concern Elior originally raised still applies.
- The docs file also gives consumers a self-serve path to generate docs for a specific release locally, which doesn't exist anywhere today.
- It's a few comment lines, zero runtime risk, zero maintenance burden, but saves the next person from re-investigating the same question.

Happy to hear if you'd rather close the ticket as "won't fix" instead, but IMO documenting an intentional design choice is worthwhile.

eliorerz (approved):

PR #866, opened 2026-07-08
OSAC-1222: remove cores/memory_gib from ComputeInstance in favor of instance_type

ygalblum:
Remove the deprecated cores and memory_gib fields from ComputeInstance and ComputeInstanceTemplate protos, replacing them with reserved directives. Update all server logic, CLI flags, reconciler, spec defaults, and tests to use instance_type as the sole compute specification method.

Assisted-by: Claude Code <noreply@anthropic.com>

## Summary by CodeRabbit

* **New Features**
  * Compute instances and templates now use **instance type** as the primary way to choose resources.

* **Bug Fixes**
  * Removed outdated CPU and memory inputs from creation and update flows.
  * Validation now consistently checks instance type availability and state.
  * Default values and required-field checks were updated to match the new instance-type-based behavior.
  * API responses and warnings no longer reference legacy CPU/memory settings.

PR #882, opened 2026-07-12
OSAC-2340: Add defaults field to NetworkClass proto

danmanor:
## [OSAC-2340](https://redhat.atlassian.net/browse/OSAC-2340): Add defaults field to NetworkClass proto

**Jira:** https://redhat.atlassian.net/browse/OSAC-2340
**Epic:** [OSAC-2339](https://redhat.atlassian.net/browse/OSAC-2339) — Default Networking

### Summary

Adds a `NetworkDefaults` message and `spec.defaults` field to NetworkClass in both public and private protos. This is the first task in the Default Networking epic — it defines the contract that downstream tasks (tenant onboarding default resource creation, installer configuration) depend on. Admins configure default CIDRs and security group rules on the NetworkClass; the system uses these to auto-create default networking resources at tenant onboarding.

### Changes

**Proto definitions:**
- Added `NetworkClassSpec` wrapper with `NetworkDefaults defaults` field, following the standard `{id, metadata, spec, status}` object pattern
- `NetworkDefaults` has per-address-family CIDR fields matching VirtualNetwork/Subnet conventions: `virtual_network_ipv4_cidr`, `virtual_network_ipv6_cidr`, `subnet_ipv4_cidr`, `subnet_ipv6_cidr`, plus `repeated SecurityRule ingress_rules/egress_rules`
- Added `protovalidate` CEL annotations: `isIpPrefix(4, true)` / `isIpPrefix(6, true)` for CIDR validation, and per-family subnet-requires-VN message-level constraints
- `spec` field on public API is `OUTPUT_ONLY` — tenants can see defaults but only admins can set them via the private API

**Server (private):**
- `validateDefaultCIDRPair()` — shared helper validating CIDR format, subnet containment within VN CIDR, and the subnet-requires-VN constraint for each address family
- `applyNetworkClassUpdate` — handles `"spec"` and `"spec.defaults"` field mask paths

**Server (public):**
- `spec` field added to `AddIgnoredFields` on `inMapper`

**Bug fix (unrelated):**
- Fixed compilation error in `catalog_item_validation_test.go` caused by `Cores` field removal in PR #866 without updating the test added by PR #783

**No database migration needed** — the JSONB `data` column stores new proto fields automatically.

### Testing
- **Unit tests:** 20 new tests covering CRUD with defaults, all validation error paths (invalid CIDR, subnet containment, invalid rules, subnet without VN), field mask update, update-path validation, OUTPUT_ONLY behavior via public API
- **Integration tests:** N/A — no new component interactions
- **Coverage:** All behavioral paths through public interfaces covered

### Acceptance Criteria

- [x] AC-1: NetworkClass proto includes `spec.defaults` with per-family CIDRs and security group rules
- [x] AC-2: NetworkClass can be created and updated with defaults configuration via API
- [x] AC-3: The `spec.defaults` field is persisted and retrievable via Get/List operations
- [x] AC-4: Server-side validation enforces CIDR format (via protovalidate CEL + Go), subnet containment, subnet-requires-VN, and rule validity
- [x] AC-5: Field mask support for updating defaults via `spec.defaults`

## Summary by CodeRabbit

* **New Features**
  * Added network class defaults for tenant onboarding, including IPv4/IPv6 virtual networks and subnets, security rules, and optional NAT gateway creation.
  * Defaults are available as output-only information through the public API.
* **Bug Fixes**
  * Improved validation for CIDRs, subnet relationships, and security rules.
  * Prevented public requests from modifying protected network class defaults.
  * Added support for replacing or clearing defaults through private API updates.


internal/servers/private_network_classes_server.go
```diff
@@ -384,6 +392,53 @@ func cloneNetworkClass(nc *privatev1.NetworkClass) *privatev1.NetworkClass {
...
+
+	var vnPrefix netip.Prefix
+	if vnCIDR != "" {
+		parsed, err := parseAndValidateCIDR(vnCIDR, cidrIPv4)
+		if err != nil {
+			return grpcstatus.Errorf(grpccodes.InvalidArgument,
+				"defaults.virtual_network_cidr: %v", err)
+		}
+		vnPrefix, _ = netip.ParsePrefix(parsed)
+		_ = vnPrefix
```
ori-amizur: I think this is not needed.

internal/servers/private_network_classes_server.go
```diff
@@ -384,6 +392,52 @@ func cloneNetworkClass(nc *privatev1.NetworkClass) *privatev1.NetworkClass {
...
+	if subnetCIDR != "" {
+		_, err := parseAndValidateCIDR(subnetCIDR, cidrIPv4)
+		if err != nil {
+			return grpcstatus.Errorf(grpccodes.InvalidArgument,
+				"defaults.subnet_cidr: %v", err)
+		}
+
+		if vnCIDR != "" {
+			subnetPrefix, _ := netip.ParsePrefix(subnetCIDR)
+			if !vnPrefix.Contains(subnetPrefix.Addr()) || subnetPrefix.Bits() < vnPrefix.Bits() {
```
ybettan: Nit: the discarded error (`_`) on `netip.ParsePrefix` is safe here since `parseAndValidateCIDR` already validated the string above, but it's fragile if someone refactors this later without noticing the precondition. Consider either assigning and checking, or adding a short comment noting that validation already happened.

Also, the condition `!Contains(addr) || Bits() < vnBits` is correct but reads a bit inside-out — the `< vnPrefix.Bits()` part means "subnet is wider than the VN". Flipping to `Bits() >= vnPrefix.Bits()` with adjusted logic would make the intent slightly more self-documenting ("subnet must be equal or narrower than VN").

internal/servers/private_network_classes_server.go
```diff
@@ -384,6 +392,52 @@ func cloneNetworkClass(nc *privatev1.NetworkClass) *privatev1.NetworkClass {
...
+	}
+
+	if subnetCIDR != "" {
+		_, err := parseAndValidateCIDR(subnetCIDR, cidrIPv4)
+		if err != nil {
+			return grpcstatus.Errorf(grpccodes.InvalidArgument,
+				"defaults.subnet_cidr: %v", err)
+		}
+
+		if vnCIDR != "" {
```
ybettan: If `subnet_cidr` is set without `virtual_network_cidr`, the containment check is skipped and the defaults are persisted with a subnet CIDR but no parent VN CIDR. When the tenant-onboarding code later tries to use these defaults, it would have a subnet to create but no VirtualNetwork to put it in.

Should this be rejected here (i.e., if `subnet_cidr` is set, require `virtual_network_cidr` too)? Or is the onboarding code expected to handle partial defaults gracefully?

danmanor: Good catch — you're right, this is a real gap. The default-networking EP treats defaults as a bundle: tenant onboarding (OSAC-2341) creates all three resources (VN + Subnet + SG) together, and Subnet is a child of VirtualNetwork in the resource hierarchy — it can't exist without a parent.

I'll add validation to reject `subnet_cidr` when `virtual_network_cidr` is empty.

internal/servers/private_network_classes_server.go
danmanor: Fair point on both. I'll handle the error instead of discarding it — will return `grpccodes.Internal` since a re-parse failure after validation would indicate an internal inconsistency rather than bad input. And I'll flip the containment condition to make it read more naturally.

proto/public/osac/public/v1/network_class_type.proto
```diff
@@ -57,6 +58,35 @@ message NetworkClass {
   optional bool is_default = 9 [
     (google.api.field_behavior) = OUTPUT_ONLY
   ];
+
+  // Default networking configuration for tenant onboarding. When a tenant is created, the system uses these
+  // defaults to auto-create a VirtualNetwork, Subnet, and SecurityGroup for the tenant.
+  NetworkDefaults defaults = 12 [
+    (google.api.field_behavior) = OUTPUT_ONLY
+  ];
```
jhernand, line 66: This looks like a candidate for a `spec.defaults` field.

proto/public/osac/public/v1/network_class_type.proto
```diff
@@ -57,6 +58,35 @@ message NetworkClass {
...
+// Default networking configuration used during tenant onboarding.
+//
+// When a tenant is created, the system reads these defaults from the NetworkClass and auto-creates
+// a default VirtualNetwork (using virtual_network_cidr), Subnet (using subnet_cidr), and SecurityGroup
+// (using ingress_rules and egress_rules) for the tenant. All fields are optional — defaults can be
+// partially populated (e.g., CIDRs only without security rules).
+message NetworkDefaults {
+  // Default CIDR for auto-created tenant VirtualNetwork. Must be valid IPv4 CIDR notation.
+  // Example: "10.0.0.0/16"
+  string virtual_network_cidr = 1;
```
jhernand: Consider using `protovalidate` to validate the format, as explained here: https://github.com/osac-project/fulfillment-service/blob/main/docs/API.md#validation-constraints.

proto/public/osac/public/v1/network_class_type.proto
```diff
@@ -57,6 +58,35 @@ message NetworkClass {
...
+
+  // Default CIDR for auto-created tenant Subnet. Must be valid IPv4 CIDR notation and fall within the
+  // virtual_network_cidr range.
+  // Example: "10.0.1.0/24"
+  string subnet_cidr = 2;
```
jhernand: Consider using `protovalidate`.

jhernand (changes requested):

proto/public/osac/public/v1/network_class_type.proto
danmanor: Done — moved `defaults` under a new `NetworkClassSpec` wrapper so it's now `spec.defaults`, following the standard `{id, metadata, spec, status}` pattern.

proto/public/osac/public/v1/network_class_type.proto
danmanor: Done — added `buf.validate` CIDR format regex on both `virtual_network_cidr` and `subnet_cidr`, plus a message-level CEL constraint enforcing that `subnet_cidr` requires `virtual_network_cidr` to be set. Deeper semantic validation (valid octets, prefix lengths, subnet containment) stays in the Go server since CEL can't do CIDR arithmetic.

proto/public/osac/public/v1/network_class_type.proto
danmanor: Addressed in the same commit — see reply above.

ybettan:
/lgtm
Approve label can be added once @ori-amizur and @jhernand are happy about the PR as well.

proto/public/osac/public/v1/network_class_type.proto
```diff
@@ -57,6 +59,51 @@ message NetworkClass {
...
+  option (buf.validate.message).cel = {
+    id: "subnet_requires_virtual_network"
+    message: "subnet_cidr requires virtual_network_cidr to be set"
+    expression: "this.subnet_cidr == '' || this.virtual_network_cidr != ''"
+  };
+
+  // Default CIDR for auto-created tenant VirtualNetwork. Must be valid IPv4 CIDR notation.
+  // Example: "10.0.0.0/16"
+  string virtual_network_cidr = 1 [(buf.validate.field).string = {
+    pattern: "^([0-9]{1,3}\\.){3}[0-9]{1,3}/[0-9]{1,2}$"
```
jhernand: These regular expressions would accept `999.999.999.999/99`. Is that correct? What about IPv6?

danmanor: Good catch — replaced the loose regex with `isIpPrefix(4, true)` CEL expressions, which properly validate IPv4 CIDR notation (rejects `999.999.999.999/99`, validates octet ranges and prefix length).

Regarding IPv6: the default-networking EP only defines IPv4 defaults — no dual-stack or IPv6 support for tenant onboarding defaults. The Go validation also explicitly uses `cidrIPv4`. IPv6 support for defaults would be a separate enhancement.

danmanor: Good catch on both points. Fixed:

1. **Validation**: Replaced the loose regex with `isIpPrefix(4, true)` / `isIpPrefix(6, true)` CEL expressions — these properly validate CIDR notation (rejects `999.999.999.999/99`).

2. **IPv6**: Split the fields per address family to match the existing VirtualNetwork/Subnet proto patterns:
   - `virtual_network_ipv4_cidr` + `virtual_network_ipv6_cidr`
   - `subnet_ipv4_cidr` + `subnet_ipv6_cidr`

   Each family has its own subnet-requires-VN CEL constraint. The Go validation also handles both families via a shared `validateDefaultCIDRPair` helper. The default-networking EP is being updated to reflect this.

jhernand (approved):

PR #886, opened 2026-07-12
NO-ISSUE: Handle 409 conflict idempotently in IdP tenant creation

danielerez:
The fulfillment-controller reconciles tenants using multiple goroutines. When both goroutines reconcile the same tenant simultaneously, they race to create the Keycloak organization:

  Goroutine A: CreateTenant("shared") → 201 Created → "Tenant created in IdP"
  Goroutine B: CreateTenant("shared") → 409 Conflict → TENANT_STATE_FAILED

Goroutine B's FAILED status overwrites goroutine A's SYNCED status (last-write-wins), and every subsequent reconcile hits 409 forever.

On a clean dev cluster this manifests as the "shared" tenant stuck in FAILED immediately after deployment:
```
  $ osac get tenants
  PROJECT  DELETING  ID      NAME    STATE
  -        -         shared  shared  FAILED
  -        -         system  system  SYNCED

  $ osac get tenant shared -o yaml
  status:
    message: 'Tenant creation in IDP failed: failed to create tenant in
      IdP: organization "shared" already exists: API error: status=409
      body={"errorMessage":"A organization with the same name already
      exists."}'
    state: TENANT_STATE_FAILED
```
Fix: treat 409 as idempotent success at the client level — when the resource already exists, look it up and return it instead of erroring. This follows the same pattern already used in groups.go for group creation. Applied to three methods:

- CreateTenant: on 409, fetch and return the existing organization
- CreateUserInRealm: on 409, look up the existing user by username
- AddUserToOrganization: on 409 (already a member), return nil

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when tenant, user, or organization membership records already exist.
  * Duplicate creation requests now reuse existing records instead of failing on conflict responses.
  * User creation conflicts now correctly fall back to existing lookups, with clearer errors when lookups fail.
  * Concurrent setup flows are better handled, reducing failures during retries or overlapping requests.
* **Tests**
  * Added coverage for conflict handling across tenant, user, and membership flows, including an end-to-end race simulation.

### DIFF
--- a/internal/idp/client.go
+++ b/internal/idp/client.go
@@ -142,15 +142,19 @@ func (b *ClientBuilder) Build() (result *Client, err error) {
 	return
 }
 
+func isConflictError(err error) bool {
+	var apiErr *apiclient.APIError
+	return errors.As(err, &apiErr) && apiErr.StatusCode == http.StatusConflict
+}
+
 // CreateTenant creates a new tenant (Keycloak organization in the configured realm).
 // Returns the created tenant with server-assigned ID and any server defaults.
 func (c *Client) CreateTenant(ctx context.Context, tenant *Tenant) (*Tenant, error) {
 	kcOrg := toKeycloakOrganization(tenant)
 	response, err := c.httpClient.DoRequest(ctx, http.MethodPost, fmt.Sprintf("/admin/realms/%s/organizations", c.realmName), kcOrg)
 	if err != nil {
-		var apiErr *apiclient.APIError
-		if errors.As(err, &apiErr) && apiErr.StatusCode == http.StatusConflict {
-			return nil, fmt.Errorf("organization %q already exists: %w", tenant.Name, err)
+		if isConflictError(err) {
+			return c.GetTenant(ctx, tenant.Name)
 		}
 		return nil, fmt.Errorf("failed to create organization: %w", err)
 	}
@@ -236,6 +240,9 @@ func (c *Client) AddUserToOrganization(ctx context.Context, tenantName string, u
 	path := fmt.Sprintf("/admin/realms/%s/organizations/%s/members", c.realmName, org.ID)
 	response, err := c.httpClient.DoRequest(ctx, http.MethodPost, path, userID)
 	if err != nil {
+		if isConflictError(err) {
+			return nil
+		}
 		return fmt.Errorf("failed to add user to organization: %w", err)
 	}
 	defer response.Body.Close()
@@ -246,6 +253,13 @@ func (c *Client) CreateUserInRealm(ctx context.Context, user *User) (*User, erro
 	kcUser := toKeycloakUser(user)
 	response, err := c.httpClient.DoRequest(ctx, http.MethodPost, fmt.Sprintf("/admin/realms/%s/users", c.realmName), kcUser)
 	if err != nil {
+		if isConflictError(err) {
+			existing, lookupErr := c.getUserByUsername(ctx, user.Username)
+			if lookupErr != nil || existing == nil {
+				return nil, fmt.Errorf("user %q already exists but failed to look up: %w", user.Username, errors.Join(err, lookupErr))
+			}
+			return existing, nil
+		}
 		return nil, fmt.Errorf("failed to create user: %w", err)
 	}
 	defer response.Body.Close()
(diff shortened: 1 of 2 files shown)


jhernand (approved):

internal/idp/client.go
```diff
@@ -142,15 +142,19 @@ func (b *ClientBuilder) Build() (result *Client, err error) {
...
+func isConflictError(err error) bool {
+	var apiErr *apiclient.APIError
+	return errors.As(err, &apiErr) && apiErr.StatusCode == http.StatusConflict
+}
+
 // CreateTenant creates a new tenant (Keycloak organization in the configured realm).
 // Returns the created tenant with server-assigned ID and any server defaults.
 func (c *Client) CreateTenant(ctx context.Context, tenant *Tenant) (*Tenant, error) {
 	kcOrg := toKeycloakOrganization(tenant)
 	response, err := c.httpClient.DoRequest(ctx, http.MethodPost, fmt.Sprintf("/admin/realms/%s/organizations", c.realmName), kcOrg)
```
ygalblum, line 154: Not sure I agree with this solution. This implementation assumes that the payload is the same. But, what happens if it isn't? For example, a different list of domains?

internal/idp/client.go
```diff
@@ -246,6 +253,13 @@ func (c *Client) CreateUserInRealm(ctx context.Context, user *User) (*User, erro
 	kcUser := toKeycloakUser(user)
 	response, err := c.httpClient.DoRequest(ctx, http.MethodPost, fmt.Sprintf("/admin/realms/%s/users", c.realmName), kcUser)
 	if err != nil {
+		if isConflictError(err) {
```
ygalblum, line 256: Similarly, what happens if the user details are not the same?

ygalblum (commented):
To implement idempotency correctly, you need to also check the payload. Also, usually a separate command type is used (e.g. `apply` vs `create`).

ygalblum:
You wrote:
> When both goroutines reconcile the same tenant simultaneously, they race to create the Keycloak organization

I think we need to understand why this is happening and can it be avoided

internal/idp/client.go
CrystalChun: That's a good point, should we do a create or update solution here? 

Perhaps compare the one from Keycloak and the one that's being reconciled and if they differ, we should perform an update.

ygalblum: I think first we need to understand why this is happening. Is this a case of multiple user requests or an issue with the internal implementation. 
If it's the former, I think we need to impose order at the entrance, otherwise we'll have a race condition (make sure the first request gets its response and only then process the second request).
If it's the latter, then we need to understand why this is happening and avoid it

danielerez: This is actually an internal reconciler issue, not multiple user requests. The Reconciler framework runs two goroutines (syncLoop and watchLoop) that independently feed the same objectChannel.
both can push the same tenant (as stale PENDING snapshots) before the first finishes processing.
the proper framework fix (re-reading from DB before each reconciliation) affects all reconcilers and needs separate design... I'll open a follow-up for that.

danielerez: About payloads, In this race the payloads are always identical, both copies are snapshots of the same tenant at the same lifecycle point (PENDING). Domain changes are handled separately by updateIDP() which runs on subsequent reconciliations after the tenant reaches SYNCED. 
iinm, Keycloak has no native upsert for organizations, so I guess create-or-update would add complexity without a real divergence scenario here.

internal/idp/client.go
danielerez: so the 409 means a user with this username already exists in Keycloak. Since this is the same reconciler racing against itself on the same tenant, it's always creating the same user with the same details. 
note that currently this method is only used for the break-glass account, but even in general the race is always same-tenant (so same user)

danielerez:
Sorry the PR was merged before I got to address comments, was auto-merged after approval...
Anyway, the 409 is caused by the reconciler's syncLoop and watchLoop both pushing stale PENDING snapshots of the same tenant before the first is processed. so payloads are always identical in this race, domain changes go through a separate updateIDP() path, and break-glass user details should be deterministic. I'll open a ticket for the proper framework fix (re-read from DB before reconciliation).

PR #887, opened 2026-07-12
OSAC-2396: restructure StorageTier proto to use spec/status pattern

rgolangh:
## Summary
- Restructure `StorageTier` proto to use `StorageTierSpec` (description, backends) and `StorageTierStatus` (state, message) sub-messages, aligning with the standard OSAC object structure per `docs/API.md`
- Update server implementation, CLI commands, rendering table definitions, and database triggers for the new field paths
- Add migration 77 to backfill existing JSONB data from flat fields to spec/status structure and update trigger functions to read from `spec.backends`

## Jira
https://redhat.atlassian.net/browse/OSAC-2396

## Test plan
- [x] `buf lint` passes
- [x] `buf generate` produces no diff
- [x] `gofmt -s -w .` produces no diff
- [x] `go build ./...` succeeds
- [x] All 32 StorageTier server tests pass (`ginkgo run --focus="storage tier" -r internal/servers`)
- [x] All 266 migration tests pass (`ginkgo run -r internal/database/migrations`)
- [x] Full unit test suite passes (`ginkgo run -r internal` — 80 suites, 0 failures)

## Release Notes

* **New Features**
  * Storage tiers now use a split model: **Spec** (description, backends) and **Status** (state, optional message).

* **Improvements**
  * Updated CLI, server behavior, and table rendering to read/write nested **Spec**/**Status** fields.
  * Create/update validation now requires non-empty **spec.backends** and enforces nested update masks.
  * Added a migration to reshape existing tier data into nested **spec**/**status** and maintain backend associations.

* **Documentation**
  * Corrected filtering/sorting examples to use **spec.description** and **status.state**.

* **Tests**
  * Expanded migration and server tests for the new nested layout and validation cases.


proto/private/osac/private/v1/storage_tier_type.proto
```diff
@@ -75,12 +75,27 @@ message StorageTier {
...
   // List of backend associations that compose this storage tier.
-  repeated BackendAssociation backends = 4;
+  repeated BackendAssociation backends = 2;
+}
 
+// Represents the current operational state of a StorageTier.
+message StorageTierStatus {
   // Current lifecycle state of the storage tier.
-  StorageTierState state = 5;
+  StorageTierState state = 1;
```
ygalblum, line 97: From what I can tell (and I might be missing it), `state` is currently not changed after creation. Who is expected to update the `state`? Is it a status report form OSAC to the user or can the user update it? If it's a status, then all good. If the user can set it, then it should be in the spec.

ygalblum (commented):
Small question, otherwise all good

proto/private/osac/private/v1/storage_tier_type.proto
rgolangh: That is why the initial implementation was not to include a status field at all, because there is no state transition.  There might a state in the future for maintenance or some state to infer 'not ready' . 
Also, no the (CSP admin) user should not change it.

ygalblum: Thanks for the explanation

ygalblum (approved):
/approve
/lgtm

akshaynadkarni (approved):
LGTM

PR #892, opened 2026-07-14
OSAC-2042: add primary field validation to BareMetalNetworkAttachment

danmanor:
## [OSAC-2042](https://redhat.atlassian.net/browse/OSAC-2042): add primary field validation to BareMetalNetworkAttachment

**Jira:** https://redhat.atlassian.net/browse/OSAC-2042
**Epic:** [OSAC-1448](https://redhat.atlassian.net/browse/OSAC-1448) — BareMetalInstance Network Attachments

### Summary

Adds CEL validation on `BareMetalInstanceSpec` to enforce the multi-NIC primary designation rule: when multiple `network_attachments` are specified, exactly one must have `primary` set to `true`. Single-attachment instances are exempt — primary is implicit per the EP.

The `primary` field itself was already added by OSAC-1508. This PR adds the declarative validation via `buf.validate` CEL annotations on both the public and private proto definitions.

### Changes

- **Proto (public + private):** Add `buf.validate.message.cel` rule on `BareMetalInstanceSpec` with expression `this.network_attachments.size() <= 1 || this.network_attachments.filter(a, a.primary == true).size() == 1`
- **Generated code:** Regenerated Go code via `buf generate`
- **Tests:** 8 new protovalidate unit tests covering: no attachments, single attachment (with/without primary), multiple attachments (valid exactly-one-primary, zero primary, multiple primary, explicit false on non-primary)
- **Test fix:** Fixed pre-existing compilation error in `catalog_item_validation_test.go` where removed `Cores` field was still referenced (replaced with `UserData`)

### Testing

- **Unit tests:** 8 new tests in `private_baremetal_instances_server_test.go` under "Network attachment primary validation" — validate the CEL rule directly via `protovalidate.Validator`
- **buf lint:** Clean
- **All 47 BMI server tests pass**

### Acceptance Criteria

- [x] CEL validation rejects multiple network_attachments with no primary
- [x] CEL validation rejects multiple network_attachments with >1 primary
- [x] CEL validation accepts multiple network_attachments with exactly one primary
- [x] CEL validation accepts single attachment without primary (implicit)
- [x] Primary field is annotated IMMUTABLE on public API


proto/public/osac/public/v1/baremetal_instance_type.proto
```diff
@@ -64,6 +65,11 @@ message BareMetalNetworkAttachment {
 
 // The spec contains the details of a bare metal instance as desired by the user.
 message BareMetalInstanceSpec {
+  option (buf.validate.message).cel = {
+    id: "baremetal_instance_primary_attachment"
+    message: "when multiple network attachments are specified, exactly one must have primary set to true"
+    expression: "this.network_attachments.size() <= 1 || this.network_attachments.filter(a, a.primary == true).size() == 1"
```
ori-amizur, line 71: 1. Is 0 attachments allowed ?
2. If there is a single attachment without primary - will there be a default route ?

danmanor: 1. yes, the default will be used. 2. primary will be inferred (only one interface)

PR #897, opened 2026-07-14
OSAC-2822: skip CI for docs-only and metadata-only PRs

rccrdpccl:
## Summary

- Add `paths-ignore` filters to all PR-triggered workflows so they don't run when only non-code files change (OWNERS, LICENSE, `*.md`, `docs/**`).
- Affected workflows: check-pull-request, publish-image, e2e-vmaas-full-install, unit-tests, integration-tests.
- Scheduled and workflow_dispatch triggers are unaffected.

## Branch protection

Requires switching any required status checks from classic branch protection to a GitHub Ruleset with "Do not require status checks on creation" enabled, so the check is enforced on code PRs but doesn't block docs-only PRs.

## Test plan

- [ ] Verify a code-changing PR still triggers all workflows normally
- [ ] Confirm docs-only PRs can merge without waiting for skipped checks

### DIFF
--- a/.github/workflows/check-pull-request.yaml
+++ b/.github/workflows/check-pull-request.yaml
@@ -18,6 +18,11 @@ on:
   pull_request:
     branches:
     - main
+    paths-ignore:
+      - 'OWNERS'
+      - 'LICENSE'
+      - '*.md'
+      - 'docs/**'
 
 permissions:
   contents: read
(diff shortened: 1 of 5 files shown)


ygalblum:
AFAIK using `paths-ignore` blocks the workflow from running altogether. As a result, if the workflow is set as required, the PR is stuck.
In other projects we split the jobs and returned success without actually testing. For example: https://github.com/dcm-project/shared-workflows/blob/main/.github/workflows/check-generate.yaml

rccrdpccl:
> AFAIK using `paths-ignore` blocks the workflow from running altogether. As a result, if the workflow is set as required, the PR is stuck. In other projects we split the jobs and returned success without actually testing. For example: https://github.com/dcm-project/shared-workflows/blob/main/.github/workflows/check-generate.yaml

yes you're right, unfortunately we cannot use the same pattern due to reusable workflow, but we changed to something similar

PR #900, opened 2026-07-14
OSAC-2185: Gate chart publish on image build success

minmzzhang:
## Summary

`publish-charts.yaml` and `publish-image.yaml` currently trigger independently off the same `v*` tag push, with no dependency between them — a chart can be marked "published" even if the matching container image never got built. This exact class of gap caused the [OSAC-2174](https://redhat.atlassian.net/browse/OSAC-2174) production incident in a sibling repo (chart referenced an image that was never built, resulting in `ImagePullBackOff`).

This PR changes `publish-charts.yaml` to trigger via `workflow_run` on `Container image` completing, gated by a new `guard` job:

- Skips (neutral, not a false green) if the completion isn't from a tag push (e.g. PR builds, pushes to `main`).
- **Fails loudly (red)** if the image build for the tag did not succeed.
- Only lets the chart-publish job proceed if the image build for that *same tag* succeeded.

The `publish-service-chart` job now `needs: guard`, pins its checkout to `needs.guard.outputs.sha`, and uses `needs.guard.outputs.tag` instead of `$GITHUB_REF_NAME` (which doesn't reliably resolve to the tag under a `workflow_run` event).

## Testing

Live end-to-end verified on a personal fork, not just static analysis:
- Pushed a `v*` test tag → image build ran → succeeded → `publish-charts` triggered afterward (not in parallel) → chart published. Full run: green.
- Confirmed the negative path: a `push` to `main` (non-tag) completing the image workflow correctly results in `publish-charts` being **skipped**, not a false green.
- Validated with `actionlint` (0 errors) on the final workflow file.

## Acceptance criteria ([OSAC-2185](https://redhat.atlassian.net/browse/OSAC-2185))

- [x] Chart publish cannot succeed unless the matching image was confirmed to exist for that tag.
- [x] A simulated image-build failure on a tag push causes `publish-charts` to fail loudly rather than complete green.

Jira: https://redhat.atlassian.net/browse/OSAC-2185

## Summary by CodeRabbit

* **Bug Fixes**
  * Chart publishing now starts only after the related container image workflow completes successfully, reducing the chance of releasing mismatched artifacts.
  * Publishing is now validated against a semver-compatible release tag format to prevent invalid chart version updates.
  * Charts are packaged from the exact commit that produced the image, with chart versioning derived from the upstream release tag.


minmzzhang:
Pushed a follow-up commit addressing the CodeRabbit review:

- `guard` job now runs with `permissions: {}` (was inheriting default token scope)
- `workflow_run.head_branch`/`conclusion` are now passed through `env:` instead of being spliced directly into the shell script (script/template-injection hardening)
- The tag is now validated against a proper semver grammar in `guard`, not just `startsWith(..., 'v')`
- `persist-credentials: false` and a SHA-pinned `actions/checkout` on the touched checkout step(s)
- Added a step that re-verifies the tag still resolves to the guarded commit immediately before creating the GitHub Release, to guard against a force-push/retag race between the image build and the release

Intentionally not doing in this PR: Sigstore/cosign signing of the published chart artifacts. That's a separate, larger effort (registry/verification infra, signing key strategy) that's outside [OSAC-2185](https://redhat.atlassian.net/browse/OSAC-2185)'s scope (gating chart publish on image build success). Happy to track it as its own follow-up ticket if wanted.

minmzzhang:
> Caution
> 
> Some comments are outside the diff and can’t be posted inline due to platform limitations.
> 
> ⚠️ Outside diff range comments (1)
> > .github/workflows/publish-charts.yaml (1)> `86-89`: _🔒 Security & Privacy_ | _🟠 Major_ | _🏗️ Heavy lift_
> > > **Chart is still pushed unsigned.**
> > > `helm push` publishes the OCI chart without a Sigstore/cosign signature. This mirrors a previously flagged major finding; per commit message the signing work is intentionally deferred, but as per path instructions ("Sign artifacts with Sigstore/cosign"), this should be tracked and addressed in a follow-up before release.
> > > 
> > > 🤖 Prompt for AI Agents
> > > ```
> > > Verify each finding against current code. Fix only still-valid issues, skip the
> > > rest with a brief reason, keep changes minimal, and validate.
> > > 
> > > In @.github/workflows/publish-charts.yaml around lines 86 - 89, Update the chart
> > > publishing flow after helm package creation and before or during the OCI push to
> > > sign the chart artifact with the repository’s established Sigstore/cosign
> > > workflow. Ensure the artifact pushed by helm push is the signed chart and
> > > preserve the existing chart_name, chart_file, and registry destination behavior.
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > 
> > > 
> > > _Source: Path instructions_
> 
> 🤖 Prompt for all review comments with AI agents
> ```
> Verify each finding against current code. Fix only still-valid issues, skip the
> rest with a brief reason, keep changes minimal, and validate.
> 
> Outside diff comments:
> In @.github/workflows/publish-charts.yaml:
> - Around line 86-89: Update the chart publishing flow after helm package
> creation and before or during the OCI push to sign the chart artifact with the
> repository’s established Sigstore/cosign workflow. Ensure the artifact pushed by
> helm push is the signed chart and preserve the existing chart_name, chart_file,
> and registry destination behavior.
> ```
> 
> ℹ️ Review info

the signer will be implemented in a separate feature.

minmzzhang:
Applying the same fix pattern used for this identical workflow across the [OSAC-2185](https://redhat.atlassian.net/browse/OSAC-2185) PRs:

- Fixed the SemVer regex to reject leading zeros and `+build` metadata (Docker/OCI tags can't contain `+`)
- Routed the remaining direct `${{ secrets.GITHUB_TOKEN }}`/`${{ github.actor }}`/`${{ github.repository }}` interpolations through `env:` instead of splicing them into the shell

rccrdpccl:
/lgtm
/approve

as discussed offline with @minmzzhang we should extract common parts to a common workflow later

rccrdpccl (approved):

PR #903, opened 2026-07-15
OSAC-1905: surface reconciliation errors in ComputeInstance status

eranco74:
## Summary

- When the ComputeInstance reconciler fails (e.g. due to missing `runStrategy`), set status to `FAILED` with error details in the `PROVISIONED` condition
- Previously, reconciliation errors were only returned to controller-runtime for retry, leaving the instance stuck in `STARTING` with no user-visible indication
- Now the FAILED status is persisted via Update before returning the error, so users can see why their instance failed

## Changes

- Modified `run()` to capture reconcile errors, call `setReconciliationFailed()` to update status, persist via Update, then return the original error for retry
- Added `setReconciliationFailed()` method that sets state to FAILED and records the error message in the PROVISIONED condition
- Updated existing "no hubs available" test to expect the new FAILED status behavior
- Added 3 new unit tests for `setReconciliationFailed`: basic failure, missing status creation, and no duplicate conditions

## Test plan

- [x] `go build ./...` — compiles cleanly
- [x] `ginkgo run internal/controllers/computeinstance/` — 38/38 specs pass (3 new)
- [x] `go vet` — clean
- [x] `gofmt` — clean
- [ ] Deploy and verify: create a ComputeInstance without `run_strategy`, confirm status shows FAILED with error message

Supersedes #856

Fixes: [OSAC-1905](https://redhat.atlassian.net/browse/OSAC-1905)

## Summary by CodeRabbit

* **Bug Fixes**
  * Compute instance reconciliation failures are now recorded in the instance status.
  * Failed reconciliations set the instance state to **FAILED** and mark provisioning as unsuccessful with an explanatory message.
  * Failure status is persisted even when errors occur during updates, deletion, or hub selection.

### DIFF
--- a/internal/controllers/computeinstance/computeinstance_reconciler_function.go
+++ b/internal/controllers/computeinstance/computeinstance_reconciler_function.go
@@ -135,26 +135,29 @@ func (r *function) run(ctx context.Context, computeInstance *privatev1.ComputeIn
 		r:               r,
 		computeInstance: computeInstance,
 	}
-	var err error
+	var reconcileErr error
 	if computeInstance.HasMetadata() && computeInstance.GetMetadata().HasDeletionTimestamp() {
-		err = t.delete(ctx)
+		reconcileErr = t.delete(ctx)
 	} else {
-		err = t.update(ctx)
+		reconcileErr = t.update(ctx)
 	}
-	if err != nil {
-		return err
+	if reconcileErr != nil {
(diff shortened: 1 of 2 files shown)


rccrdpccl:
@eranco74 please address/dismiss coderabitt's comment, LGTM otherwise

ygalblum (approved):
/approve
/lgtm

PR #907, opened 2026-07-15
OSAC-2156: Implement ClusterVersion server

sk-ilya:
## Summary

- Extract `validateCELSyntax` and composable per-field filter defaults (`composeFilterDefaults`) into `cel_filter.go`, refactoring `InstanceTypesServer` to use the shared helper
- Implement public and private ClusterVersion gRPC servers with state lifecycle (ACTIVE/DEPRECATED/OBSOLETE), `is_default` management with atomic swap via partial unique index, immutability guards on version/image/name, SemVer validation, and auto-generated DNS-safe names
- Public server uses composable filter defaults (state + enabled), maps `is_default` as read-only, and delegates to the private server preserving private-only fields (e.g. `spec.image`)
- Add OPA policy for ClusterVersions public access and integration tests for access control

## Test plan

- [x] `ginkgo run -r internal` -- all 80 suites pass
- [x] `uv run dev.py lint` -- 0 issues
- [x] `buf lint` -- clean
- [x] Integration tests (`ginkgo run it`) -- requires kind cluster

Assisted-by: Claude <noreply@anthropic.com>

## Summary by CodeRabbit

* **New Features**
  * Added public and private Cluster Versions APIs (list, view, create, update, delete, signaling) with default visibility and filter-aware behavior.
  * Improved Cluster Versions and Instance Types list filtering by composing default predicates with CEL expressions.
* **Bug Fixes**
  * Retained deprecation/obsolescence timestamps when transitioning back to active status.
  * Expanded authorization to allow additional cluster-versions gRPC methods for permitted client-permission subjects.
* **Documentation**
  * Updated cluster-version proto comments to reflect timestamp retention semantics.
* **Tests**
  * Added end-to-end and unit test coverage for public/private Cluster Versions behavior and filtering utilities.

### DIFF
--- a/internal/servers/catalog_item_validation_test.go
+++ b/internal/servers/catalog_item_validation_test.go
@@ -352,22 +352,4 @@ var _ = Describe("addPublishedFilter", func() {
 		Entry("unbalanced opening paren", `(true`),
 	)
 
-	DescribeTable("validateCELSyntax",
-		func(input string, shouldPass bool) {
-			err := validateCELSyntax(input)
-			if shouldPass {
-				Expect(err).ToNot(HaveOccurred())
-			} else {
-				Expect(err).To(HaveOccurred())
-			}
-		},
-		Entry("valid simple expression", "true", true),
-		Entry("valid field reference", "this.published", true),
-		Entry("valid comparison", "this.id == '123'", true),
-		Entry("valid compound", "this.a && this.b || this.c", true),
-		Entry("unbalanced closing paren", "true)", false),
-		Entry("unbalanced opening paren", "(true", false),
-		Entry("injection attempt", `true) || (true`, false),
-		Entry("empty string is not valid CEL", "", false),
-	)
 })
--- a/internal/servers/cel_filter.go
+++ b/internal/servers/cel_filter.go
@@ -1,0 +1,157 @@
+/*
+Copyright (c) 2025 Red Hat Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with the
+License. You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific
+language governing permissions and limitations under the License.
+*/
+
+package servers
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/google/cel-go/cel"
+	"github.com/google/cel-go/common/ast"
+)
+
+var celSyntaxEnv = func() *cel.Env {
+	env, err := cel.NewEnv()
+	if err != nil {
+		panic(fmt.Sprintf("failed to create CEL environment: %v", err))
+	}
+	return env
+}()
+
+// validateCELSyntax checks that a filter string is a syntactically valid, complete CEL expression.
+// This prevents filter bypass attacks where a malicious filter like "true) || (true" could
+// break out of parenthesized composition and change operator precedence.
+func validateCELSyntax(filter string) error {
+	_, issues := celSyntaxEnv.Parse(filter)
+	if issues != nil && issues.Err() != nil {
+		return fmt.Errorf("syntax error: %w", issues.Err())
+	}
+	return nil
+}
+
+// filterReferencedFields parses a CEL filter and returns the set of all dotted field paths
+// referenced in the expression (e.g. "this.spec.state", "this.metadata.name"). Paths that do not
+// root at an identifier (e.g. function call results) are excluded.
+func filterReferencedFields(filter string) (map[string]bool, error) {
+	parsed, issues := celSyntaxEnv.Parse(filter)
+	if issues != nil && issues.Err() != nil {
+		return nil, fmt.Errorf("syntax error: %w", issues.Err())
+	}
+	fields := make(map[string]bool)
+	nav := ast.NavigateAST(parsed.NativeRep())
+	ast.MatchDescendants(nav, func(e ast.NavigableExpr) bool {
+		if e.Kind() != ast.SelectKind {
+			return false
+		}
+		path := resolveSelectPath(e.AsSelect())
+		if path != "" {
+			fields[path] = true
+		}
+		return false
+	})
+	return fields, nil
+}
+
+// fieldMatchesPrefix reports whether a dotted field path equals the prefix or is a direct child of
+// it (e.g. "this.spec.state" matches prefix "this.spec.state", and "this.spec.state.sub" matches
+// too, but "this.spec.state_machine" does not).
+func fieldMatchesPrefix(field, prefix string) bool {
+	return field == prefix || strings.HasPrefix(field, prefix+".")
+}
+
+// filterReferencesAnyField reports whether the CEL filter expression references a field whose dotted
+// path starts with any of the given prefixes (e.g. "this.spec.state"). Returns an error if
+// the filter has invalid syntax.
+func filterReferencesAnyField(filter string, prefixes ...string) (bool, error) {
+	fields, err := filterReferencedFields(filter)
+	if err != nil {
+		return false, err
+	}
+	for field := range fields {
+		for _, prefix := range prefixes {
+			if fieldMatchesPrefix(field, prefix) {
+				return true, nil
+			}
+		}
+	}
+	return false, nil
+}
+
+// filterDefault pairs a field prefix with the CEL predicate to apply when the user's filter does
+// not reference that field. Used with composeFilterDefaults for per-field default composition.
+type filterDefault struct {
+	field     string // field prefix to check, e.g. "this.spec.state"
+	predicate string // CEL predicate applied when field is absent, e.g. "this.spec.state == 1"
+}
+
+// composeFilterDefaults composes per-field default predicates with a user-provided filter. For each
+// default, the predicate is applied only if the user's filter does not reference that field. The
+// filter is parsed and walked once regardless of the number of defaults.
+func composeFilterDefaults(filter string, defaults []filterDefault) (string, error) {
+	if filter == "" {
+		clauses := make([]string, len(defaults))
+		for i, d := range defaults {
+			clauses[i] = d.predicate
+		}
+		return strings.Join(clauses, " && "), nil
+	}
+
+	fields, err := filterReferencedFields(filter)
+	if err != nil {
+		return "", fmt.Errorf("invalid filter: %w", err)
+	}
+
+	var missing []string
+	for _, d := range defaults {
+		referenced := false
+		for field := range fields {
+			if fieldMatchesPrefix(field, d.field) {
+				referenced = true
+				break
+			}
+		}
+		if !referenced {
+			missing = append(missing, d.predicate)
+		}
+	}
+
+	if len(missing) == 0 {
+		return filter, nil
+	}
+	return "(" + filter + ") && " + strings.Join(missing, " && "), nil
+}
+
+// resolveSelectPath reconstructs the full dotted field path from a chain of select expressions
+// rooted at an identifier. For example, this.spec.state becomes "this.spec.state".
+// Returns "" if the chain does not terminate at an identifier (e.g. a function call result).
+func resolveSelectPath(sel ast.SelectExpr) string {
+	var parts []string
+	for {
+		parts = append(parts, sel.FieldName())
+		operand := sel.Operand()
+		switch operand.Kind() {
+		case ast.SelectKind:
+			sel = operand.AsSelect()
+		case ast.IdentKind:
+			parts = append(parts, operand.AsIdent())
+			// Reverse to get root-to-leaf order:
+			for i, j := 0, len(parts)-1; i < j; i, j = i+1, j-1 {
+				parts[i], parts[j] = parts[j], parts[i]
+			}
+			return strings.Join(parts, ".")
+		default:
+			return ""
+		}
+	}
+}
--- a/internal/servers/cluster_versions_server.go
+++ b/internal/servers/cluster_versions_server.go
@@ -1,0 +1,323 @@
+/*
+Copyright (c) 2025 Red Hat Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with the
+License. You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific
+language governing permissions and limitations under the License.
+*/
+
+package servers
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"log/slog"
+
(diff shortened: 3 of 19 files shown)


internal/servers/catalog_item_validation_test.go
```diff
@@ -180,146 +150,6 @@ var _ = Describe("applyFieldDefinitions", func() {
 	})
 })
 
-var _ = Describe("applyFieldDefinitions rejects unlisted fields", func() {
-	It("rejects a single unlisted field on ClusterSpec", func() {
-		pullSecret := "my-secret"
```
sk-ilya, line 185: Consolidated in `cel_filter_test.go`

internal/servers/private_cluster_versions_server.go
```diff
@@ -0,0 +1,511 @@
...
+// appendHexSuffix truncates the name to 58 characters and appends a 4-character FNV-32a hex suffix
+// derived from the original version string, producing a 63-character DNS-safe name.
+func appendHexSuffix(name string, version string) string {
+	if len(name) > 58 {
+		name = name[:58]
+	}
+	name = strings.TrimRight(name, "-")
+	h := fnv.New32a()
+	h.Write([]byte(version))
+	return fmt.Sprintf("%s-%04x", name, h.Sum32()%0x10000)
```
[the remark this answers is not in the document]
sk-ilya: Resolved

internal/servers/private_cluster_versions_server.go
```diff
@@ -0,0 +1,511 @@
...
+	if oldState == newState {
+		if requestDep := spec.GetDeprecation(); requestDep != nil {
+			if existingDep := existing.GetSpec().GetDeprecation(); existingDep != nil {
+				requestDep.SetDeprecationTimestamp(existingDep.GetDeprecationTimestamp())
+				requestDep.SetObsolescenceTimestamp(existingDep.GetObsolescenceTimestamp())
+			} else {
+				requestDep.ClearDeprecationTimestamp()
+				requestDep.ClearObsolescenceTimestamp()
+			}
+		}
```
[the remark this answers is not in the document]
sk-ilya: Resolved

internal/servers/private_cluster_versions_server.go
```diff
@@ -0,0 +1,511 @@
...
+		return nil, err
+	}
+
+	existing := getResponse.GetObject()
+
+	if err := validateClusterVersionImmutability(existing, request); err != nil {
+		return nil, err
+	}
+
+	applyClusterVersionStateEffects(existing, request)
```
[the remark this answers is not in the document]
sk-ilya: Resolved

internal/servers/cluster_versions_server_test.go
```diff
@@ -0,0 +1,223 @@
...
+			})
+
+			getResponse, err := publicServer.Get(ctx, publicv1.ClusterVersionsGetRequest_builder{
+				Id: cv.GetId(),
+			}.Build())
+			Expect(err).ToNot(HaveOccurred())
+			Expect(getResponse.GetObject().GetSpec().GetIsDefault()).To(BeTrue())
+		})
+
+		It("Public Update ignores caller-set is_default", func() {
```
ygalblum, line 193: Is silently ignoring the expected behaviour?  What can the user update via the public APIs?

internal/servers/private_cluster_versions_server.go
```diff
@@ -0,0 +1,502 @@
...
+	applyClusterVersionDefaults(cv)
+
+	// Clear caller-provided ID so the DAO always generates a UUID:
+	cv.SetId("")
+
+	// Safe to clear existing defaults before Create: the gRPC interceptor wraps the entire
+	// RPC in a single transaction, so the clear and Insert commit together — other connections
+	// never see a state with zero defaults.
+	if cv.GetSpec().GetIsDefault() {
+		if err := validateIsDefaultEligibility(cv.GetSpec().GetState(), cv.GetSpec().GetEnabled()); err != nil {
```
ygalblum: Isn't it simpler to pass the `cv` or `cv.GetSpec()` once?

internal/servers/private_cluster_versions_server.go
```diff
@@ -0,0 +1,502 @@
...
+		return nil, grpcstatus.Errorf(grpccodes.InvalidArgument, "object spec is mandatory when updating spec fields")
+	}
+
+	if err := validateClusterVersionImmutability(existing, request); err != nil {
+		return nil, err
+	}
+
+	applyClusterVersionStateEffects(existing, request)
+
+	if resolveIsDefault(existing, request) {
```
ygalblum, line 197: Maybe I'm missing something here. From what I see, when the update does not change `spec.is_default`, `resolveIsDefault` return the current state. So, updating the current default will return `true` which will trigger unsetting it.

internal/servers/private_cluster_versions_server.go
```diff
@@ -0,0 +1,502 @@
...
+// Concurrent default-swap requests are safe: the unique partial index
+// cluster_versions_single_default (migration 74) prevents two concurrent transactions from both
+// committing is_default=true. The losing transaction receives AlreadyExists from the DAO;
+// callers that need a more specific error code should remap it (see Update).
+func (s *PrivateClusterVersionsServer) unsetPreviousDefaultClusterVersion(ctx context.Context, excludeID string) error {
+	filter := "this.spec.is_default == true && !has(this.metadata.deletion_timestamp)"
+	if excludeID != "" {
+		filter += fmt.Sprintf(" && this.id != %s", strconv.Quote(excludeID))
+	}
+	listResponse, err := s.generic.dao.List().
```
ygalblum, line 245: Should we verify the length of the list?

internal/servers/private_cluster_versions_server_test.go
```diff
@@ -0,0 +1,988 @@
...
+
+				// Verify first is no longer default:
+				getResponse, err := server.Get(ctx, privatev1.ClusterVersionsGetRequest_builder{
+					Id: first.GetObject().GetId(),
+				}.Build())
+				Expect(err).ToNot(HaveOccurred())
+				Expect(getResponse.GetObject().GetSpec().GetIsDefault()).To(BeFalse())
+			})
+
+			It("Update setting is_default clears previous default", func() {
```
ygalblum, line 733: Regarding my comment about updating the current default, there's no test for this use case

internal/servers/private_cluster_versions_server_test.go
```diff
@@ -0,0 +1,988 @@
...
+				}.Build())
+				Expect(err).ToNot(HaveOccurred())
+				Expect(getResponse.GetObject().GetSpec().GetIsDefault()).To(BeFalse())
+			})
+
+			It("Auto-clears is_default on OBSOLETE version", func() {
+				object := createWithState("clear-default-obsolete", "4.17.0",
+					privatev1.ClusterVersionState_CLUSTER_VERSION_STATE_OBSOLETE)
+
+				_, err := server.Update(ctx, privatev1.ClusterVersionsUpdateRequest_builder{
```
ygalblum, line 876: Should't it fail instead of silently ignoring the request?

internal/servers/private_cluster_versions_server_test.go
```diff
@@ -0,0 +1,988 @@
...
+
+				getResponse, err := server.Get(ctx, privatev1.ClusterVersionsGetRequest_builder{
+					Id: cv.GetObject().GetId(),
+				}.Build())
+				Expect(err).ToNot(HaveOccurred())
+				Expect(getResponse.GetObject().GetSpec().GetIsDefault()).To(BeFalse())
+			})
+
+			It("Rejects is_default on OBSOLETE create", func() {
+				_, err := server.Create(ctx, privatev1.ClusterVersionsCreateRequest_builder{
```
ygalblum, line 942: Here create fails such a request. But, update doesn't, just silently ignores.

internal/servers/cluster_versions_server_test.go
sk-ilya: Yes - In general, the public API is read-only (see  [authz.rego](https://github.com/osac-project/fulfillment-service/pull/907/changes#diff-c39a0c44bd0dbc166e2c380752ba6c38eaff16620122aa32037de898acb2c242)) while setting images and `is_default` should be done by admins through the private API.

internal/servers/private_cluster_versions_server.go
sk-ilya: >  Maybe I'm missing something here. From what I see, when the update does not change spec.is_default, resolveIsDefault return the current state. 

Correct - it returns the effective state. 

> So, updating the current default will return true which will trigger unsetting it.

If it was already `true`, it will still follow the same path and attempt to unset `is_default` on the previously default cluster version. Since the cluster version being updated is excluded from that lookup, no other default will be found, and `unsetPreviousDefaultClusterVersion` will just no-op.

This adds one extra query, but the overhead is negligible for such an infrequent operation, so I preferred to keep the implementation simpler

internal/servers/private_cluster_versions_server.go
sk-ilya: Why? It's at most 1

internal/servers/private_cluster_versions_server_test.go
sk-ilya: It should automatically clear `is_default`. However, I added validation to reject update requests that explicitly attempt to set an obsolete version as the default

internal/servers/private_cluster_versions_server_test.go
sk-ilya: Added "Updating unrelated field on current default preserves is_default"

internal/servers/private_cluster_versions_server_test.go
sk-ilya: See https://github.com/osac-project/fulfillment-service/pull/907#discussion_r3594988142

internal/servers/cluster_versions_server.go
```diff
@@ -0,0 +1,323 @@
...
+	privateRequest.SetOffset(request.GetOffset())
+	privateRequest.SetLimit(request.GetLimit())
+	composedFilter, err := composeFilterDefaults(request.GetFilter(), clusterVersionFilterDefaults)
+	if err != nil {
+		return nil, grpcstatus.Errorf(grpccodes.InvalidArgument, "%v", err)
+	}
+	privateRequest.SetFilter(composedFilter)
+	privateRequest.SetOrder(request.GetOrder())
+
+	privateResponse, err := s.delegate.List(ctx, privateRequest)
```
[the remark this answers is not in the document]
sk-ilya: Other servers have  the same gap... Fixing it in one place but not the others is arguably worse than not fixing it at all

internal/servers/private_cluster_versions_server.go
```diff
@@ -0,0 +1,517 @@
...
+	mask := request.GetUpdateMask()
+
+	// Same-state update — enforce OUTPUT_ONLY contract on deprecation timestamps by
+	// unconditionally restoring existing values, regardless of what the client sent:
+	if oldState == newState {
+		if existingDep := existing.GetSpec().GetDeprecation(); existingDep != nil {
+			spec.SetDeprecation(proto.Clone(existingDep).(*privatev1.ClusterVersionDeprecation))
+		} else {
+			spec.ClearDeprecation()
+		}
```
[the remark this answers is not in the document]
sk-ilya: Done

internal/servers/private_cluster_versions_server_test.go
```diff
@@ -0,0 +1,1005 @@
...
+				}.Build())
+				Expect(err).ToNot(HaveOccurred())
+
+				// Verify first is no longer default:
+				getResponse, err := server.Get(ctx, privatev1.ClusterVersionsGetRequest_builder{
+					Id: first.GetObject().GetId(),
+				}.Build())
+				Expect(err).ToNot(HaveOccurred())
+				Expect(getResponse.GetObject().GetSpec().GetIsDefault()).To(BeFalse())
+			})
```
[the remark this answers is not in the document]
sk-ilya: Done

it/it_access_test.go
```diff
@@ -63,6 +63,21 @@ var _ = Describe("Access control", func() {
...
+		})
+
+		It("Denies regular users creating cluster versions", func() {
+			client := publicv1.NewClusterVersionsClient(tool.ExternalView().UserConn())
+			_, err := client.Create(ctx, publicv1.ClusterVersionsCreateRequest_builder{}.Build())
+			Expect(err).To(HaveOccurred())
+			status, ok := grpcstatus.FromError(err)
+			Expect(ok).To(BeTrue())
+			Expect(status.Code()).To(Equal(grpccodes.PermissionDenied))
+		})
```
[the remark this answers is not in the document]
sk-ilya: Done

internal/servers/cluster_versions_server_test.go
ygalblum: If the public API is Read-Only, why does it expose the other operations? Isn't it better to not support them at all? Like here: https://github.com/osac-project/fulfillment-service/blob/2f3f87208e0740fad65ef4902425d6ae857d8fbb/proto/public/osac/public/v1/external_ip_pools_service.proto

internal/servers/private_cluster_versions_server.go
ygalblum: Maybe add a comment about it? Not a must

internal/servers/private_cluster_versions_server.go
ygalblum: That was my point. Make sure that it in fact 1 and nothing went wrong. Again, not a must only a suggestion

ygalblum (approved):
It seems that for RO, some services rely on auth while others do not implement the endpoint. Generally, I think that there is no point in having an endpoint that is always blocked. But, I guess it shouldn't hold this PR. Same for my suggestions.

/approve
/lgtm

internal/servers/cluster_versions_server_test.go
sk-ilya: `ExternalIPPools` and `PublicIPPools` are the only CRUD-ish servers that do not implement some of these methods. I'm not sure that is correct, so I preferred to keep this consistent. At the very least, the generic CLI uses reflection and expects all of these methods to be present, and I'm not sure what else might rely on them

internal/servers/private_cluster_versions_server.go
sk-ilya: It's already guaranteed by the partial unique index. I can add a warn-level log in a follow-up PR, I guess

internal/servers/private_cluster_versions_server.go
sk-ilya: Will do in the next PR

internal/servers/cel_filter.go
```diff
@@ -0,0 +1,157 @@
...
+package servers
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/google/cel-go/cel"
+	"github.com/google/cel-go/common/ast"
+)
+
```
tzvatot, line 23: 💡 **`celSyntaxEnv` initialization panics instead of propagating errors**

The old `sync.Once` pattern propagated CEL env creation errors. The new package-level init panics on failure. `cel.NewEnv()` with no options should never fail in practice, so this is fine - just noting the behavior change. If a future caller adds CEL options that could fail, this would need to be revisited.

tzvatot (approved):
/approve
/lgtm

tzvatot:
💡 **No test for deleting the current default version** - Default-swap is tested for Create and Update, but not Delete. After deleting the default, no version would be the default. May be intentional (admin choice) or a gap worth validating.

PR #911, opened 2026-07-15
NO-ISSUE: Remove breakglass account password

CrystalChun:
Previously, we accidentally stored this password in our database long-term, but we should clear it after we've created the tenant and breakglass accounts in Keycloak. 

We should also remove the breakglass account credentials from previously stored tenants using the migration.

The flow now is:
1. Cloud provider admin creates tenant and gets the breakglass credentials
2. The tenant and the breakglass credentials are temporarily stored in the DB
3. The tenant is reconciled and the breakglass account and tenant are created in Keycloak
4. The breakglass account credentials are removed from the tenant in the DB

Get tenant now shows only the ID of the breakglass account:
```sh
$ ./osac get tenant -oyaml osac-demo 
'@type': type.googleapis.com/osac.private.v1.Tenant
id: osac-demo
metadata:
  creation_timestamp: "2026-07-16T20:51:09.143519Z"
  creator: system
  finalizers:
    - fulfillment-controller
  labels:
    region: us-east
    tier: enterprise
  name: osac-demo
  tenant: osac-demo
  version: 2
spec: {}
status:
  break_glass_user_id: dd99681b-5ff4-4eb9-bc89-b85e539982c3
  idp_tenant_name: osac-demo
  state: TENANT_STATE_SYNCED
```

Checked the DB as well and it no longer has the credentials
```sh
service=# select id, data->'status'->'break_glass_credentials' from tenants;
     id      | ?column? 
-------------+----------
 shared      | 
 system      | 
 development | 
 sales       | 
 users       | 
 engineering | 
 osac-demo   | 
(7 rows)

service=# 
```

/cc @jhernand @DakCrowder 

## Summary by CodeRabbit

* **Security**
  * Break-glass credentials are now one-time: credentials are cleared immediately after successful account creation and no longer persisted for later use.
  * Existing tenant records are cleaned up to remove stored break-glass username/password fields while keeping related identifiers intact.

* **Reliability**
  * Field-change calculation now correctly includes scenarios where fields are cleared between versions.
  * Migration behavior is verified with added test coverage for active and archived tenant records.


internal/controllers/tenant/tenant_reconciler_function.go
```diff
@@ -124,6 +132,10 @@ func (r *function) Run(ctx context.Context, tenant *privatev1.Tenant) error {
 		}.Build())
 	}
 
+	if savedCredentials != nil {
+		tenant.GetStatus().SetBreakGlassCredentials(savedCredentials)
+	}
```
CrystalChun: This will always show the credentials to the cloud provider admin if they query for a tenant. Is that ok? 

There's an option to strip the credentials on any action that's not `CREATE` in the private server... https://github.com/CrystalChun/fulfillment-service/compare/remove-bg-creds...CrystalChun:fulfillment-service:strip-creds?expand=1

May I get your input on this @jhernand / @DakCrowder?

jhernand: My understanding is that we are generating the password ourselves, as part of the reconciliation. Can we instead generate it inside the private tenants server? Then we show it to the user once, as the result of the `Create` operation. We then save it to the database, and during the reconciliation we use it to create the account in Keycloak and then we delete it from the database. There will be a short (hopefully) period of time where the password is stored in the database, but I think that is acceptable.

Once we have the secret support, can we store this password in a secret?

DakCrowder: If its taking place in the reconciliation then its async correct?  Which makes things more complicated - from the initial requirements it seems to have been thought of more synchronously where it just gets displayed once and never stored in fulfillment-service.

So it seems we either have to store the password in some fashion, or make it synchronous (either as part of the org create or could be a separate endpoint I guess?)

Additionally, something I just thought of - if we do make it a secret and a cloud provider admin creates an org should they be the one who can access the secret?  What tenant should the secret live under?

CrystalChun: Hmm my understanding from the [enhancement](https://github.com/osac-project/enhancement-proposals/tree/main/enhancements/organizations#keycloak-configuration-management) is that we shouldn't store this password at all

> The initial password is returned in the organization creation response to the Cloud Provider Admin (since the password is hashed in Keycloak storage, it cannot be retrieved later). The Cloud Provider Admin is responsible for securely communicating it to the appropriate personnel.

I like your suggestion Juan:

>  Can we instead generate it inside the private tenants server? Then we show it to the user once, as the result of the Create operation. 

But I think Dakota has a good point about sync vs async. 


---
> Additionally, something I just thought of - if we do make it a secret and a cloud provider admin creates an org should they be the one who can access the secret? What tenant should the secret live under?

Based off of the enhancement, it sounds like only the cloud provider admin should be able to access this, but would it make sense to allow tenant admins access? 

I think the secret should live under the tenant it's created for.

CrystalChun: Should we update the enhancement to say that the password will be stored temporarily? I agree that having it stored momentarily and removed later would be nicer.

If we go with this temporary storage, I'm not sure whether or not it's worth storing in a secret if we're going to remove it.

DakCrowder: Ah I think i slightly misunderstood Juan's proposal as well, so as I understand now:

1.  Create pw at creation time - return to user
2.  Save that PW in a (brief) interim step between synchronous create -> async reconciliation 
3.  Once pw is used to generate the actual break glass account, remove pw

Once secret flow exists it should be fairly simple to move it

CrystalChun: Gotcha! Thanks for both of your inputs. I've updated the changes to reflect this: creating the tenant will show the password to the admin and temporarily store it in the DB until the account is created in Keycloak.

jhernand (approved):

PR #934, opened 2026-07-19
OSAC-2157: Implement allowed_upgrades management

sk-ilya:
## Summary by CodeRabbit

- **New Features**
  - Added support for filtering objects by values in repeated fields, including strings, numbers, and booleans.
  - Added validation for allowed upgrade references, including eligibility and self-reference checks.
  - Automatically removes deleted version references from remaining cluster versions.

- **Bug Fixes**
  - Improved handling of database deadlocks by returning a retryable “Aborted” error.
  - Added validation for unique upgrade names, maximum list size, and name length limits.


internal/servers/private_cluster_versions_server.go
```diff
@@ -518,3 +567,144 @@ func generateNameFromVersion(version string) string {
...
+			strings.Join(quoted, ", "),
+		)).
+		SetLimit(1000).
+		Do(ctx)
+	if err != nil {
+		s.logger.ErrorContext(ctx, "Failed to look up allowed_upgrades targets",
+			slog.Any("error", err),
+		)
+		return grpcstatus.Errorf(grpccodes.Internal, "failed to validate allowed_upgrades")
+	}
```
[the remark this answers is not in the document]
sk-ilya: ClusterVersion is a global resource, single project

internal/database/dao/generic_dao_update.go
```diff
@@ -259,6 +259,10 @@ func (r *UpdateRequest[O]) translateError(ctx context.Context, id, name, tenant
 			)
 			return &ErrReference{}
 		}
+	case errReferenceCode:
+		return &ErrReference{
+			Reason: pgErr.Message,
+		}
```
rccrdpccl, line 265: Could this happen on create as well?

internal/servers/private_cluster_versions_server.go
```diff
@@ -518,3 +567,144 @@ func generateNameFromVersion(version string) string {
...
+		}
+		if len(listResponse.GetItems()) < 1000 {
+			break
+		}
+	}
+
+	// Sort IDs to prevent deadlocks when concurrent deletions lock
+	// overlapping row sets.
+	slices.Sort(ids)
+
```
rccrdpccl, line 657: This will prevent deadlocks within a single version, but what happens if we concurrently delete version A and version B where version A references B and B references A?

internal/servers/private_cluster_versions_server.go
```diff
@@ -518,3 +567,144 @@ func generateNameFromVersion(version string) string {
...
+		var filtered []string
+		for _, n := range names {
+			if n != deletedName {
+				filtered = append(filtered, n)
+			}
+		}
+		if filtered == nil {
+			filtered = []string{}
+		}
+		au.SetVersionNames(filtered)
```
rccrdpccl, line 697: IIUC this will delete other objects reference to the current version, however this is specifically forbidden in the API docs: https://github.com/osac-project/fulfillment-service/blob/main/docs/API.md#spec-and-status-ownership

I understand this is a design decision, however it conflicts with the definition. Should we change the definition, or the implementation?

internal/database/dao/filter_translator.go
```diff
@@ -988,6 +991,13 @@ func (t *FilterTranslator[O]) translateSelectJsonField(operandSql string, msgDes
 		return
 	}
 	fieldDesc := msgDesc.Fields().ByName(protoreflect.Name(fieldName))
+	// Repeated fields are stored as JSONB arrays — use -> to preserve the array structure:
+	if fieldDesc.IsList() {
```
rccrdpccl, line 1001: what happens if the list is not a string, but rather int type? AFAIU `?` only take a text operand

internal/database/dao/generic_dao_update.go
sk-ilya: Yes. It's handled the exact same way there:
https://github.com/osac-project/fulfillment-service/blob/ac71574f840ae31923ce5243e25e108568c7f018/internal/database/dao/generic_dao_create.go#L219-L222

rccrdpccl: oops sorry I've somehow missed it

internal/servers/private_cluster_versions_server.go
```diff
@@ -518,3 +567,144 @@ func generateNameFromVersion(version string) string {
 	h.Write([]byte(version))
 	return fmt.Sprintf("%s-%04x", result, h.Sum32()%0x10000)
 }
+
+// validateAllowedUpgrades validates that each name references an existing, enabled,
+// non-obsolete ClusterVersion that is not the object itself.
+func (s *PrivateClusterVersionsServer) validateAllowedUpgrades(
```
rccrdpccl, line 564: nit: can we extract inner parts of large methods? It would be much easier to read

sk-ilya: Extracted a couple of helpers

internal/database/dao/filter_translator.go
sk-ilya: Fixed

internal/servers/private_cluster_versions_server.go
sk-ilya: Nice catch :) but it's such a rare case, and PG will detect the deadlock anyway. I've added handling for deadlock errors so that they return `grpccodes.Aborted` instead of an internal error.

internal/servers/private_cluster_versions_server.go
sk-ilya: I agree it's not ideal, but `ClusterVersion`s are not reconciled, unlike most other objects. They are static entries managed by admins or externally.
The `spec`/`status` ownership rule is designed for reconciliation loops, where the user declares intent in spec and the system reports the observed state in status, so Im not sure it applies here. I also don't think it should be formalized in a general API doc unless we come up with a specification that all other similar resources will follow. So I'd keep both as they are, at least in the scope of the PR

internal/servers/private_cluster_versions_server.go
rccrdpccl: I agree, I think the solution is enough to address this

PR #947, opened 2026-07-22
OSAC-2914: Fix UNSPECIFIED state corruption in ClusterVersion Update

sk-ilya:
## Summary

- Fix `resolveState()` to treat `CLUSTER_VERSION_STATE_UNSPECIFIED` (proto3 zero value) as absent, preventing it from overwriting existing state during updates
- Normalize the request's state field before the generic server's mask merge so the correct value is persisted
- Add a `default` case in `applyClusterVersionStateEffects` to reject unhandled state values (defense-in-depth)

## Problem

When a client sends an Update with a field mask covering `spec.state` (directly or via parent `spec`) without explicitly setting state, the non-optional proto3 enum defaults to `UNSPECIFIED` (0). This zero value was passed through to the generic server's mask merge, which persisted it — corrupting the existing state. A secondary effect was spurious empty `ClusterVersionDeprecation` records.

Three trigger paths:
- `["spec"]` parent mask: generic server copies entire spec including zero-value state
- `["spec.state"]` leaf mask: `protoreflect.Has()` returns false for zero-value enum, triggering `Clear()` which also sets state to 0
- `nil` mask (full replacement): request object replaces the clone directly

## Test plan

- [x] 4 regression tests covering all three mask trigger paths and the deprecation side-effect
- [x] Full ClusterVersion test suite (78 passed, 0 failed)
- [x] Full servers test suite (1533 passed, 0 failed)

Assisted-by: Claude <noreply@anthropic.com>

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved the existing cluster version state when updates omit a state value or provide an unspecified state.
  * Prevented unnecessary deprecation record changes when the effective state remains unchanged.
  * Update requests now report errors from state-related processing instead of silently continuing.
  * Invalid cluster version states now return a clear validation error.

### DIFF
--- a/internal/servers/private_cluster_versions_server.go
+++ b/internal/servers/private_cluster_versions_server.go
@@ -209,7 +209,9 @@ func (s *PrivateClusterVersionsServer) Update(ctx context.Context,
 		}
 	}
 
-	applyClusterVersionStateEffects(existing, request)
+	if err := applyClusterVersionStateEffects(existing, request); err != nil {
+		return nil, err
+	}
 
 	if resolveIsDefault(existing, request) {
 		if err := s.unsetPreviousDefaultClusterVersion(ctx, id); err != nil {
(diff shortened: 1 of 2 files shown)


vladikr:
@sk-ilya looks very nice. 
It just needs a rebase and I'm happy to ack it.

PR #949, opened 2026-07-22
OSAC-2190: Mark floating image tags as placeholders, add CI guard

minmzzhang:
Mark floating `:latest`/`:main` image tags in `charts/*/values.yaml` as placeholders with inline comments, and add a CI workflow (`check-floating-tags.yaml`) that rejects unmarked floating tags on PRs.

- Add `# PLACEHOLDER -- overwritten at release time by publish-charts.yaml` comment above each floating tag
- Add `.github/scripts/check-floating-tags.sh` lint script
- Add `.github/workflows/check-floating-tags.yaml` CI guard on PR

No change to release-publish behavior -- the `publish-charts.yaml` workflow still patches these values at release time.

Ref: [OSAC-2190](https://redhat.atlassian.net/browse/OSAC-2190)

## Summary by CodeRabbit

* **New Features**
  * Added a repository validation to detect unmarked “floating” container image tags (for example, `latest` or `main`) in Helm chart `values.yaml`.
  * Introduced a CI pull request check that runs when relevant chart files or the validator change, and fails the build on detected violations.
* **Chores**
  * Updated the service chart’s `images` configuration with a release-time marker indicating it is overwritten during chart publishing.


minmzzhang:
@coderabbitai review

minmzzhang:
@coderabbitai review

.github/workflows/check-floating-tags.yaml
```diff
@@ -0,0 +1,51 @@
...
+      with:
+        persist-credentials: false
+        path: pr
+
+    - name: Check for unmarked floating image tags
+      env:
+        TRUSTED_SCRIPT: ../trusted/.github/scripts/check-floating-tags.sh
+        PR_SCRIPT: .github/scripts/check-floating-tags.sh
+      run: |
+        if [ -f "$TRUSTED_SCRIPT" ]; then
```
sk-ilya: This doesn't add much - the PR can modify the workflow itself anyway. Since the job has read-only permissions and no secrets, we can just run the PR version of the script and keep it simple

charts/service/values.yaml
```diff
@@ -108,6 +108,7 @@ log:
 
 # Image configuration:
 images:
+  # PLACEHOLDER -- overwritten at release time by publish-charts.yaml
```
sk-ilya: Nit: I think placing the comment inline would be more readable, since the current placement makes it look like the entire `images` section is a placeholder:

```yaml
service: ghcr.io/osac/fulfillment-service:main  # PLACEHOLDER
```

minmzzhang: Good suggestion! Moved the comment inline and updated the script to detect both inline and preceding-line PLACEHOLDER markers.

.github/workflows/check-floating-tags.yaml
minmzzhang: You're right that the workflow file itself runs from the base branch on `pull_request` events, so modifying it in a PR has no effect. But the script is a different story: it's checked out from the PR tree by default. Without the trusted checkout, a contributor could replace `check-floating-tags.sh` with a no-op in the same PR that introduces an unmarked floating tag, and the CI gate would pass.

The trusted-script checkout runs the base branch's validator against the PR's chart files, closing that gap. The fallback to the PR copy only kicks in for the initial merge (when the base branch doesn't have the script yet).

.github/scripts/check-floating-tags.sh
```diff
@@ -0,0 +1,37 @@
...
+
+rc=0
+while IFS= read -r -d '' f; do
+  lineno=0
+  prev_line=""
+  while IFS= read -r line || [ -n "$line" ]; do
+    lineno=$((lineno + 1))
+    code_line="$(printf '%s\n' "$line" | sed -E 's/[[:space:]]+#.*$//')"
+    if printf '%s\n' "$code_line" | grep -qE "(^|[[:space:]])tag:[[:space:]]*['\"]?(latest|main)['\"]?([]},[:space:]]|$)|:(latest|main)['\"]?([]},[:space:]]|$)" && \
+       printf '%s\n' "$code_line" | grep -qvE '^[[:space:]]*#'; then
```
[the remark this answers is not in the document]
minmzzhang: [:space:] inside a POSIX bracket expression is valid ERE and expands correctly in GNU grep. Verified:

repo:latestest → no match (correctly rejected)
repo:latest at EOL → match ✓
repo:latest (trailing space) → match ✓
The boundary pattern []},[:space:]] means "one of ], }, ,, or whitespace" — which is the intended set of YAML value terminators. No change needed.

.github/workflows/check-floating-tags.yaml
sk-ilya: Again, suppose I open a PR and do this
```suggestion
        if false; then
```
It would then execute my version of the script. More importantly, we do not need this level of protection for a simple lint-style workflow. If we need to change the script later, this will only add unnecessary hassle.

minmzzhang: Sure! Agree! Simplified the part to run the PR script.

sk-ilya (approved):

PR #955, opened 2026-07-23
OSAC-2959: Private Secrets server

DakCrowder:
Adds a private Secrets server over the postgres db.  Dispatch to secrets backends is not yet implemented, and those methods are stubbed out.

## Summary by CodeRabbit

* **New Features**
  * Added private secrets API support through gRPC and REST.
  * Added create, retrieve, list, update, delete, and signal operations for secrets.
  * Added Vault and HUB backend validation, tenant-aware access, filtering, ordering, and pagination.
  * Secret data is automatically redacted from responses and emitted events.
  * Added validation, optimistic locking, update restrictions, and automatic ID generation.

### DIFF
--- a/internal/servers/private_secrets_server.go
+++ b/internal/servers/private_secrets_server.go
@@ -1,0 +1,256 @@
+/*
+Copyright (c) 2026 Red Hat Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with the
+License. You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific
+language governing permissions and limitations under the License.
+*/
+
+package servers
+
+import (
+	"context"
+	"errors"
+	"log/slog"
+
+	"github.com/prometheus/client_golang/prometheus"
+	grpccodes "google.golang.org/grpc/codes"
+	grpcstatus "google.golang.org/grpc/status"
+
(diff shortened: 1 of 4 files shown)


internal/servers/private_secrets_server.go
```diff
@@ -0,0 +1,256 @@
...
+func (s *PrivateSecretsServer) validateSecretUpdate(_ context.Context,
+	newSecret *privatev1.Secret, existingSecret *privatev1.Secret) error {
+	if newSecret.GetSpec().GetBackend() != privatev1.SecretBackend_SECRET_BACKEND_UNSPECIFIED &&
+		newSecret.GetSpec().GetBackend() != existingSecret.GetSpec().GetBackend() {
+		return grpcstatus.Errorf(grpccodes.InvalidArgument,
+			"field 'spec.backend' is immutable and cannot be changed from '%s' to '%s'",
+			existingSecret.GetSpec().GetBackend(), newSecret.GetSpec().GetBackend())
+	}
+	if existingSecret.GetSpec().GetBackend() == privatev1.SecretBackend_SECRET_BACKEND_HUB &&
+		len(newSecret.GetSpec().GetData()) > 0 {
```
CrystalChun, line 251: maybe naiive but could we reuse the [validateHubSecret](https://github.com/osac-project/fulfillment-service/pull/955/changes#diff-a4fe407fffca9982921374a0cbbd3db8661ac1196e34c31caf15f5f1edcd6046R230-R240) function? Or do we not care about the coordinates field during update?

DakCrowder: I think we can allow for metadata (e.g. label) updates without coordinates being touched so the create and update flows are slightly different.

However I do think that once we have coordinates populated we can do some additional validation there when the backends are fully connected.  So I would expect a bit more when hub is wired up but just not 100% sure what exactly those fields are right now.

CrystalChun (approved):
/lgtm

PR #958, opened 2026-07-26
OSAC-3060: use kind-dev GH Action for CI integration tests

omer-vishlitzky:
## Summary

- Add kind-dev reusable GitHub Action to CI integration test workflow
- Uses `osac-workspace/kind-dev/setup.sh --skip-osac` as single source of truth for Kind environment
- Go test code detects existing cluster and reuses it (faster CI, no duplicate infra install)

### Motivation

Converge on kind-dev as the single deployment path for Kind-based environments across all OSAC repos. Infrastructure versions (cert-manager, trust-manager, etc.) are managed in one place.

### How it works

The kind-dev GH Action creates a cluster named `fulfillment-service-it` with all infrastructure installed. When `ginkgo run it` starts, `it/it_tool.go` calls `Start()` which detects the existing cluster (`existed=true`) and skips infrastructure installation. The tests then build/deploy fulfillment-service on top and run.

### Companion PRs

- osac-test-infra#273 — kind-dev GH Action + itenv removal
- osac-operator#389 — kustomize removal + kind-dev migration
- bare-metal-fulfillment-operator#77 — kustomize removal + kind-dev migration

## Test plan

- [ ] CI integration tests pass with kind-dev providing the cluster
- [ ] Local `IT_KEEP_KIND=true ginkgo run it` still works (Go code creates its own cluster when none exists)

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -40,6 +40,13 @@ jobs:
       with:
         persist-credentials: false
     - uses: ./.github/actions/setup-go
+
+    - name: Create Kind cluster with OSAC infrastructure
+      uses: osac-project/osac-test-infra/.github/actions/kind-dev@main
+      with:
+        cluster-name: fulfillment-service-it
+        mode: skip-osac
+
     - run: |
         echo '127.0.0.1 fulfillment-api.osac.svc.cluster.local' | sudo tee -a /etc/hosts
         echo '127.0.0.1 fulfillment-internal-api.osac.svc.cluster.local' | sudo tee -a /etc/hosts


it/it_tool.go
```diff
@@ -562,10 +566,30 @@ func (t *Tool) startCluster(ctx context.Context) error {
 	if err != nil {
 		return fmt.Errorf("failed to create cluster: %w", err)
 	}
-	err = t.cluster.Start(ctx)
+	existed, err := t.cluster.Start(ctx)
 	if err != nil {
 		return fmt.Errorf("failed to start cluster: %w", err)
 	}
+	if !existed {
```
rccrdpccl: Since we're migrating, can we setup from Makefile and just pass a kubeconfig for go test?

rccrdpccl: we have multiple kind environment efforts going on, I think we should unify them (with toggleable components) possibly under the same chart used to deploy osac, or at least a standalone kind one 
https://github.com/osac-project/osac-workspace/tree/main/kind-dev

omer-vishlitzky: we can't really unify, the link u shared deploys a full osac stack and is much heavier than what we need here. we want a shared IT infrastructure, so a shared go library makes sense here

rccrdpccl: I really have to disagree here. We're trying to have a single way of deploying, this just adds one more way. What problem would this shared library solve? How is this going to be kept in sync? Are we sure all consumers will run the same version? What happens with breaking changes? To me it sounds like a lot more work than just `kind create` && `helm install <osac-chart>`
Furthermore, it will make even more difficult to unify the kind deployment liked above.

If anything, I would move the opposite direction and leverage the current kind manifest (or give kind support to the osac-installer chart)

rccrdpccl (changes requested):

PR #964, opened 2026-07-27
OSAC-2341: create default networking resources at tenant onboarding

danmanor:
## Summary

- Add `DefaultNetworkingProvisioner` that creates default networking resources (VirtualNetwork, IPv4 Subnet, IPv6 Subnet, SecurityGroup, and optionally ExternalIP + NATGateway) when a tenant is created
- Resources are created via DAOs directly to bypass server-level async preconditions (VN must be READY for Subnet, ExternalIP must be ALLOCATED for NATGateway) — all start in PENDING state and are reconciled by the operator
- Wire provisioner into `PrivateTenantsServer.Create` as an optional dependency

## Jira

https://redhat.atlassian.net/browse/OSAC-2341

## Design

[Default Networking EP](https://github.com/osac-project/enhancement-proposals/pull/107) — §Tenant Onboarding

## Changes

**New file: `internal/servers/default_networking_provisioner.go`**
- `DefaultNetworkingProvisioner` with builder pattern
- `Provision(ctx, tenantName)` — reads default NetworkClass, creates VN/Subnets/SG/NATGateway
- All resources labeled `osac.openshift.io/default: "true"` with correct tenant annotation
- Subnets and SG get `osac.openshift.io/owner-reference` annotation pointing to VN
- ExternalIP pool selection: READY pool with most available capacity
- NATGateway auto-allocated ExternalIP with pool capacity decrement

**Modified: `internal/servers/private_tenants_server.go`**
- Added optional `DefaultNetworkingProvisioner` dependency via `SetDefaultNetworkingProvisioner()`
- `Create()` calls `Provision()` after successful tenant creation (same DB transaction)

**Modified: `internal/cmd/service/start/grpcserver/start_grpc_server_cmd.go`**
- Wire provisioner into private tenants server at startup

## Test plan

- [x] 10 unit tests for `DefaultNetworkingProvisioner` covering:
  - Happy path: VN + IPv4/IPv6 Subnets + SG with correct labels, annotations, CIDRs, state
  - NATGateway skipped when `enable_nat_gateway` is false
  - Implementation strategy propagation from NetworkClass
  - No-op when no default NetworkClass exists
  - No-op when NetworkClass has no defaults
  - Partial defaults (IPv4-only)
- [x] 2 unit tests for tenant server wiring:
  - Tenant Create with provisioner creates default resources
  - Tenant Create without provisioner preserves existing behavior
- [x] Full servers test suite passes (1410 specs)
- [x] Full unit test suite passes (80 suites across all internal/ packages)
- [x] `golangci-lint`: 0 issues
- [x] `gofmt`: no formatting changes

## Summary by CodeRabbit

- **New Features**
  - Newly created tenants can automatically receive default networking resources (virtual network, security group, and IPv4/IPv6 subnets when configured).
  - Optional NAT gateway and external IP attachment are provisioned when NAT is enabled and capacity is available.
  - Defaults are applied only when a default network configuration exists.
- **Bug Fixes**
  - Tenant creation returns an internal error if default networking provisioning fails.
  - The gRPC server now refuses to start if the default networking provisioner cannot be constructed.
- **Tests**
  - Added coverage for default, partial, and missing defaults, plus NAT capacity/error scenarios.

### DIFF
--- a/internal/servers/default_networking_provisioner.go
+++ b/internal/servers/default_networking_provisioner.go
@@ -433,6 +433,9 @@ func (p *DefaultNetworkingProvisioner) createDefaultNATGateway(
 			Labels: map[string]string{
 				defaultLabel: "true",
 			},
+			Annotations: map[string]string{
+				ownerReferenceAnnotation: vnID,
+			},
 			Creator: "system",
 		}.Build(),
 		Spec: privatev1.NATGatewaySpec_builder{
--- a/internal/servers/default_networking_provisioner.go
+++ b/internal/servers/default_networking_provisioner.go
@@ -76,11 +76,6 @@ func (b *DefaultNetworkingProvisionerBuilder) Build() (result *DefaultNetworking
 		return
 	}
 
-	newDAO := func(t any) error {
-		return nil
-	}
-	_ = newDAO
-
 	networkClassDao, err := dao.NewGenericDAO[*privatev1.NetworkClass]().
 		SetLogger(b.logger).
 		SetTenancyLogic(b.tenancyLogic).
@@ -169,6 +164,10 @@ func (b *DefaultNetworkingProvisionerBuilder) Build() (result *DefaultNetworking
 
 // Provision creates default networking resources for the given tenant. Returns nil if no default
 // NetworkClass exists or if the NetworkClass has no defaults configured.
+//
+// Callers must ensure ctx carries a database transaction — all resources are created within that
+// transaction so a failure rolls back the entire set. The gRPC interceptor provides this when
+// called from an RPC handler.
 func (p *DefaultNetworkingProvisioner) Provision(ctx context.Context, tenantName string) error {
 	nc, err := findDefaultNetworkClass(ctx, p.logger, p.networkClassDao)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


internal/servers/default_networking_provisioner.go
```diff
@@ -0,0 +1,494 @@
...
+			Annotations: map[string]string{
+				ownerReferenceAnnotation: vnID,
+			},
+			Creator: "system",
+		}.Build(),
+		Spec: privatev1.SecurityGroupSpec_builder{
+			VirtualNetwork:         vnID,
+			Ingress:                defaults.GetIngressRules(),
+			Egress:                 defaults.GetEgressRules(),
+			ImplementationStrategy: "network_policy",
```
ori-amizur, line 328: Is `"network_policy"` intentionally hardcoded here? The VirtualNetwork correctly copies `implementation_strategy` from the NetworkClass, but the SecurityGroup uses a static value. Same question about `Region: "default"` on the VN (line 246) — should these come from the NetworkClass or deployment config, or are they always fixed?

internal/servers/default_networking_provisioner.go
```diff
@@ -0,0 +1,494 @@
...
+	err = p.updateExternalIPAttachedFlag(ctx, externalIPID, true)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (p *DefaultNetworkingProvisioner) findExternalIPPool(ctx context.Context) (*privatev1.ExternalIPPool, error) {
+	listResponse, err := p.externalIPPoolDao.List().Do(ctx)
```
ori-amizur, line 375: `findExternalIPPool` lists all pools with no tenant filter. In a multi-tenant deployment, this would return pools from every tenant. Should it be scoped to system-level pools or pools available to the target tenant?

internal/servers/default_networking_provisioner.go
```diff
@@ -0,0 +1,494 @@
...
+		SetId(poolID).
+		SetLock(true).
+		Do(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to get ExternalIPPool for capacity update: %w", err)
+	}
+
+	pool := getResponse.GetObject()
+	pool.GetStatus().SetAllocated(pool.GetStatus().GetAllocated() + delta)
+	pool.GetStatus().SetAvailable(pool.GetStatus().GetAvailable() - delta)
```
ori-amizur: After the locked `Get`, there is no check that `available >= delta` before decrementing. The `available > 0` filter in `findExternalIPPool` runs before the lock is acquired, so by the time the row is locked and re-read here, the capacity could have changed. A guard like:
```go
if pool.GetStatus().GetAvailable() < delta {
    return fmt.Errorf("ExternalIP pool %s has insufficient capacity", poolID)
}
```
after the locked read would close this gap.

ori-amizur (commented):
Review comments on a few spots — mostly clarification questions and one potential gap.

internal/servers/default_networking_provisioner.go
danmanor: Both are intentional:

- `"network_policy"` — matches the existing pattern in `private_security_groups_server.go:31` (`const securityGroupImplementationStrategy = "network_policy"`). The SG server always hardcodes this on Create (line 161). SGs are enforced via Kubernetes NetworkPolicy, not via the fabric manager.

- `"default"` region — matches the public VN server behavior when region is unset (`virtual_networks_server.go:244-245`). Region is a private-API field; the public server always sets it to `"default"`.

internal/servers/default_networking_provisioner.go
danmanor: ExternalIPPools are platform-level resources, not tenant-scoped. The existing `private_external_ips_server.go` also does unscoped pool lookups via `validatePoolReference` (line 275) — pools are managed by Cloud Provider Admins and shared across tenants. The DAO `List()` here follows the same pattern.

internal/servers/default_networking_provisioner.go
danmanor: Good catch — fixed in commit 90fb3984. Added `newAvailable < 0` guard after the locked read, matching the existing pattern in `private_external_ips_server.go:325`.

danmanor:
/override "E2E BMaaS Full Install / e2e"

PR #976, opened 2026-07-28
OSAC-3247: grant tenant admins Create/Update/Delete on cluster and compute instance catalog items

ElayAharoni:
## Summary
- The OPA policy (`authz.rego`) allowed tenant admins to Create/Update/Delete `BareMetalInstanceCatalogItems` but not `ClusterCatalogItems`/`ComputeInstanceCatalogItems`, which fell through to admin-only despite `Get`/`List` already being open to tenant admins.
- This contradicts the Tenant Admin persona (`docs/personas.md`: "manages organization-specific catalogs of templates") and the `catalog-items` enhancement proposal, which already documents tenant-scoped CRUD for all three catalog item types.
- No application-layer changes needed: `ClusterCatalogItemsServer`/`ComputeInstanceCatalogItemsServer` are structurally identical to `BareMetalInstanceCatalogItemsServer` and already delegate to the same tenancy-aware generic DAO, so tenant-scoped enforcement (auto-set tenant on create, restrict update/delete to the caller's own tenant) already works for all three — this PR only opens the OPA gate.
- Added authz-interceptor unit tests covering tenant-admin allow / regular-user deny for all three catalog item types (none had coverage at this layer before, including the already-working `BareMetalInstanceCatalogItems` case).

## Jira
[OSAC-3247](https://redhat.atlassian.net/browse/OSAC-3247)

## Test plan
- [x] `go build ./...`
- [x] `go test ./internal/auth/...` (240/240 specs passing, including the 12 new entries)
- [x] `go vet ./internal/auth/...`

### DIFF
--- a/internal/auth/grpc_authz_interceptor_test.go
+++ b/internal/auth/grpc_authz_interceptor_test.go
@@ -660,6 +660,80 @@ var _ = Describe("Rego authorization interceptor", func() {
 			Expect(handled).To(BeFalse())
 		})
 
+		DescribeTable(
+			"Allows tenant admin to manage catalog items",
+			func(ctx context.Context, grpcMethod string) {
+				token := createKeycloakUserToken("my-tenant", "my-user", jwt.MapClaims{
+					"realm_access": map[string]any{
+						"roles": []any{
+							"tenant-admin",
+						},
+					},
+				})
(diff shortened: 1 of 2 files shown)


eranco74:
Note https://github.com/osac-project/fulfillment-service/pull/998 already merged

PR #984, opened 2026-07-29
OSAC-3159: Add GpuSpec to InstanceType proto and enforce immutability

Tzif-Morgen:
## [OSAC-3159](https://redhat.atlassian.net/browse/OSAC-3159): Add GpuSpec to InstanceType proto and enforce immutability

**Jira:** [OSAC-3159](https://issues.redhat.com/browse/OSAC-3159)
**Story type:** [DEV]

### Summary

Adds a `GpuSpec` sub-message to `InstanceTypeSpec` in both public and private protos, enabling Cloud Provider Admins to define GPU hardware specifications (PCI device selector, resource name, and count) on instance types. GPU fields are immutable after creation, matching the existing behavior of `cores` and `memory_gib`.

### Changes

- **Proto definitions** (`proto/{public,private}/osac/.../instance_type_type.proto`):
  - Added `GpuSpec` message with `pci_device_selector`, `resource_name`, and `count` fields
  - Added `buf.validate` annotations for field constraints (min_len, gte/lte)
  - Added `optional GpuSpec gpu = 6` to `InstanceTypeSpec` (enables CEL `has()` expressions)

- **Server** (`internal/servers/private_instance_types_server.go`):
  - GPU field validation enforced by proto `buf.validate` annotations via the protovalidate interceptor (no manual server-side duplication)
  - Added GPU immutability check in `validateInstanceTypeImmutability` using `proto.Equal`

- **Unit tests** (`internal/servers/private_instance_types_server_test.go`):
  - Added "Creates object with GPU" and "Creates object without GPU" tests
  - Added "Rejects update of gpu" immutability test
  - Added GPU CEL filter tests (`has(this.spec.gpu)` and `resource_name` filter)
  - Added GPU spec to existing "Update object", "Delete object", and "List objects" tests

- **Integration tests** (`it/it_private_instance_types_test.go`):
  - Added `DescribeTable` for invalid GPU field rejection on Create (4 entries: empty pci_device_selector, empty resource_name, count < 1, count > 16)
  - Tests exercise the full gRPC stack including the protovalidate interceptor

### Testing

- **Unit tests:** 9 new/modified test cases covering create, immutability, CEL filters, update, delete, and list with GPU-enabled instance types
- **Integration tests:** 4 new test cases verifying proto annotation validation via the protovalidate interceptor

### Acceptance Criteria

- [ ] AC-1: GpuSpec message exists in both public and private InstanceType protos
- [ ] AC-2: InstanceType can be created with a valid gpu field
- [ ] AC-3: InstanceType can be created without a gpu field (non-GPU instance type)
- [ ] AC-4: Create rejects gpu with empty pci_device_selector or resource_name
- [ ] AC-5: Create rejects gpu with count < 1 or count > 16
- [ ] AC-6: Update rejects changes to the gpu field (immutability)
- [ ] AC-7: Update allows changes to description and state on GPU-enabled InstanceTypes
- [ ] AC-8: List returns gpu fields in response objects
- [ ] AC-9: CEL filter `has(this.spec.gpu)` returns only GPU-enabled InstanceTypes
- [ ] AC-10: CEL filter on `this.spec.gpu.resource_name` returns matching InstanceTypes
- [ ] AC-11: Cloud Provider Admin can delete GPU-enabled InstanceTypes


internal/servers/private_instance_types_server.go
```diff
@@ -145,6 +145,21 @@ func (s *PrivateInstanceTypesServer) Create(ctx context.Context,
...
+	if gpu := spec.GetGpu(); gpu != nil {
+		if gpu.GetPciDeviceSelector() == "" {
+			err = grpcstatus.Errorf(grpccodes.InvalidArgument, "field 'spec.gpu.pci_device_selector' must not be empty")
+			return
+		}
+		if gpu.GetResourceName() == "" {
+			err = grpcstatus.Errorf(grpccodes.InvalidArgument, "field 'spec.gpu.resource_name' must not be empty")
+			return
+		}
+		if gpu.GetCount() < 1 || gpu.GetCount() > 16 {
```
sk-ilya: Can
```
[(buf.validate.field).int32 = {gte: 1, lte: 16}];
```
catch it? ideally, the proto field constraints should be source of truth, and here the same validation is duplicated

proto/private/osac/private/v1/instance_type_type.proto
```diff
@@ -74,6 +75,18 @@ message InstanceType {
...
+// Hardware specification for GPU devices attached to compute instances provisioned with this instance type.
+message GpuSpec {
+  // PCI device selector identifying the GPU hardware (e.g., "10DE:20B0").
+  string pci_device_selector = 1 [(buf.validate.field).string.min_len = 1];
+
+  // Kubernetes device plugin resource name (e.g., "nvidia.com/A100").
+  string resource_name = 2 [(buf.validate.field).string.min_len = 1];
+
+  // Number of GPU devices of this type.
+  int32 count = 3 [(buf.validate.field).int32 = {gte: 1, lte: 16}];
```
sk-ilya, line 87: 16 max is hard-coded. should it be configurable instead or  set to an unrealistically high value eg 100?

internal/servers/private_instance_types_server.go
Tzif-Morgen: You're completely right. I originally added the manual checks in the server just because the other fields (like cores and memory_gib) were handled there, but you're right, the proto annotations should be the single source of truth, and all validations (cores and gib as well) really belong in the proto.

The issue is that unit tests call the server directly, bypassing the gRPC interceptor chain, which means Create validation isn't triggered during tests.

To fix this without duplicating validation logic in Go, I'm thinking of creating a test helper (e.g., createWithValidation) that invokes protovalidate directly before calling the server. It might not be the most elegant pattern, but it closely mirrors what happens in production via the interceptor, while eliminating duplicated validation code.

What do u think?

sk-ilya: Probably having such a helper used for only one server out of many is worse than the duplication... Alternatively, maybe these tests should be moved to exercise the path that includes validation

Tzif-Morgen: Done - removed the GPU validation from the server and moved the Create rejection tests to integration tests (it/it_private_instance_types_test.go) where the protovalidate interceptor enforces the buf.validate annotations. Didn't touch cores/memory_gib, I think it's out of scope for this PR. Wdyt?

proto/private/osac/private/v1/instance_type_type.proto
```diff
@@ -74,6 +75,18 @@ message InstanceType {
...
+}
```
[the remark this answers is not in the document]
Tzif-Morgen: Dismissing - PCI device selectors and resource names vary across hardware vendors and Kubernetes device plugins. A strict regex would risk rejecting valid inputs. min_len is the right level of validation here.

sk-ilya (approved):
lgtm 
nit  https://github.com/osac-project/fulfillment-service/pull/984#discussion_r3676452224

proto/private/osac/private/v1/instance_type_type.proto
Tzif-Morgen: 16 matches the largest GPU servers available today. Making it configurable would add complexity without practical benefit, and a high value like 100 wouldn't catch input errors.

PR #991, opened 2026-07-29
OSAC-2862, OSAC-2863, OSAC-2864: add public BareMetalInstanceTypes gRPC service with integration

ajamias:
## [OSAC-2862](https://redhat.atlassian.net/browse/OSAC-2862): add public BareMetalInstanceTypes gRPC service

**Depends on:** #988 ([OSAC-2861](https://redhat.atlassian.net/browse/OSAC-2861): private BareMetalInstanceTypes service)

### Summary

Implements the public gRPC service for bare metal instance types, providing read-only List and Get operations for tenants to discover available hardware profiles. The public server delegates to the private server (from [OSAC-2861](https://redhat.atlassian.net/browse/OSAC-2861)) and maps types using GenericMapper, following the same pattern as the existing InstanceTypes public server.

### Changes

- **Proto:** Add `bare_metal_instance_types_service.proto` with List/Get RPCs and REST transcoding annotations under `proto/public/`
- **Server:** Implement `BareMetalInstanceTypesServer` (public) with builder pattern, delegating to `PrivateBareMetalInstanceTypesServer` with public-to-private type mapping
- **Registration:** Register the public server in both the gRPC server and REST gateway startup chains
- **Authorization:** Add `BareMetalInstanceTypes/Get` and `BareMetalInstanceTypes/List` to OPA client permissions policy
- **Generated code:** 4 generated files from `buf generate`

### Testing

- **Unit tests:** 4 builder validation specs (logger, attribution logic, tenancy logic required)
- **Integration tests:** List, Get, filtering by ID/name, pagination with limit/offset
- **Regression:** Full suite passes (83 suites, 0 failures)

### Dependencies

- Depends on [OSAC-2861](https://redhat.atlassian.net/browse/OSAC-2861) (private BareMetalInstanceTypes gRPC service) — PR #988
- Depends on [OSAC-2860](https://redhat.atlassian.net/browse/OSAC-2860) (database migration) — already merged
- Depends on [OSAC-2859](https://redhat.atlassian.net/browse/OSAC-2859) (proto type definition) — already merged

## Summary by CodeRabbit

* **New Features**
  * Added REST and gRPC APIs for bare metal instance types, with public read-only **List**/**Get** endpoints and private **Create/List/Get/Update/Delete/Signal** operations.
  * **List** supports pagination, filtering, and ordering, returning `items`, `size`, and `total`.
* **Bug Fixes**
  * Updated authorization to allow public clients to **List** and **Get** bare metal instance types.
* **Changes**
  * Relaxed schema-level validation constraints for bare metal instance type request messages.


internal/servers/bare_metal_instance_types_server.go
```diff
@@ -0,0 +1,211 @@
...
+var _ publicv1.BareMetalInstanceTypesServer = (*BareMetalInstanceTypesServer)(nil)
+
+type BareMetalInstanceTypesServer struct {
+	publicv1.UnimplementedBareMetalInstanceTypesServer
+
+	logger    *slog.Logger
+	delegate  privatev1.BareMetalInstanceTypesServer
+	inMapper  *GenericMapper[*publicv1.BareMetalInstanceType, *privatev1.BareMetalInstanceType]
+	outMapper *GenericMapper[*privatev1.BareMetalInstanceType, *publicv1.BareMetalInstanceType]
+}
```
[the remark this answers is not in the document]
ajamias: Keeping `inMapper` for consistency with `InstanceTypesServer` (which also has `inMapper` and uses it for Create/Update). If write RPCs are added to the public BareMetalInstanceTypes service later, the mapper will already be in place.

internal/servers/private_bare_metal_instance_types_server.go
```diff
@@ -0,0 +1,366 @@
...
+func (b *PrivateBareMetalInstanceTypesServerBuilder) Build() (result *PrivateBareMetalInstanceTypesServer, err error) {
+	// Check parameters:
+	if b.logger == nil {
+		err = errors.New("logger is mandatory")
+		return
+	}
+	if b.tenancyLogic == nil {
+		err = errors.New("tenancy logic is mandatory")
+		return
+	}
```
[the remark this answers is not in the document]
ajamias: The reference `PrivateInstanceTypesServer.Build()` follows the same pattern — it validates `logger` and `tenancyLogic` but not `attributionLogic`, relying on `GenericServer`'s builder to catch it downstream. Keeping the same validation structure here for consistency across private servers.

internal/servers/private_bare_metal_instance_types_server.go
```diff
@@ -0,0 +1,366 @@
...
+		case "spec.hardware":
+			base.GetSpec().SetHardware(update.GetSpec().GetHardware())
+		case "spec.host_label_selector":
+			base.GetSpec().SetHostLabelSelector(update.GetSpec().GetHostLabelSelector())
+		default:
+			// For unknown paths, fall through - the generic handler will
+			// reject invalid paths if needed.
+		}
+	}
+}
```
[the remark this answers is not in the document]
ajamias: This follows the same `default:` no-op pattern used by `PrivateInstanceTypesServer` and other private servers in the codebase. The comment about the generic handler is acknowledged as imprecise, but changing the error handling here would be inconsistent with the established server pattern. If we want to reject unknown paths, that should be a cross-cutting change across all servers, not scoped to this one.

internal/servers/private_bare_metal_instance_types_server.go
```diff
@@ -0,0 +1,366 @@
...
+			return
+		}
+		if accelerator.MemoryGb != nil && *accelerator.MemoryGb <= 0 {
+			err = grpcstatus.Errorf(grpccodes.InvalidArgument, "field 'spec.hardware.accelerators[%d].memory_gb' must be greater than zero", i)
+			return
+		}
+	}
+
+	// Validate network port specifications:
+	for i, port := range hardware.GetNetworkPorts() {
```
adriengentil, line 207: I saw you added the validation in the proto fields. Is it still required to add them here?

internal/servers/private_bare_metal_instance_types_server.go
```diff
@@ -0,0 +1,366 @@
...
+		return grpcstatus.Errorf(grpccodes.InvalidArgument,
+			"field 'name' is immutable and cannot be changed from '%s' to '%s'",
+			existing.GetMetadata().GetName(), merged.GetMetadata().GetName())
+	}
+
+	// Validate immutable hardware specs (core hardware configuration):
+	existingHw := existing.GetSpec().GetHardware()
+	mergedHw := merged.GetSpec().GetHardware()
+
+	if existingHw != nil && mergedHw != nil {
```
adriengentil: can't we just do a reflect.DeepEqual between existingHw and mergedHw ?

internal/servers/private_bare_metal_instance_types_server.go
ajamias: It seems like almost none of the other proto files use validation the way it's used for BareMetalInstanceTypes. I'll get rid of validation in the proto fields to be consistent with the pattern.

internal/servers/private_bare_metal_instance_types_server.go
ajamias: The field-by-field approach gives more control over what's considered immutable vs mutable, following the same pattern used by PrivateInstanceTypesServer.

internal/servers/private_bare_metal_instance_types_server.go
adriengentil: the `buf.validation` was introduced recently, and I think it's a good practice to use that as it reduces the amount of boilerplate.We just need to understand how to use that efficiently.

internal/servers/private_bare_metal_instance_types_server.go
adriengentil: everything is immutable under spec.hardware, right? I don't see the value to do it field by field, it might increase the maintenance/introduce bug if we add a new field.

PR #1001, opened 2026-07-30
OSAC-3393: Remove users in Organization Group when removed from project

CrystalChun:
If a ProjectMembership still exists, but a user is removed from the list of users in that project, the user should be removed from the group in Keycloak.

/cc @danielerez @DakCrowder @rccrdpccl 

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved project membership recovery after failed synchronization attempts.
  - Preserved successfully processed users when some assignments or removals fail, enabling more targeted retries.
  - Prevented unnecessary retries for terminal failures such as deleted projects, invalid roles, or missing tenants.
  - Treated users already belonging to a group as successfully added.
  - Preserved private membership settings when updating records without an update mask.

### DIFF
--- a/internal/controllers/projectmembership/project_membership_reconciler_function.go
+++ b/internal/controllers/projectmembership/project_membership_reconciler_function.go
@@ -22,6 +22,7 @@ import (
 	"errors"
 	"fmt"
 	"log/slog"
+	"sort"
 	"strings"
 
 	"google.golang.org/grpc"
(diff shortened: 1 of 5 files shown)


internal/controllers/projectmembership/project_membership_reconciler_function.go
```diff
@@ -147,16 +147,23 @@ func (t *task) update(ctx context.Context) error {
...
-	if state == privatev1.ProjectMembershipState_PROJECT_MEMBERSHIP_STATE_FAILED {
-		return nil
-	}
-
 	// For READY memberships, detect user list changes and sync accordingly
 	if state == privatev1.ProjectMembershipState_PROJECT_MEMBERSHIP_STATE_READY {
 		return t.handleUserListChange(ctx)
 	}
 
+	// For FAILED memberships, use the diff path if we have a previous user list,
```
danielerez: So it won't be a terminal error state as before, right? I mean, do we want to keep retrying on every cycle even for permanent failures (as a deleted project, [etc](https://github.com/osac-project/fulfillment-service/blob/049b1d0b398013b5a171265fc6232eccaf2c1425/internal/controllers/projectmembership/project_membership_reconciler_function.go#L330))?

internal/controllers/projectmembership/project_membership_reconciler_function_test.go
```diff
@@ -109,6 +110,182 @@ var _ = Describe("Default Values", func() {
...
+			r:          functionObj,
+			membership: membership,
+		}
+
+		err := task.update(ctx)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(membership.GetStatus().GetMessage()).To(Not(ContainSubstring("Failed to sync")))
+	})
+})
+
```
danielerez, line 517: Maybe worth also adding a test for this scenario: failed to add user-b while user-a succeeded in an earlier sync.

danielerez (commented):
Looks good. Just a couple of comments.

internal/controllers/projectmembership/project_membership_reconciler_function.go
CrystalChun: Yes that's a great point! I added a check to see if it's a failure that should stop reconcile. What do you think?

internal/controllers/projectmembership/project_membership_reconciler_function_test.go
CrystalChun: Good idea! Added a test, how does it look?

CrystalChun:
@coderabbitai review

DakCrowder (approved):

internal/controllers/projectmembership/project_membership_reconciler_function.go
danielerez: Looks good, thanks:)

